PR #58, opened 2017-09-12
Docs eric

erbou:


docs/sphinx.sh
```diff
@@ -0,0 +1,103 @@
+#! /bin/bash
```
jirikuncar: What about using `#!/usr/bin/env bash` ?

docs/sphinx.sh
```diff
@@ -0,0 +1,103 @@
...
+
```
jirikuncar: IMHO we should add a license header here too.

docs/sphinx.sh
```diff
@@ -0,0 +1,103 @@
...
+        esac
+    done
+
+    pushd "${ROOT}"
+    run_python_cmd "${python_exe}" "${virtual_env}"
+    popd
+}
+
+main "$@"
+
```
jirikuncar: probably the empty line can be removed

PR #61, opened 2017-09-13
adding doc about UI

ebiiii:


erbou:
http://localhost/admin/swagger as mentioned in the doc does not work for me, we need to add a / to it, i.e. http://localhost/admin/swagger/

PR #62, opened 2017-09-13
Add ELK docker

ebiiii:


jirikuncar:
Can we put this on hold after the release? We should probably have simple docker-compose and full tutorial for production where the logging runs on separate service stack.

leafty:
Cleanup

PR #63, opened 2017-09-13
Added volumes for persistent data (db+cassandra)

### LINKED ISSUE #23: Use volumes for db and cassandra?

leafty:
Closes #23 if merged.


jirikuncar:
Shouldn't we turn `./services/storage/data:/data` into volume too?

ebiiii:
For the storage, having the files easily accessible on the local file-system is much more convenient. 
In the case of the databases(PG, Cassandra), the files only make sense when interpreted by an application and thus being a volume is fine (you can mount them in a new PG/cassandra container to browse the data). But the file storage backend uses regular files, that a user may want to access directly, without having to start a new docker image to browse them and extract them.

Issue #49, opened 2017-09-08
Support deployment on different domain than localhost


jirikuncar:
- [x] configurable docker-compose.yml
- [ ] generate certificates using letsencrypt

rokroskar:
revisit this in a deployment sprint for the development branch

PR #66, opened 2017-09-13
Configurable docker-compose

jirikuncar:
addresses #49


docker-compose.yml
```diff
@@ -121,6 +126,7 @@ services:
       traefik.frontend.rule: PathPrefix:/api/resource-manager
   deployer:
     image: ${IMAGE_REPOSITORY}renga-deployer:$PLATFORM_VERSION
+    command: uwsgi /code/docker/uwsgi/uwsgi.ini
```
leafty: Makes deployer fail on startup, overkill and prevents stack traces from being printed.

PR #71, opened 2017-09-14
Added docker-compose as a prerequisite

leafty:
docker-compose is not installed with docker, unlike on mac.

### DIFF
--- a/docs/user/setup.rst
+++ b/docs/user/setup.rst
@@ -18,6 +18,7 @@ Prerequisites
 -------------
 
 * `Docker <http://www.docker.com>`_
+* `Docker Compose >=1.14 <https://docs.docker.com/compose/install/>`_ (**Linux Only**)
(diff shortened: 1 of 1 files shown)


rokroskar:
it doesn't come for free with docker?

docs/user/setup.rst
```diff
@@ -18,6 +18,7 @@ Prerequisites
 -------------
 
 * `Docker <http://www.docker.com>`_
+* (Linux) `Docker Compose (>=1.14) <https://docs.docker.com/compose/install/>`_
```
jirikuncar: Can you move it to separate subsection for Linux?

rokroskar: or `Docker Compose (>=1.14 **Linux only**) <https://docs.docker.com/compose/install/>`_

leafty:
@rokroskar No, see the linked doc (https://docs.docker.com/compose/install/)

PR #75, opened 2017-09-15
Added networks to compose

leafty:


rokroskar:
closing for now, we will get back to network issues at a later point when we start to build actual deployment examples

PR #78, opened 2017-09-15
Fix endpoints environment variable in Makefile

erbou:
Fix the environment variables RENGA_ENDPOINT and RENGA_CONTAINERS_ENDPOINT in the Makefile. We may need to use a similar approach in the stop target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,12 +107,14 @@ docker-images: $(scala-services) $(dockerfile-services)
 # Platform actions
 .PHONY: start stop restart test
 start:
-	export RENGA_ENDPOINT=http://localhost
-	export RENGA_CONTAINERS_ENDPOINT=http://`docker network inspect bridge --format='{{(index (index .IPAM.Config) 0).Gateway}}'`
-	@docker-compose build
-	@docker-compose create
-	@docker-compose up -d
-	@./scripts/wait-for-services.sh
+ifeq ($(RENGA_ENDPOINT),)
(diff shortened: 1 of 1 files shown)


rokroskar:
@erbou I changed this part quite a lot in the final PR, can you please have a look at the current master?

Makefile
```diff
@@ -132,10 +132,14 @@ docker-images: $(scala-services) $(dockerfile-services)
 # Platform actions
 .PHONY: start stop restart test
 start:
+ifeq ($(RENGA_ENDPOINT),)
+	RENGA_ENDPOINT="http://localhost" RENGA_CONTAINERS_ENDPOINTS=http://`docker network inspect bridge --format='{{(index (index .IPAM.Config) 0).Gateway}}'` $(MAKE) start
+else
```
rokroskar: won't this skip the docker-compose commands if `RENGA_ENDPOINT` is defined?

erbou: If RENGA_ENDPOINT is undefined or empty, it will set RENGA_ENDPOINT and RENGA_CONTAINERS_ENDPOINT to the default values, and recursively invoke the make start command with the defined variables.
If it is defined and not empty (e.g. in the recursive call), it will invoke the docker-compose command.

rokroskar: Ah didn't see the recursion, sorry -- I also did this similarly on L43 and L48, not sure which method is preferred?

erbou: Sorry, I didn't realize that it was part of the change when I resolved the conflicts. Yes you can ignore this PR.

PR #79, opened 2017-09-17
Fix replacement of variable in Keycloak template

jirikuncar:


rokroskar:
the problem seems to be that the travis user has uid 2000 but the jboss user inside the keycloak container has uid 1000 so it doesn't have permission to write to the mounted directory

docker-compose.yml
```diff
@@ -17,6 +17,8 @@
 # limitations under the License.
 
 version: '3'
+volumes:
+  keycloak-data: {}
```
rokroskar: this just makes the volume to be mounted as root and the file write fails again:

```
[jboss@57bab8d1bca9 ~]$ ls -la /data
total 72
drwxr-xr-x 2 root  root   4096 Sep 18 19:17 .
drwxr-xr-x 1 root  root   4096 Sep 18 19:17 ..
-rwxr-xr-x 1 jboss jboss  1168 Sep 18 14:53 docker-entrypoint.sh
-rw-r--r-- 1 jboss jboss 59979 Sep 18 08:46 renga-realm.json.tpl
```

rokroskar:
I know 503ad6a is not the dynamic configuration we all want for keycloak and the user will be required to start the platform with `make start` (unless they create their own keycloak configuration) -- however, until we have clarified why docker mounts directories under different uids depending on the host system, I think this is the safe option to avoid failing builds elsewhere.

docker-compose.yml
```diff
@@ -56,9 +55,7 @@ services:
       - "-Dkeycloak.migration.file=/data/renga-realm.json"
       - "-Dkeycloak.migration.strategy=IGNORE_EXISTING"
     volumes:
-      - ./services/keycloak/docker-entrypoint.sh:/data/docker-entrypoint.sh:ro
-      - ./services/keycloak/renga-realm.json.tpl:/data/renga-realm.json.tpl:ro
-      - ./services/keycloak/data:/data:rw
+      - ./services/keycloak/renga-realm.json:/data/renga-realm.json
```
leafty: Not read-only?

rokroskar: actually I couldn't see any difference on the resulting file permissions when I used the flags in docker-compose

leafty: This does not necessarily translates into file permissions, but the volume driver will effectively forbid writting of the file

PR #83, opened 2017-09-19
Added python version in the firststeps section.

3C111:


docs/user/firststeps.rst
```diff
@@ -8,7 +8,8 @@ First Steps
     :keywords: hello world, first steps, starter, primer
 
 To try out Renga, you first need a platform to connect to: see :ref:`setup`
-for instructions on how to get one running on your own machine in a few minutes.
+for instructions on how to get one running on your own machine in a few minutes. 
+You will also need Python with version 3.x.x.
```
rokroskar: maybe add it as a note? it will stand out more, e.g. 

```
.. note::
    You will also need Python with version 3.x.x.
```

3C111: Okay, can do

docs/user/firststeps.rst
```diff
@@ -8,7 +8,10 @@ First Steps
     :keywords: hello world, first steps, starter, primer
 
 To try out Renga, you first need a platform to connect to: see :ref:`setup`
-for instructions on how to get one running on your own machine in a few minutes.
+for instructions on how to get one running on your own machine in a few minutes. 
+.. note::
+
+    You will also need Python with version 3.x.x.
```
jirikuncar: `3.6.x`

jirikuncar: We can simply change `pip` to `pip3` to make it clear.

PR #87, opened 2017-09-19
Added targets to build renga-ui properly

### LINKED ISSUE #82: running renga-ui image built with Makefile fails
the Makefile target for `renga-ui` should run the `npm` commands before building the docker image

leafty:
Fixes #82.


Makefile
```diff
@@ -126,6 +126,13 @@ renga-commons-artifact: renga-graph-artifact
 $(dockerfile-services): %: $(PLATFORM_BASE_DIR)/%
 	docker build --tag $(IMAGE_REPOSITORY)$@:$(PLATFORM_VERSION) $(PLATFORM_BASE_DIR)/$@
 
+# be sure to build ui before make docker image
```
jirikuncar: IMHO this should be part of the Dockerfile in renga-ui.

leafty: Not sure it makes sense, the container doesn't have nodejs installed.

jirikuncar: Can you just install it in Dockerfile and then clean it after the build?

ebiiii: If you don't want to have to install npm locally (which you would anyway need to dev in the UI), there are 3 solutions:

 - add the "dist" folder and its content to the git, but then you need to make sure that it is rebuild at each change (maybe with travis)
 - add a second Dockerfile based on nodejs for building (that would cp out the compiled files)
 - add nodejs, build, cleanup in the main Dockerfile (but this make the build sensibly longer)

jirikuncar: I agree to have one Dockerfile for building the Python backend and second one for nodejs. The final docker-compose will mount volume from the nodejs based image to the Python backend.

Makefile
```diff
@@ -126,6 +128,11 @@ renga-commons-artifact: renga-graph-artifact
 $(dockerfile-services): %: $(PLATFORM_BASE_DIR)/%
 	docker build --tag $(IMAGE_REPOSITORY)$@:$(PLATFORM_VERSION) $(PLATFORM_BASE_DIR)/$@
 
+	# build docker images from makefiles
```
jirikuncar: Is the indentation intended?

leafty: No it wasn't.

PR #107, opened 2017-10-05
tests: always install requirements

jirikuncar:


jirikuncar:
@3C111 I have removed the old scaladomain extension.

PR #108, opened 2017-10-06
Docs api spec

ableuler:
Still work in progress

### DIFF
--- a/docs/developer/web.rst
+++ b/docs/developer/web.rst
@@ -1,0 +1,234 @@
+.. _developer_web:
+
+
+Renga Web API
+=============
+
+This page describes a set of rules and recommendations that a developer of
+any **Renga** related web service should follow when developing new or
+refactoring existing code.
+
+.. note::
+
+   This is not a formal specification.
+
+0. First steps
+--------------
+
+Make sure that your data model has started to stabilize. Focus on consistent
+and concrete naming of your *resources* and possible relations between them.
+
+
+1. Specification and Best Practices
+-----------------------------------
+
+The focus is on finding a right balance between standards and pragmatic API
+design which should serve its purpose.
+
+- `JSON API Specification`_ - a specification for building APIs in JSON
+- `Pragmatic RESTful API`_ - blog post on best practices for designing
+  a pragmatic RESTful API
+
+.. _`JSON API Specification`: http://jsonapi.org/format/
+.. _`Pragmatic RESTful API`: http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api
+
+There are some key principles that the developer of an API should aim for.
+
+2. HTTP verbs
+-------------
+
+Most of the *resources* should support the four basic functions of `CRUD`_:
+
+**Create**
+
+    ``POST`` method **should** be used when a request is non-idempotent (for
+    example a new resource is added to a database).
+
+**Read**
+
+    ``GET`` method **should not** change the state of the object itself. Side
+    effects are possible for endpoints that require authorization only, but in
+    any case these side effects should only create new objects used for
+    capturing the resource lineage.
+
+**Update**
+
+    ``PUT`` method should be used to replace a full state of an existing
+    object. To avoid `lost update <https://www.w3.org/1999/04/Editing/#3.1>`_,
+    each request **must** include an ``ETag`` of the resource it is replacing.
+    The ``ETag`` **must** be part of the ``If-Match`` header.
+
+    ``PATCH`` method should be used for partial modification of
+    a resource. Each request **should** include an ``ETag`` of the resource
+    it is replacing.
+
+**Delete**
+
+    ``DELETE`` method deletes the specified resource.
+
+2.1 Beyond CRUD
+~~~~~~~~~~~~~~~
+
+Furthermore, resources **should** also support the following methods:
+
+**Options**
+
+    ``OPTIONS`` method should be used to query for the allowed methods of the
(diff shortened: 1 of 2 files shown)


docs/developer/web.rst
```diff
@@ -0,0 +1,224 @@
...
+A relation between resources **must** be represented in the URL structure if
+the related resources only exist in the context of the relation. For example,
+if in a situation where a file has primary key composed of a bucket identifier
+and a file identifier, the URLs would be structured as followed:
+
+- URL for all files inside the bucket:
+    ``/buckets/{bucket}/files``
+
+- URL for a specific file in a given bucket:
+    ``/buckets/{bucket}/files/{file}``
```
ebiiii: As a file can technically be replicated in several buckets, would this one be also a redirect to something like /files/{file} like in the example below ?

docs/developer/web.rst
```diff
@@ -0,0 +1,224 @@
...
+
+- Retrieving all files, ordered alphabetically by file type and by descending file size:
+    ``GET /files?sort=type,-size``
+
+6.3 Pagination
+~~~~~~~~~~~~~~
+
+Pagination **must** be supported by all endpoints returning more than just a single
+resource. Pagination **must be** implemented using the query parameters ``page``
+and ``per_page``.
```
ebiiii: maybe "page_size" ? (I've seen it more often. Especially while working with django-rest-framework)

docs/developer/web.rst
```diff
@@ -0,0 +1,224 @@
...
+6. Query Parameters
+-------------------
+
+6.1 Filtering
+~~~~~~~~~~~~~
+
+Unique query parameters **must** be used for each field that allows filtering.
+
+- Retrieving all executions which are currently running:
+    ``GET /executions?status=running``
```
ebiiii: It will be important to document all the accepted parameters. Or would those be dynamic according to the model of the resource (allowing to filter on all attributes)?

docs/developer/web.rst
```diff
@@ -0,0 +1,224 @@
...
+    but in any case these side effects should only create new objects used
+    for capturing the resource lineage.
+
+**Update**
+
+    ``PUT`` method should be used to replace a full state of an existing
+    object. To avoid `lost update <https://www.w3.org/1999/04/Editing/#3.1>`_,
+    each request **must** include an ``ETag`` of the resource it is
+    replacing. The ``ETag`` must be part of the ``If-Match`` header.
+
```
3C111: Can you elaborate on how this "ETag" works concretely? Is this something which is in the headers of the request or does the back-end send them explicitely?

jirikuncar: @3C111 the backend should send them explicitly. There is a simple example https://www.w3.org/1999/04/Editing/new-entry.txt

docs/developer/web.rst
jirikuncar: I don't think you can access a file outside a context of bucket.

docs/developer/web.rst
jirikuncar: > used for each field that allows filtering

It should be a documented list of attributes that allow filtering.

docs/developer/web.rst
jirikuncar: This can be an open discussion + voting.

docs/developer/web.rst
leafty: Specified in https://tools.ietf.org/html/rfc7232

docs/developer/web.rst
leafty: Buckets exist for administrative and technical reasons. Files are conceptually not dependent on buckets.

jirikuncar: Feel free to find another example, however I still think that the only way how you can access file is via a bucket.

ebiiii: A file exists independently of its actual storage and you may want to refer to it in a general way (just with its UUID for example), then when actually reading the file it would be dereferenced to the most suited location, like closest to the current execution (that's the reason why we have the file_location nodes which are themselves necessarily linked to a bucket). This allows also for referencing files that are outside of the platform and which would have a sort of "external ref" node instead of the file_location. 

You can think about a "file" as a conceptual dataset (like the well-known iris dataset), file_location is the physical instance of this dataset stored somewhere (like iris.dat in the bucket "my-data" on SWITCH), you also have file_versions which represents a specific version of this file on the specific storage (which could be using S3 versioning system, or just adding the timestamp to the filename as we do for local storage).

jirikuncar: I don't think the discussion about external files should be linked to URL structure. The point here was that if we have files attached to bucket that can have an access control, then we need to make sure that the file is accessed via that bucket. Hence, the developer should structure the URL as described.

Personally, I would like to separate the discussion about protocol/API to access file and the resource definition.

ableuler: So I guess in renga we don't have a really good example for the case where a resource exists only in relation to another resource then. I suggest I just use a made-up non-renga example here instead.

ebiiii: I agree that the discussion of the representation of the specific resource of type "file" is not the main point I wanted to bring here. I was just saying that depending on how it is dealt with later on, taking the example of the file may lead to misunderstanding of the users. As most of our resources will need some very unique identifiers, they may all end up being first-class citizens that can be accessed directly.
As I have no better example of a resource that is meaningless without its relation with another one, we can either keep it like it is now until the resource representation is clearly defined or take a fake example to convey our message, which is still relevant in this document.

PR #109, opened 2017-10-18
Updated the explorer calls

3C111:
Updated the explorer docs with the new api routes.

### DIFF
--- a/docs/developer/services/explorer_service.rst
+++ b/docs/developer/services/explorer_service.rst
@@ -49,8 +49,7 @@ The explorer API provides the following endpoints.
 
 **GET /storage/bucket/:id/files/*path**
 
-  ``StorageExplorerController.fileMetadatafromPath(id: Long, path:
-  String)``
+  ``StorageExplorerController.fileMetadatafromPath(id: Long, path:  String)``
 
   Retrieve metadata of a file with ``path`` (filename) in bucket ``id``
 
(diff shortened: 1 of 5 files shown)


docs/developer/services/explorer_service.rst
```diff
@@ -93,18 +90,53 @@ The explorer API provides the following endpoints.
 
   Retrieve the metadata of a node with ``id``. Valid for all nodes.
 
+**GET /graph/node/:id/edges**
```
rokroskar: I'm not sure this conforms to the API spec -- shouldn't the nouns be plural? I'm guessing this might affect other endpoints so leave until API refactor?

docs/developer/services/explorer_service.rst
```diff
@@ -93,18 +90,53 @@ The explorer API provides the following endpoints.
...
   Retrieve the metadata of a node with ``id``. Valid for all nodes.
 
+**GET /graph/node/:id/edges**                    
+
+  ``GenericExplorerController.retrieveNodeEdges( id: Long )``
+
+  Get all in- and outgoing edges of a node with  ``id``.
+
+
+**GET /graph/nodes/:prop**
```
rokroskar: should these be done in the form of query parameters rather then URL? e.g. /graph/nodes?prop=woop

Same goes for the ones below. It's fine for me to leave these as is for the moment if that's how they're implemented in the explorer, but they will likely need to be changed during the API refactor.

3C111: For now I had just updated the docs with the current state of the API. I didn't look into the refactoring yet

PR #126, opened 2018-02-23
Adapt ui client urls for dev setting

ableuler:
Added http://localhost:3000 as allowed renga-ui client url  to gitlab and keycloak setup for a better ui development setting.


rokroskar:
I think the dev-specific configurations for renga-ui should be in the renga-ui repo, not here.

ableuler:
Almost all of them are, but I can't run the ui without keycloak and gitlab, so keycloak and gitlab must allow me to redirect back to http://localhost:3000

PR #127, opened 2018-02-26
added env target

leafty:
So that we don't have to read the file to guess the variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,10 @@ scala-artifact = \
 .PHONY: all
 all: docker-images
 
+.PHONY: docker-env
+docker-env:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -105,6 +105,10 @@ scala-artifact = \
 .PHONY: all
 all: docker-images
 
+.PHONY: env
+env:
```
jirikuncar: We should call it `docker-env` or `docker-compose-env` to keep the naming consistent.

leafty:
@rokroskar can you check now?

rokroskar:
@leafty `make: Nothing to be done for `env'.`

leafty:
it's `make docker-env` now

Issue #129, opened 2018-02-28
Define and use volumes


leafty:
https://github.com/SwissDataScienceCenter/renga/blob/28e40ab6b91e2b2ccda7f8d888dc8c6fa36ac0a0/docker-compose.yml#L127-L130

Issue: files generated by services are note wiped after `make wipe`.
Cause: the files are put in `./services` and never removed (also, they can only be removed by `root` on linux, c.f. user trick in https://github.com/SwissDataScienceCenter/renga-storage/blob/master/docker-entrypoint.sh).
Solution: do not bind mount `rw` volumes except the docker socket. Instead, declare and use volumes in the compose file.

rokroskar:
yes, this was a mistake carried over from the incubator repo -- I thought we fixed it but maybe it was in a branch that was never merged. What's wrong with just removing the `services/gitlab` directory in `make wipe`? (right now it's removing `./gitlab` which doesn't exist)

leafty:
The files are owned by `root` on linux. The behavior of `osxfs` storage driver is not the usual one (ownership not mapped).

rokroskar:
👍 got it. Volumes it is.

rokroskar:
we should do the same for `gitlab-runner`, btw

rokroskar:
closed by #132

PR #130, opened 2018-03-01
kubernetes helm charts

rokroskar:
Produce k8s helm charts for deploying renga. 

- [x] integrate the pieces from `docker-compose.yml`
- [x] automate adding the `renga-ui` app to gitlab db
- [x] move the gitlab config from the environment variable to a template file
- [x] remove unneeded pieces from gitlab (redis, postgres)
- [x] persistent storage for gitlab configuration (#129)

---

addresses #135  and https://github.com/SwissDataScienceCenter/incubator-renga-k8s/issues/1


charts/renga/charts/gitlab/values.yaml
```diff
@@ -0,0 +1,52 @@
...
+  # existingClaim:
+
+  # storageClass: "-"
+  accessMode: ReadWriteOnce
+  size: 30Gi
+
+  gitlab-data:
+    subPath: data
+    mountPath: /var/opt/gitlab
+  gitlab-config:
```
rokroskar: I think the mountpaths for config and logs are switched @leafty?

charts/renga/values.yaml
```diff
@@ -26,6 +26,8 @@ global:
     postgresDatabase: gitlabhq_production
     postgresUser: gitlab
     postgresPassword: gitlab
+    clientSecret: open_secret
+    rootPassword: root_password
```
leafty: This doesn't need to be global AFAIK.
can we move it to the `gitlab section`? Rename the exisitng `gitlab.password` value to `gitlab.rootPassword` if desired.

rokroskar: Well I think we want to provide sensible defaults in `charts/gitlab/values.yaml` and potentially override them in the top-level `values.yaml`. But you're right, these should not in any case be in a part of `global`.

rokroskar: actually I think it does need to be global because both keycloak and gitlab need to access the value?

leafty: No, that values is only found in `charts/gitlab/templates/secret.yaml`

rokroskar: ah I thought you were referring also to `clientSecret` -- yes, `rootPassword` can be removed, I just did in 97319b0

charts/minikube-values.yaml
```diff
@@ -0,0 +1,15 @@
...
+  enabled: true
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/ssl-redirect: false
+  hosts:
+  - ""
+
+global:
+  renga:
+    domain: "192.168.99.100"
```
leafty: minikube ip are not fixed

rokroskar: fixed in 44159be

PR #132, opened 2018-03-02
[WIP] Added volumes for services

### LINKED ISSUE #129: Define and use volumes

leafty:
Resolves #129 

To do: disable lfs and remove associated volume

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,10 @@
 version: '3'
 volumes:
   db-data: {}
-  cassandra-data: {}
+  gitlab-config: {}
+  gitlab-logs: {}
+  gitlab-data: {}
+  gitlab-runner-data: {}
 services:
   db:
     image: postgres:9.6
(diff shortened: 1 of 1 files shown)


rokroskar:
@leafty is this still WIP or can it be merged?

leafty:
Yes it can be merged I was waiting a bit to see if storage would be merged beforehand. Proper volumes needs to be configured for it too.

Issue #117, opened 2018-02-15
Supporting the simple(st) user story


rokroskar:
This is a meta-issue to keep track of work done to support the basic data science user story.

Please add a link to the relevant issue once it's created in the corresponding repository. Track the fine-grained issues in [sprint project](https://github.com/orgs/SwissDataScienceCenter/projects/15).

## 0. Housekeeping
- [x] Building docs from development on readthedocs
  - [x] renga-python
  - [x] renga
- [x] Updating all the READMEs on current masters to inform of work on development and make the instructions relevant to the development branch (see #116)
- [x] Integrate pieces of old Makefile that builds images across all repos with the new Makefile in development (#121)
- [ ] Travis-ci integration test for the development branch of `renga`
- make `make start` should bring up all the needed services
  - [x] ui (#125, related is https://github.com/SwissDataScienceCenter/renga-ui/pull/119))
  - [x] gitlab
  - [x] keycloak
  - [ ] storage (#128)

## 1. Initiate a project, import a dataset
- [ ] Finalize dataset import to work with renga repos (https://github.com/SwissDataScienceCenter/renga-python/issues/99)
- [x] update `.gitattributes` to track outputs of `renga run` and datasets in LFS (https://github.com/SwissDataScienceCenter/renga-python/issues/104)
* metadata 
  - [ ] document the ontology for projects, datasets, and CWL steps/workflows (https://github.com/SwissDataScienceCenter/renga-ontology/issues/3)
  - [x] include JSON-LD context with all metadata (https://github.com/SwissDataScienceCenter/renga-python/issues/105) (@jirikuncar)

## 2. Explore the dataset using a notebook (isolated environment)
- [ ] Look into capturing data artifacts the same way as with `renga run`

## 3. Run code on data, recording data provenance for reproducibility
- [ ] write user stories as test cases for `renga run`
- [x] Fix bugs with provenance calculation (https://github.com/SwissDataScienceCenter/renga-python/issues/106)
- [x] Resolve lineage in submodules (https://github.com/SwissDataScienceCenter/renga-python/issues/120)
- [x] Implement `renga status` that checks the validity of data provenance (https://github.com/SwissDataScienceCenter/renga-python/issues/107)
- [x] Write out workflows, fix problem with relative paths (https://github.com/SwissDataScienceCenter/renga-python/issues/98)
* (Dockerization & python environments -- for future sprint)

## 4. Push project to a server for sharing 
- [ ] Finalize storage service 
[...]

PR #133, opened 2018-03-02
eliminate using localhost everywhere

rokroskar:
The PR:

* removes references to `localhost`and `docker.for.mac.localhost` and instead uses `renga.local` as the default domain for development 
* the developer must add `renga.local`, `gitlab.renga.local` and `keycloak.renga.local` to `/etc/hosts` in order for the names to resolve. 
* Traefik uses host-based rules instead of prefix-based. 

---

addresses #117

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,11 +22,7 @@ else
     detected_OS := $(shell sh -c 'uname -s 2>/dev/null || echo not')
 endif
 
-ifeq ($(detected_OS), Darwin)
-	PLATFORM_DOMAIN?=docker.for.mac.localhost
-else
-	PLATFORM_DOMAIN?=localhost
-endif
+PLATFORM_DOMAIN?=renga.local
 
 PLATFORM_BASE_DIR?=..
 PLATFORM_BASE_REPO_URL?=https://github.com/SwissDataScienceCenter
@@ -39,8 +35,8 @@ endif
 
 GIT_MASTER_HEAD_SHA:=$(shell git rev-parse --short=12 --verify HEAD)
 
-GITLAB_URL?=http://$(PLATFORM_DOMAIN):5080
-GITLAB_REGISTRY_URL?=http://$(PLATFORM_DOMAIN):5081
+GITLAB_URL?=http://gitlab.$(PLATFORM_DOMAIN)
+GITLAB_REGISTRY_URL?=http://gitlab.$(PLATFORM_DOMAIN):5081
 GITLAB_DIRS=config logs git-data lfs-data runner
 
 DOCKER_REPOSITORY?=rengahub/
@@ -68,7 +64,7 @@ endif
 
 ifndef RENGA_UI_URL
 	# The ui should run under localhost instead of docker.for.mac.localhost
-	RENGA_UI_URL=http://localhost
+	RENGA_UI_URL=http://$(PLATFORM_DOMAIN)
 	export RENGA_UI_URL
 endif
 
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -188,6 +200,7 @@ services:
       --defaultentrypoints=http,https \
       --entrypoints='Name:http Address::80' \
       --entrypoints='Name:https Address::443 TLS:/ssl/test.crt,/ssl/test.key' \
+      --entrypoints='Name:service Address::8080' \
```
jirikuncar: why do we need an extra entrypoint?

rokroskar: because of token issuer problems. Note that 8080 is the standard port.

Makefile
```diff
@@ -247,6 +243,9 @@ endif
 	@echo
 	@echo "[Success] Renga UI should be under $(RENGA_UI_URL) and GitLab under $(GITLAB_URL)"
 	@echo
+	@echo "[Info] If running locally, make sure $(PLATFORM_DOMAIN) keycloak.$(PLATFORM_DOMAIN) and gitlab.$(PLATFORM_DOMAIN)"
```
jirikuncar: Can we generate a command to run only if the names are not defined in `/etc/hosts` already (e.g. `echo "127.0.0.1 $(PLAFORM_DOMAIN) ... >> /etc/hosts`)?

docker-compose.yml
```diff
@@ -101,7 +104,7 @@ services:
                 # CAREFUL: This must be accessible from inside the keycloak container
                 # for server-to-server communication.
                 # Use docker.for.mac.localhost as RENGA_ENDPOINT for example.
```
jirikuncar: Use renga.local as ...

Makefile
rokroskar: you mean just to show it in the info?

docker-compose.yml
```diff
@@ -101,7 +104,7 @@ services:
...
-                'site' => '${RENGA_ENDPOINT}/auth/',
+                'site' => 'http://keycloak.${PLATFORM_DOMAIN}:8080/auth',
```
jirikuncar: `http://keycloak.${PLATFORM_DOMAIN}/auth`

rokroskar: no keycloak exposes 8080 -- if you can find an option to change the port it listens on, I'm all for changing it to 80 :)

docker-compose.yml
```diff
@@ -120,7 +123,7 @@ services:
         gitlab_rails['initial_root_password'] = 'root_password'
         # Add any other gitlab.rb configuration here, each on its own line
     ports:
-      - '5080:5080'
+      - 80
       - '5443:443'
```
jirikuncar: we should make sure that the TSL termination is done on KeyCloak instead.

Makefile
jirikuncar: Yes. Also we should only show the info if the configuration is wrong (e.g. you can't ping/traceroute the name).

docker-compose.yml
rokroskar: I suggest to make an issue about the keycloak port and move on. 8080 is not ideal, but this is a detail in my view

rokroskar:
@jirikuncar the keycloak port issue is here: https://github.com/SwissDataScienceCenter/renga/issues/134

rokroskar:
@jirikuncar when you get a chance, can you test-run this setup?

PR #138, opened 2018-03-08
Use `hostname` for local deployment

leafty:
No need for `/etc/hosts` shenanigans anymore.

This now assumes that you don't mess with OS internals (hostname, gethostname, etc.), so that `hostname` resolves to the local loop or the host ip.


Makefile
```diff
@@ -186,12 +180,12 @@ services/gitlab/%:
 register-gitlab-oauth-applications: unregister-gitlab-oauth-applications
 	@$(DOCKER_COMPOSE_ENV) docker-compose exec gitlab /opt/gitlab/bin/gitlab-psql \
 		-h /var/opt/gitlab/postgresql -d gitlabhq_production \
-		-c "INSERT INTO oauth_applications (name, uid, scopes, redirect_uri, secret, trusted) VALUES ('renga-ui', 'renga-ui', 'api read_user', '$(RENGA_UI_URL)/login/redirect/gitlab http://localhost:3000/login/redirect/gitlab', 'no-secret-needed', 'true')" >& /dev/null
```
jirikuncar: We don't need standard output.

docker-compose.yml
```diff
@@ -56,15 +56,11 @@ services:
     restart: on-failure
     labels:
       traefik.enable: "true"
-      traefik.frontend.rule: "Host:keycloak.${PLATFORM_DOMAIN}"
```
jirikuncar: The whole idea is to have the development as close as possible to production, where the subdomains are preferable.

jirikuncar:
It basically reverts #133 which is not desirable at this moment.

rokroskar:
This is about host-based vs. path-based routing in the development environment. The bonus here is that we don't need to modify `/etc/hosts` -- but should we find that we want subdomains there won't be any way around it anyway. Discuss.

Makefile
leafty: It was broken on my machine (device not found)

rokroskar: https://unix.stackexchange.com/questions/407798/syntax-error-bad-fd-number ;)

leafty:
This going off he rails

PR #142, opened 2018-03-14
point to development first steps

rokroskar:


jirikuncar:
@rokroskar we should focus on making development branch merged to master since the `maint-0.1` was created by @erbou.

rokroskar:
Absolutely, but it should be done when the other repos are ready. Until then we should keep links consistent to avoid confusion.

PR #153, opened 2018-03-27
Add documentation for chart values

### LINKED ISSUE #151: Document chart values
Document chart values. Follow examples from stable like: https://github.com/kubernetes/charts/blob/master/stable/nginx-ingress/values.yaml

leafty:
Resolves #151 

~~The new file will replace `charts/renga/values.yaml` when it's done.~~


charts/renga/values.yaml
```diff
@@ -1,49 +1,269 @@
...
+## Keycloak configuration
+keycloak:
+  ## Keycloak admin user configuration
+  username: admin
+  password: admin
+
+  ## Keycloak image
+  # image:
+  #   pullPolicy: IfNotPresent
+  #   repository: jboss/keycloak
```
rokroskar: why is this commented out? presumably because it's set in the keycloak config -- it might be confusing to have it defined in two places?

charts/renga/values.yaml
```diff
@@ -1,49 +1,269 @@
...
+
+## PostgreSQL configuration
+postgresql:
+  ## Password for the user postgres (database admin user)
+  postgresPassword: postgres
+
+  ## Postgres image
+  # image:
+  #   pullPolicy: IfNotPresent
+  #   repository: postgres
```
rokroskar: same question as above re: commenting out the image definition

charts/renga/values.yaml
```diff
@@ -1,49 +1,269 @@
...
+## Gitlab configuration
+gitlab:
+  ## Password for the `root` user
+  password: gitlabadmin
+
+  ## Gitlab image
+  # image:
+  #   pullPolicy: IfNotPresent
+  #   repository: gitlab/gitlab-ce
+  #   tag: latest
```
rokroskar: should eventually pin the image tag

PR #173, opened 2018-04-16
notebook: fix readonly checkout

rokroskar:
This should fix the problems with readonly checkouts. 

@ciyer and/or @ableuler can you please test?


ciyer:
Works for me as well, now.

Issue #176, opened 2018-04-19
enable custom notebook images


rokroskar:
We want to be able to launch notebooks for custom environments so that users can have all the code dependencies taken care of. 

To-do: 

- [x] launch custom images based on project notebook naming scheme e.g. `<registry>/<namespace>/<project>/<env_slug>:<commit-sha>`
- [x] build images as part of the CI pipeline (#260)
- [x] add private registry credentials to pod (https://github.com/SwissDataScienceCenter/renku/issues/276)

leafty:
Reopening as #276 is still open

rokroskar:
closed via https://github.com/SwissDataScienceCenter/renku-notebooks/pull/21 and https://github.com/SwissDataScienceCenter/renku/pull/301

PR #177, opened 2018-04-19
jh: launch notebooks with custom image if present

rokroskar:
Modify the docker spawner to launch notebooks with a custom image. The image name/tag is based on the notebook URL naming scheme. The image has to be built manually for now, see the todo list in #176.

---
addresses #176

### DIFF
--- a/services/jupyterhub/spawners.py
+++ b/services/jupyterhub/spawners.py
@@ -18,8 +18,9 @@
 """Implement integration for using GitLab repositories."""
 
 import os
-
 from urllib.parse import urlsplit, urlunsplit
+
+import docker
 from tornado import gen, web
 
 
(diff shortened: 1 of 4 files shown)


services/jupyterhub/spawners.py
```diff
@@ -112,7 +112,19 @@ def start(self, *args, **kwargs):
...
+                     '/{namespace}'\
+                     '/{project}'\
+                     '/{environment_slug}'\
+                     ':{commit_sha}'.format(image_registry=os.getenv('IMAGE_REGISTRY'), **options)
+        self.log.info(self.image)
+
+        try:
+            result = yield super().start(*args, **kwargs)
+        except docker.errors.ImageNotFound:
+            self.image = 'rengahub/singeluser:development'
```
jirikuncar: we should probably log the exception so we know about it.

PR #187, opened 2018-04-26
Do not show the gitlab login page

### LINKED ISSUE #185: renga-only login page
Remove gitlab from the login page. 

See here: 

https://stackoverflow.com/questions/26618030/gitlab-disable-regular-username-password-login-and-only-allow-omniauth-login

and

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/723

Trouble is that you can no longer then log in as the root user.

rokroskar:
- [x] skip the gitlab login page 
- [x] create an admin/admin user
- [x] fix failing test
- [x] test with docker-compose 
- [ ] test with k8s chart

## Side effects

the `demo` user is made an admin. Could not create an additional `admin` user as this is a protected username in gitlab. 

## Possible bug

Not sure if this is related: cannot log out of gitlab until I delete the `keycloak.renga.build` application data. 

---
closes #185

### DIFF
--- a/tests/test_renga.py
+++ b/tests/test_renga.py
@@ -28,10 +28,6 @@ def test_renga_login(browser):
     url = os.getenv('RENGA_ENDPOINT', 'http://localhost')
     browser.visit(url)
 
-    assert browser.is_element_present_by_id(
-        'oauth-login-oauth2_generic', wait_time=10)
-    browser.find_by_id('oauth-login-oauth2_generic').click()
-
     assert browser.is_element_present_by_id('username', wait_time=60)
     browser.fill('username', 'demo')
     browser.fill('password', 'demo')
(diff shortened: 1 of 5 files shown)


rokroskar:
@jirikuncar I'm unable to get the keycloak image from https://github.com/SwissDataScienceCenter/docker-keycloak-postgres-proxy/pull/1 to work properly (fails to load theme) -- could you please have a look?

tests/test_renga.py
```diff
@@ -28,10 +28,6 @@ def test_renga_login(browser):
     url = os.getenv('RENGA_ENDPOINT', 'http://localhost')
     browser.visit(url)
 
-    assert browser.is_element_present_by_id(
```
jirikuncar: maybe we can change it to:

```python
if browser.is_element_present_by_id(...):
   browser....click()
```

PR #190, opened 2018-04-27
helm: move gitlab to a subdomain

jirikuncar:
Moves GitLab to a sub-domain and fixes all mentions of GitLab URL.

---
It can be merged directly.  cc @rokroskar


jirikuncar:
@rokroskar this PR is not about `Auto-DevOps.yml`. It's about general strategy that keeps 3rd party services on separate sub-domains starting with GitLab. We should discuss `Auto-DevOps.yml` vs. `.gitlab-ci.yml` separately.

charts/renga/charts/gitlab/templates/_gitlab.rb.tpl
```diff
@@ -13,7 +13,7 @@ GitLab Omnibus configuration
 ##! URL on which GitLab will be reachable.
 ##! For more details on configuring external_url see:
 ##! https://docs.gitlab.com/omnibus/settings/configuration.html#configuring-the-external-url-for-gitlab
-external_url '{{ template "http" . }}://{{ .Values.global.renga.domain }}/gitlab'
+external_url '{{ template "http" . }}://gitlab.{{ .Values.global.renga.domain }}'
```
rokroskar: could this just be `.Values.global.renga.gitlab_url`? Then it's just up to the ingress spec and whoever is running the chart can decide where they want the services to run.

charts/renga/charts/gitlab/templates/deployment.yaml
```diff
@@ -59,7 +59,7 @@ spec:
           containerPort: 8105
         livenessProbe:
           httpGet:
-            path: /gitlab/help
```
rokroskar: could we template this? Or use an environment variable for the prefix? That would allow someone to run it on a path instead of subdomain if they wanted to.

charts/renga/charts/gitlab/templates/deployment.yaml
```diff
@@ -70,7 +70,7 @@ spec:
           failureThreshold: 10
         readinessProbe:
           httpGet:
-            path: /gitlab/help
```
rokroskar: same as above

charts/renga/charts/jupyterhub/templates/hub/deployment.yaml
```diff
@@ -164,7 +164,7 @@ spec:
           value: {{ $value | quote }}
         {{ end -}}
         - name: GITLAB_HOST
-          value: {{ template "http" . }}://{{ .Values.global.renga.domain }}/gitlab
+          value: {{ template "http" . }}://gitlab.{{ .Values.global.renga.domain }}
```
rokroskar: this should be specified as an `extraEnv` and not built in to the chart (#216 implements this change)

charts/minikube-values.yaml
```diff
@@ -2,13 +2,17 @@
...
+    domain: renga-k8s.build
+
 ingress:
   enabled: true
   annotations:
     kubernetes.io/ingress.class: nginx
     nginx.ingress.kubernetes.io/ssl-redirect: false
   hosts:
-  - ""
+  - "renga-k8s.build"
```
leafty: missing the new host?

jirikuncar: @leafty I don't understand the question. Can you please clarify it?

leafty: i'm expecting the ingress to reject HTTP requests for host `gitlab.renga-k8s.build` (!= `renga-k8s.build`)

jirikuncar: Even with line https://github.com/SwissDataScienceCenter/renga/pull/190/files#diff-36d359d0ba9a7120256f9209a50eb5faR50 ?

leafty: ok. but then the value is also used for the TLS config, with auto LE with `cert-manager`

charts/README.rst
```diff
@@ -9,9 +9,34 @@ Testing locally
...
+
+We can now disable kube-dns (this is necessary, otherwise it will still continue to receive some DNS requests):
+
+.. code-block:: console
+
+    $ minikube addons disable kube-dns
+    $ kubectl -n kube-system delete deploy kube-dns
+
+The command :code:`minikube addons disable kube-dns` may fail, so to mark the
+addon as diabled, edit `~/.minikube/config/config.json` and insert :code:`"kube-dns": false`.
```
rokroskar: diabled --> disabled

charts/README.rst
```diff
@@ -27,12 +52,11 @@ Requires minikube, kubectl, helm and python.
     $ helm upgrade renku --install \
         --namespace renku \
         -f minikube-values.yaml \
-        --set global.renku.domain=$(minikube ip) \
-        --set ui.gitlabUrl=http://$(minikube ip)/gitlab \
-        --set ui.jupyterhubUrl=http://$(minikube ip)/jupyterhub \
-        --set jupyterhub.hub.extraEnv.GITLAB_HOST=http://$(minikube ip)/gitlab \
         ./renku
 
+Make sure you have `$(minikube ip) renku-k8s gitlab.renku-k8s` line
```
rokroskar: --> renku-k8s.build gitlab.renku-k8s.build

charts/minikube-values.yaml
```diff
@@ -2,34 +2,40 @@
...
 
-global:
-  gitlab:
-    urlPrefix: /gitlab
-
 gitlab:
   sshPort: 2222
 
+ui:
+  gitlabUrl: gitlab.renku-k8s.build
```
rokroskar: http?

leafty:
> The job exceeded the maximum time limit for jobs, and has been terminated.

:disappointed:

rokroskar:
Looks like docker pull stalled?

rokroskar:
The issue with the coredns setup is that it will be impossible to run this in a CI test. We're using traefik to expose the correct backends on the correct (fake) local hostnames -- could it help here as well?

rokroskar:
I still get a 500 from gitlab when logging in:

```
Faraday::ConnectionFailed (Failed to open TCP connection to renku-k8s.build:80 (getaddrinfo: Name or service not known)):
  lib/gitlab/middleware/multipart.rb:95:in `call'
  lib/gitlab/request_profiler/middleware.rb:14:in `call'
  lib/gitlab/middleware/go.rb:17:in `call'
  lib/gitlab/etag_caching/middleware.rb:11:in `call'
  lib/gitlab/middleware/read_only/controller.rb:28:in `call'
  lib/gitlab/middleware/read_only.rb:16:in `call'
  lib/gitlab/request_context.rb:18:in `call'
  lib/gitlab/metrics/requests_rack_middleware.rb:27:in `call'
  lib/gitlab/middleware/release_env.rb:10:in `call'
```

fgeorgatos:
on my machine, after some trial'n'error, this one is working now:

```
        rewrite name regex (.*)renku-k8s.build renku-k8s.build
        hosts {
            192.168.99.100 renku-k8s.build
            fallthrough
        }
```

@rokroskar , @leafty :  comments? flames?

rokroskar:
almost there, thanks @fgeorgatos for the coredns fix. Now I get an "invalid redirect url" when logging out of gitlab.

rokroskar:
I had to add `gitlab.renku-k8s.build` to the valid redirect URLs in keycloak client settings for gitlab in order to make logout work. No such extra setting is required normally -- any idea why this might be the case here?

leafty:
@rokroskar could be the realm template needs to be adjusted. Which OIDC client was it?

rokroskar:
@leafty it's the gitlab client. But the point is that the other configs we have (e.g. using our `Makefile` + `docker-compose` which also uses subdomains) don't require this change so I'm wondering why it's the case here.

leafty:
@rokroskar did you check after a `helm del --purge renku`?

Could be the keycloak db was populated from before the PR.

From the code, the realm should be initialized properly: https://github.com/jirikuncar/renga/blob/gitlab-sub-domain/charts/renku/charts/keycloak/templates/_renku-realm.tpl#L496-L498

rokroskar:
@leafty yep, that is in there:

image

Have you tried the logout?

rokroskar:
another host resolution issue -- this time from the jupyterhub spawner: 

```
[E 2018-06-13 11:40:52.497 JupyterHub spawners:102] HTTPConnectionPool(host='gitlab.renku-k8s.build', port=80): Max retries exceeded with url: /api/v4/projects/demo%2Ftest (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f8e2c018e48>: Failed to establish a new connection: [Errno -5] No address associated with hostname',))
```

leafty:
@rokroskar confirmed that the redirect uri for the gitlab client is too specific.

rokroskar:
I'm not sure we are going to do this anytime soon. In k8s, the routing is up to the site-specific ingress anyway so I'm not too concerned. In minikube, the setup seems to be quite annoying to get right, unless we do something like what is suggested for the gitlab helm chart: https://gitlab.com/charts/gitlab/blob/master/doc/minikube/README.md#handling-dns

If no one wants to take this PR under their wing and do the work of merging all the changes with current master, I suggest we close. @leafty @jirikuncar what say you?

leafty:
@rokroskar The approach I would favor is to to have a config value switch between host based and path based routing (so having two ingress templates, similar to nginx-ingress with deployment vs deamonset type https://github.com/helm/charts/tree/master/stable/nginx-ingress/templates).

Using the nip.io trick would be useful to validate host based routing when developing locally.

rokroskar:
@leafty that would sound like the best of both worlds. Having no way to test the host-based setup is a bit of a problem atm.

jirikuncar:
This PR is quite outdated.

PR #193, opened 2018-05-02
Fix the jupyterhub dockerfile after the 0.9 release of jupyterhub

ableuler:


rokroskar:
@jirikuncar @ableuler did `jupyterhub.renga.build/services/projects` work for you? It doesn't work for me -- neither does `jupyterhub.renga.build` for that matter (bad gateway).

Issue #194, opened 2018-05-05
jupyterhub 0.9 incompatible with current setup


rokroskar:
#193 breaks the jupyterhub traefik configuration, results in "bad gateway". Going to commit a15c8742d5e73759e3752921624994cb50e228e8 and using `jupyterhub/jupyterhub:0.8.1` works. We should either revert #0.8.1 or resolve the proxy issue.

erbou:
I can reproduce the issue on development branch 17cae25
make wipe
PLATFORM_VERSION=development make start
waiting .... [OK]
jupyterhub.renga.build results in "bad gateway"

jirikuncar:
```
ARG JUPYTERHUB_VERSION=master
FROM jupyterhub/jupyterhub:$JUPYTERHUB_VERSION
```

this variable should be set in the Makefile and a target for jupyterhub in the Makefile should use the --build-arg flag to override the default.

PR #198, opened 2018-05-07
Jupyterhub 0.8.1

jirikuncar:
addresses #194

### DIFF
--- a/services/jupyterhub/Dockerfile
+++ b/services/jupyterhub/Dockerfile
@@ -1,4 +1,4 @@
-FROM jupyterhub/jupyterhub:latest
+FROM jupyterhub/jupyterhub:0.8.1
 
 RUN curl -fsSL get.docker.com | sh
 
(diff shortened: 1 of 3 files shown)


services/jupyterhub/Dockerfile
```diff
@@ -1,13 +1,6 @@
-FROM jupyterhub/jupyterhub:0.9
+FROM jupyterhub/jupyterhub:0.8.1
```
rokroskar: how about:

```
ARG JUPYTERHUB_VERSION=master
FROM jupyterhub/jupyterhub:$JUPYTERHUB_VERSION
```

this variable should be set in the `Makefile` and a target for jupyterhub in the `Makefile` should use the `--build-arg` flag to override the default.

services/notebook/Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM $BASE_IMAGE
 MAINTAINER Swiss Data Science Center <info@datascience.ch>
 
 ADD install /tmp/install
-ARG JUPYTERHUB_VERSION=master
+ARG JUPYTERHUB_VERSION=0.8.1
```
rokroskar: same as above in `services/jupyterhub/Dockerfile`

PR #202, opened 2018-05-07
Gitlab notebook url

jirikuncar:


rokroskar:
waiting for @leafty to test with #201 before merging

jirikuncar:
IMHO the environment API is a good intermediate layer so we have a flexibility to change the final notebook URL. There are many potentially moving parts - `jupyterhub` subdomain vs. path prefix, service name, etc. Also note that UI only depends on GitLab for now. If we want to avoid using environment API, we should add an explicit configuration option for JupyterHub to UI.

rokroskar:
@jirikuncar I think it needs to be the latter (a Jupyterhub path option to the UI) -- as it is now, you can't launch a notebook if you are attaching to e.g. a gitlab that we don't control. If you construct the paths independently of the gitlab CI environment, this would work out of the box.

PR #204, opened 2018-05-07
use V1Container model

### LINKED ISSUE #203: init container should use the kubernetes V1Container
for consistency, use the `V1Container` model instead of a dict

rokroskar:
closes #203 

contingent on #201

### DIFF
--- a/services/jupyterhub/spawners.py
+++ b/services/jupyterhub/spawners.py
@@ -22,6 +22,7 @@ import os
 from urllib.parse import urlsplit, urlunsplit
 
 import docker
+from kubernetes import client
 from tornado import gen, web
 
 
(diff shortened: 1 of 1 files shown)


services/jupyterhub/spawners.py
```diff
@@ -263,26 +264,31 @@ def get_pod_manifest(self):
...
+                    **{
+                        'name':
+                            container_name,
+                        'image':
+                            'alpine/git',
+                        'command': ['sh', '-c'],
+                        'args': [
+                            'git clone {repository} {mount_path} && '
+                            '(git checkout {branch} || '
+                            'git checkout -b {branch}) && '
```
leafty: you should keep the ||'ed commands on one line for clarity

rokroskar: Ideally, yes but it makes the line too long

leafty: Indent then

PR #211, opened 2018-05-09
charts: added chartpress

### LINKED ISSUE #169: build images on travis
Our custom images used by the charts and docker-compose should be available from dockerhub (i.e. built on . The custom configs should be baked into the images and not mounted. 

- [x] jupyterhub
- [ ] notebooks service
- [x] apispec

jirikuncar:
see https://github.com/jupyterhub/chartpress/blob/master/README.md

The chartpress functionality reduces the overhead in handling chart and image inter-dependencies. This makes it easier to test various development versions. 

Charts are published at http://swissdatasciencecenter.github.io/helm-charts and can be installed with helm as standard dependencies (see the `README` in this PR). The images and charts are built and published with travis-ci. 

This functionality is already being used with the `renga-ui` repository as a test case.

The chartpress spec now includes all the images that are needed by the chart *and* by docker-compose -- as a result, it closes #169.


charts/renku/values.yaml
```diff
@@ -291,14 +291,15 @@ jupyterhub:
     image:
       name: renku/jupyterhub-k8s
       tag: latest
+      repository: renku/jupyterhub-k8s
```
leafty: Choose `name` or `repository`.

charts/renku/values.yaml
```diff
@@ -346,9 +348,24 @@ jupyterhub:
   auth:
     type: gitlab
     gitlab:
-      clientId: "jupyterhub"
-      clientSecret: "no-secret-needed"
+      clientId: jupyterhub
+      clientSecret: no-secret-needed
```
leafty: Can we use something a bit less confusing like `replace-me` or `dummy-secret`?

rokroskar:
@leafty fixed the remaining issues, I think it's good to go

jirikuncar:
@rokroskar thank you for the improvements and fixes. It looks like it's good to go. :shipit:

rokroskar:
@jirikuncar one (minor) sticky point that I'm trying to work out is a response to @leafty's request to keep either `name` or `repository` in the image spec but not both. To me it seems like docker itself is very vague on what constitutes a name -- i.e. is a **name** the `<repository>:<tag>` combination? Or is `<repository> == <name>`? The default templates created by helm specify images as `{{ .Values.image.repository }}:{{ .Values.image.tag }}`. But chartpress seems to think `repository` and `name` are the same thing and if you look at the jupyterhub charts they use `{{ .Values.image.name }}:{{ .Values.image.tag }}` which is the "convention" I followed in the notebooks service. Point is, even if I don't include `repository` in the image spec, `chartpress` adds it back in.

jirikuncar:
@rokroskar the `name` vs. `repository` is part of `chartpress`. They fill both `name` and `repository` due to some legacy charts that still need `repository`.

rokroskar:
@jirikuncar ok so we should default to `repository`

jirikuncar:
I have created a PR for chartpress https://github.com/jupyterhub/chartpress/pull/11. Can we test the branch here?

PR #212, opened 2018-05-09
Add gitlab-admin to keycloak

ableuler:
Add a gitlab-admin user (gitlab-admin / gitlab-admin) to the keycloak realm, make this user a gitlab admin instead of the demo user.


rokroskar:
This also seems to break gitlab logout -- not sure what changed with the client setup?

ableuler:
I plan to revisit this when tackling https://github.com/SwissDataScienceCenter/renga/issues/214, let's leave the PR open until then.

rokroskar:
I think with #264 closed we can close this PR?

ableuler:
@rokroskar agree

PR #213, opened 2018-05-09
Helm tests

rokroskar:
Initial implementation of a helm test that allows us to run 

```
helm test renga
```

At the moment, the test is identical to the normal test we run in travis, but perhaps this should be changed? I don't think the helm test should really test the docs as well, for example, and there may be some k8s-specific tests in the future. Opinions on this are welcome. 

 ---

addresses https://github.com/SwissDataScienceCenter/renga/issues/195

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,8 +187,12 @@ $(dockerfile-services): %: .env services/%/Dockerfile
 jupyterhub-k8s: .env services/jupyterhub/jupyterhub-k8s.Dockerfile
 	docker build --tag $(DOCKER_REPOSITORY)$@:$(PLATFORM_VERSION) -f services/jupyterhub/$@.Dockerfile services/jupyterhub/
 
+.PHONY: renga-tests
+renga-tests:
+	docker build --tag $(DOCKER_REPOSITORY)$@:$(PLATFORM_VERSION) -f tests.Dockerfile .
+
 .PHONY: tag
-tag: $(dockerfile-services) jupyterhub-k8s
+tag: $(dockerfile-services) jupyterhub-k8s renga-tests
 
 # build docker images from makefiles
 .PHONY: $(makefile-services)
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -187,8 +187,12 @@ $(dockerfile-services): %: .env services/%/Dockerfile
 jupyterhub-k8s: .env services/jupyterhub/jupyterhub-k8s.Dockerfile
 	docker build --tag $(DOCKER_REPOSITORY)$@:$(PLATFORM_VERSION) -f services/jupyterhub/$@.Dockerfile services/jupyterhub/
 
+.PHONY: renga-tests
+renga-tests:
+	docker build --tag $(DOCKER_REPOSITORY)$@:$(PLATFORM_VERSION) -f tests.Dockerfile .
```
jirikuncar: we should add it to chartpress too.

rokroskar: eventually

tests/conftest.py
```diff
@@ -17,6 +17,15 @@
...
 
+import pytest
+import splinter
 from dotenv import find_dotenv, load_dotenv
 
 load_dotenv(find_dotenv())
+
+
+@pytest.fixture()
+def browser():
```
jirikuncar: Do you plan to use it in this PR?

rokroskar: it's used in the test. I was using the pytest-splinter extension before but running into problems so I defined my own browser fixture here. The name stays the same.

tests.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    mv geckodriver /usr/local/bin && \
+    apt-get update -y && \
+    apt-get install -y firefox-esr xvfb enchant shellcheck
+
+COPY . /code
+
+WORKDIR /code
+RUN pip install -r tests/requirements.txt
+RUN pip install -r docs/requirements.txt
+CMD xvfb-run scripts/run-tests.sh
```
leafty: Can you make it so it runs only things from `tests`? There's no point in testing the content of the documentation nor do a spell check in there.

rokroskar: done in e0c503a

Issue #186, opened 2018-04-26
improving the "first-steps" experience


rokroskar:
1) Prerequesistes 
  - [x] prerequisites need to be updated in the doc: (done in #251)
    - docker (tested with 18.03.0), must be running
    - sbt (I don't think this is needed anymore with container-only build?)
    - python3.6 (3.6.4 Anaconda, Inc)
    - pip (10.0.1)
2) System configuration
    - [x] document the `/etc/hosts` requirement (or just add this line to the `Makefile`?
    `grep -qi 'renga.build' /etc/hosts || sudo echo "127.0.0.1       keycloak.renga.build gitlab.renga.build renga.build jupyterhub.renga.build" >> /etc/hosts` (see https://github.com/SwissDataScienceCenter/renga/pull/174)

3) Wipe clean previous renga and all docker images
  - [x] document `make wipe` (done in #225)

5) Start renga
  - [x] document the use of `PLATFORM_VERSION` -- it must be set if the renga repo is on a branch that is different from the other component branches (done in #225)
  - [x] remove this message from `Makefile` (not needed anymore):
    "[Info] To make notebooks available as deployed environments, set the
            RENGA_NOTEBOOK_TOKEN and RENGA_REVIEW_DOMAIN CI variables in gitlab project settings."
        is not very helpful ... (done in #225)
  - [x] include pulling of alpine/git:latest (#207)

7) Renga user first steps:
 CLI:
    - [x] provide details (doc or online help) about --force https://github.com/SwissDataScienceCenter/renku-python/issues/199
    - [x] provide a more helpful message when the repo is dirty (fixed in https://github.com/SwissDataScienceCenter/renga-python/pull/160; also partially addressed by https://github.com/SwissDataScienceCenter/renga-python/issues/158)
    - [x] document the CLI (being done in https://github.com/SwissDataScienceCenter/renku-python/pull/191)
    ** ISSUE: could it be a problem that the commit message shows the full pathname of the command on the user's system? I can see the pros (e.g. better assist the user) and cons of having it vs not having it. However the cons may outweight the pros: this info will be shared and people may have an issue with that. --> actually I don't think this is a problem and the information may be critical for debugging problems with result reproducibility
     ~~simplify the commit message for `renga run`~~ (also see https://github.com/SwissDataScienceCenter/renga-python/issues/156)

```
$ renga dataset create dataset1
[...]

Issue #149, opened 2018-03-21
Improve readability of the Makefile


leafty:
Whenever time allows, render the makefile easier to read.

rokroskar:
closed via #225

PR #225, opened 2018-05-16
Makefile refactor

### LINKED ISSUE #206: Docker-compose trying to pull non-existant tag for jupyterhub
Command "PLATFORM_VERSION=development docker-compose pull"
 
Results in 
"Pulling singleuser (rengahub/singleuser:development)...
ERROR: manifest for rengahub/singleuser:development not found"

After which pull does not continue.

rokroskar:
Major change is the removal of deployment-related pieces from the `Makefile`. These are moved to two scripts in the `scripts` directory instead. Documentation (READMEs and the docs) has been adapted accordingly. 

addresses #149 #186 
closes #206


leafty:
the way this works now, i would also remove the `.env` target

rokroskar:
@leafty you still need some way to remember those environment variables

rokroskar:
However, I might soon advocate for removing the docker-compose and only using helm. It's getting pretty ridiculous trying to maintain the two set up and helm seems now less error prone.

leafty:
maybe i'm wrong, but i thought only the `$PLATFORM_VERSION` was used for builds. If that's correct, it's way simpler to have it as a regular env var.

the other vars are part of the start script for me, so that's why i asked about moving the target to some part of the script.

rokroskar:
@leafty if you run the script twice and you change the token values between runs, the registration/unregistration of various pieces will fail. That's why we used the `.env` file in the first place.

leafty:
not denying that. I'm saying that creating/managing the `.env` file should happen in the scripts, not in the Makefile.

jirikuncar:
> I'm saying that creating/managing the `.env` file should happen in the scripts, not in the Makefile.

The `.env` file is a generated file, hence it's a make target.

erbou:
I could test the branch and run the demo. In the make start I got the error:

Checking that services are reachable...
make: *** [start] Error 2

That's because *.renga.build did not respond to the ping command at the end of the renga-start.sh. Curl is more heavy handed than ping, but since it works we should use that.

There is an issue in renga-wipe.sh, where the script continues even if .env does not exist.

rokroskar:
I don't understand, so ping fails somehow but curl works?

erbou:
The local host does not respond to ICMP, the ping command hangs and timeouts with 100% packet loss:

>>>> ping -c1 renga.build
PING keycloak.renga.build (127.0.0.1): 56 data bytes
--- keycloak.renga.build ping statistics ---
1 packets transmitted, 0 packets received, 100.0% packet loss

rokroskar:
is this something you have specifically disabled? or is this normal? I've never seen that on my machine.

erbou:
The firewall on my macbook is on (always). By default it does not filter TCP/UDP connections coming from my localhost, but it does filter ICMP. It is not unusual for IT departments to check or impose that their employees have their Firewall enabled, and it wouldn't hurt to support that.

rokroskar:
Yep makes sense will fix it

Makefile
```diff
@@ -132,10 +197,12 @@ scala-artifact = \
 	renga-graph
```
leafty: TODO: remove the scala artifacts and update readme to remove sbt as a req.

NB: `renku-storage` cab be built with docker only. Also, cascading changes from `renku-commons` is not possible from the docker build so building it here has no value.

rokroskar:
@leafty an error in the singleuser base image crept in while I was merging in the changes from master -- fixed now, can you please reapprove?

PR #236, opened 2018-05-24
global: GITLAB_HOST -> GITLAB_URL

jirikuncar:


jirikuncar:
@rokroskar I don't think so. It's only to keep the naming consistent.

PR #241, opened 2018-05-25
fix deployment with gitlab runner dependency

rokroskar:


leafty:
Thanks! I noticed it too, but you were faster than me ;)

rokroskar:
couldn't help it, failed travis builds are like needles for my eyes... ;)

PR #244, opened 2018-05-28
Fix post-install hook

### LINKED ISSUE #243: Do not remove post-install-job on failure
Keep it, as it's pretty hard to debug when the logs vasnish:

https://github.com/SwissDataScienceCenter/renku/blob/4d4d1c888e8b64b3635a35b21a447ba8fb0211ba/charts/renku/templates/post-install-job.yaml#L12

leafty:
Closes #243

### DIFF
--- a/charts/renku/templates/post-install-job.yaml
+++ b/charts/renku/templates/post-install-job.yaml
@@ -9,7 +9,7 @@ metadata:
   annotations:
     "helm.sh/hook": post-install,post-upgrade
     "helm.sh/hook-weight": "5"
-    "helm.sh/hook-delete-policy": "hook-succeeded,before-hook-creation"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
 spec:
   template:
     metadata:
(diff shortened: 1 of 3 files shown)


charts/renku/templates/post-install-job.yaml
```diff
@@ -9,8 +9,7 @@ metadata:
   annotations:
     "helm.sh/hook": post-install,post-upgrade
     "helm.sh/hook-weight": "5"
-    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
-    # FIXME with newer version of helm "before-hook-creation"
+    "helm.sh/hook-delete-policy": "hook-succeeded,before-hook-creation"
```
jirikuncar: I would personally keep the job there after success too and go with a single value `"before-hook-creation"`.

---
https://github.com/kubernetes/helm/blob/master/docs/charts_hooks.md#automatically-delete-hook-from-previous-release

PR #246, opened 2018-05-28
Enable registry

### LINKED ISSUE #238: set up the gitlab image registry on helm deployments
Configure the `registry_external_url` option in the omnibus config and expose the correct port via the ingress. 

addresses #176

leafty:
Fix #238

~~I can't make it work :cry: . I always get 405'd by nginx.~~

~~It works on minikube. To have it work on cluster deployment, it would be preferable to wait after #190.~~


rokroskar:
@leafty what's missing here still? I can't recall if it's been tested on minikube.

rokroskar:
Is there any way to get this to work without the insecure registry workaround?

leafty:
@rok Injecting a root cert bundle into multiple places? Sounds more or less the same poison to me.

rokroskar:
it's not for security's sake, it's for ease-of-use in development. I'm running minikube on hyperkit and I'm not entirely sure where to add the insecure registry. For dind, every docker host needs the insecure docker registry set.

leafty:
Every host with a docker daemon which wants to use the registry, basically. Injecting self-signed certificates would happen on the same set of hosts too.

rokroskar:
so how does the gitlab omnibus chart get around this? they enable the https registry by default.

rokroskar:
@leafty can you provide a full example of how you got the registry to work on minikube?

leafty:
As for the omnibus chart:

> Prerequisites 
> [...]
> A wildcard DNS entry, which resolves to the external IP address
> [...]
> To support the GitLab services and dynamic environments, a wildcard DNS entry is required which resolves to the Load Balancer or External IP. Configuration of the DNS entry will depend upon the DNS service being used.
> [...]
> SSL certificates are automatically created and configured by kube-lego.

My guess is that you cannot make it work on minikube...

Issue #195, opened 2018-05-05
create integration test using renga-demo


rokroskar:
Now we have a nice script that creates a non-trivial project from scratch. We should use this for an integration test to avoid e.g. #194 and #199

3C111:
Make a test that will touch all the components to make sure they all work together as expected 
[] run the `renga-demo` script on a clean install of the platform 

[] "click" around like it's done in `test_renga.py`

Issue #223, opened 2018-05-16
make CI explicit


rokroskar:
Recently we moved the CI to server-wide default AutoDevOps -- this is nice for automated functionality but bad for keeping things explicit and transparent. We should move back to running the CI from a `.gitlab-ci.yml` file located in the project repository. We should then also, in various places like the UI and the CLI, compare the version in the user's repo to the currently published version. If they differ, we should give the user instructions on how to update. 

- [x] move CI back to `.gitlab-ci.yml` in the repo (https://github.com/SwissDataScienceCenter/renku-python/issues/174)
~~ configure server-wide the location to be `.renga/gitlab-ci.yml`~~ we won't do this now -- requires more customization of gitlab which we are trying to avoid
~~provide checks for outdated CI~~ this can come later as needed. Potentially needs to involve the API gateway.

PR #247, opened 2018-05-28
Integration test

rokroskar:
This adds clicking through to the "Launch Notebook" button to the test.

Note: this will probably continue failing until at least the first part of #223 is fixed. 

closes https://github.com/SwissDataScienceCenter/renku/issues/195


rokroskar:
bleh, it passed earlier -- maybe need to put some timeouts somewhere.

PR #249, opened 2018-05-29
Specify minikube memory in README.rst

ableuler:
Recommend to run minikube with 6 GB of memory. The default of 2 GB (v0.27.0) doesn't seem to be nearly enough for renku.


rokroskar:
I like to make it an even 7000, but not all that important :)

PR #254, opened 2018-05-30
Refactor notebooks and jupyterhub

rokroskar:
This PR removes the notebooks service from the renku repo and uses it instead as a dependency to the helm chart. It also removes the jupyterhub chart and uses JH instead as a dependency as well. 

These changes will allow us to: 
1. have standalone development and tests for the notebooks service 
2. more easily stay on top of changes/improvements to JH in the future


leafty:
I just did:
```bash
git clone https://github.com/jupyterhub/zero-to-jupyterhub-k8s.git
cd zero-to-jupyterhub-k8s
git checkout -b verify v0.6
rsync -av <path-to-renku>/charts/jupyterhub .
```
And verified the diff.

Looks like the differences are minimal, so moving to a helm dependency should be easy. :+1: 

```
--- a/jupyterhub/templates/hub/configmap.yaml
+++ b/jupyterhub/templates/hub/configmap.yaml
@@ -180,7 +180,7 @@ data:
   {{ else if eq .Values.hub.db.type "sqlite-memory" -}}
   hub.db_url: "sqlite://"
   {{ else -}}
-  hub.db_url: {{ .Values.hub.db.url | quote }}
+  hub.db_url: {{ template "hub.db_url" . }}
   {{- end }}
   auth.admin.access: {{ .Values.auth.admin.access | quote }}
   {{ if .Values.auth.admin.users -}}

--- a/jupyterhub/templates/hub/deployment.yaml
+++ b/jupyterhub/templates/hub/deployment.yaml
@@ -80,6 +80,7 @@ spec:
         image: {{ .Values.hub.image.name }}:{{ .Values.hub.image.tag }}
         command:
           - jupyterhub
+          - --debug
           - --config
           - /srv/jupyterhub_config.py
           # We want to do automatic upgrades for sqlite-pvc by default, but allow users

--- /dev/null
+++ b/jupyterhub/templates/hub/_helpers.tpl
@@ -0,0 +1,8 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+hub db url
+*/}}
+{{- define "hub.db_url" -}}
+{{- $hostname := include "postgresql.fullname" . -}}
+{{- printf "postgres+psycopg2://%s:%s@%s:5432/%s" .Values.global.jupyterhub.postgresUser .Values.global.jupyterhub.postgresPassword $hostname .Values.global.jupyterhub.postgresDatabase | quote -}}
+{{- end -}}
```

rokroskar:
@leafty this is already done in 880a881

PR #256, opened 2018-05-31
Fix an erronous reference to 'renkuhub' for the jupyterhub image

ableuler:


rokroskar:
I'm just really confused how no one noticed this for over a week. `make start` should fail...

ableuler:
It doesn’t fail. There’s still the “build” line next to the “image” line - was actually wondering if that was intentional...

PR #264, opened 2018-06-05
demo user not gitlab admin

### LINKED ISSUE #242: Remove or make optional demo user as gitlab admin
User `demo` should not be a gitlab admin:
https://github.com/SwissDataScienceCenter/renku/blob/4d4d1c888e8b64b3635a35b21a447ba8fb0211ba/charts/renku/templates/configmap.yaml#L42-L43

leafty:
Fix #242

### DIFF
--- a/charts/minikube-values.yaml
+++ b/charts/minikube-values.yaml
@@ -32,6 +32,7 @@ global:
 
 gitlab:
   sshPort: 2222
+  demoUserIsAdmin: true
 
 notebooks:
   jupyterhub_base_url: /jupyterhub/
--- a/charts/renku/templates/configmap.yaml
+++ b/charts/renku/templates/configmap.yaml
@@ -56,7 +56,7 @@ data:
         "provider": "oauth2_generic",
         "skip_confirmation": "true",
         "reset_password": "true",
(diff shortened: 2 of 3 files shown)


rokroskar:
Making the demo user an admin was not an accident -- but until we have another admin user (e.g. gitlab-admin?) this is the only way to not get locked out of gitlab if we also want to not go through the gitlab login page. So yes, I agree `demo` should not be an admin user, but this needs to be done by first completing https://github.com/SwissDataScienceCenter/renku/pull/212. I suggest we close this PR and make the change in #212 instead.

leafty:
Why would you be locked out? I believe you can register on keycloak to create new accounts and the admin token is linked to user `root` so will continue to work.

rokroskar:
Might work, didn't try that

ableuler:
You can do what @leafty described, but it would be nice to have this automated rather than having to create an new user through kc and make it admin through the gitlab api every time you restart the platform.

leafty:
@ableuler Made it optional with sensible default.
Please merge quickly, as I don't like having demo being put back as admin with every CD update.

PR #267, opened 2018-06-13
Fix #259 (kubespawner)

### LINKED ISSUE #259: [notebooks] spawners should handle LFS objects
Using the `renku-demo` script and actually running the notebook gives this: 

image

leafty:


rokroskar:
@leafty tested?

leafty:
Nope :D

leafty:
@rok if the only required actions is to install git-lfs, I think we can try to make a PR for https://github.com/alpine-docker to have an auto build of something like `alpine/git:lfs`

rokroskar:
@leafty good idea -- it's possible it already exists somewhere, but I was not able to find it.

Issue #214, opened 2018-05-10
keycloak: use API to create new clients and users


jirikuncar:
We should consider using REST API to create clients and users in KeyCloak db instead of templated config dump.

ableuler:
Agree - I actually started looking into this already yesterday.

jirikuncar:
Hopefully, we can also switch to the `incubator/keycloak` chart.

rokroskar:
@ableuler is there a branch with your work on this somewhere already? @fgeorgatos started to work on it as well, would be a shame to duplicate efforts.

ableuler:
No branch, what I did related related to this issue can be summarised as followed:

- I looked at how to use the admin credentials to obtain a temporarily (only 1 minute or so...) valid access token with admin privileges for keycloak which can then be used to create the actual users, clients, etc (see https://github.com/SwissDataScienceCenter/renku-demo/blob/master/steps/2-create-users.py) for an example.
- I went through the API documentation to realise that the realm itself can not be created through the API, meaning that the empty Renku realm should still be created from the template. However, this template is not going to change much (if at all) when users and clients can be added through an api.

Hope this helps, great if @fgeorgatos can move forward on this issue 👍

leafty:
Forgot about the existence of this issue.

I believe everything can be manipulated by the API, the thing is that the API itself is not documented (while all the actions from the admin console are well described).

Scripting the API calls can be made in bash with the use of the keycloak admin cli. I tried a few examples with @3C111 which you can find here: https://github.com/SwissDataScienceCenter/incubator-kc-abac

fgeorgatos:
ugh, perhaps it should be me perhaps the person being asked at, but: 
* what is the status on this issue?
(somehow, when #268 was closed, it fall completely out of my view)

rokroskar:
status is "open" :)

Issue #208, opened 2018-05-08
refactor charts to use dependencies when possible


rokroskar:
In addition, do not modify the gitlab database directly where it is not necessary, e.g. for application configuration options, you can use the REST API (see #187).

- [x] postgres (#293)
- [x] keycloak (https://github.com/SwissDataScienceCenter/renku/pull/336)
- [x] redis (#337)

~~We should include the gitlab configuration in the image, which should allow us to use the stock gitlab chart as a helm dependency directly.~~ 

~~instead, use an omnibus [gitlab chart](https://docs.gitlab.com/ee/install/kubernetes/gitlab_omnibus.html)?~~ the omnibus chart leaves a lot to be desired, see comments below.

leafty:
https://gitlab.com/charts/gitlab-omnibus chart does not integrate with existing ingress, so I would wait until the native cloud one is in a usable form.

rokroskar:
@leafty makes sense -- though I think this issue can be closed as I don't think we use init containers anywhere anymore?

leafty:
I was thinking we could rephrase it a bit...
I think we can use the same idea to make postgresql a chart dependency, with a job doing the database init.

rokroskar:
yes, exactly. We can keep this open and increase the scope. I guess postgres is the biggest one because of the the db init steps. Redis can easily be a dependency. Once we write the renku realm as REST calls, then keycloak can also follow.

rokroskar:
closed by #337

PR #268, opened 2018-06-13
[WIP] add keycloak as explicit dependency

fgeorgatos:
initiating a working area for this...

- [ ] script that uses REST API to set up the renku keycloak realm
- [ ] initialize keycloak in the `start-renku` script for dev environment with the `Makefile`
- [ ] initialize keycloak in the helm post-install hook

closes https://github.com/SwissDataScienceCenter/renku/issues/214 
addresses #208


leafty:
Now, the hard part is finding a way to push the contents of https://github.com/SwissDataScienceCenter/renku/blob/master/charts/renku/charts/keycloak/templates/_renku-realm.tpl into keycloak somehow.

rokroskar:
is this still being worked on? Did @ebiiii take it over?

leafty:
Progress won't be fast, we'll need to rewrite the realm template as a script.

leafty:
Closing in favor of #336

PR #270, opened 2018-06-14
ci: remove image builds on cron jobs

rokroskar:
This gets rid of building all images from scratch on cron jobs. Not necessary at the moment and it's causing problems with disk space on travis.


rokroskar:
@leafty ok put the code back in with a comment. Please re-approve.

Issue #263, opened 2018-06-04
check image build status


rokroskar:
Once we have image builds working (contingent upon #260 and #246) we will want some way of letting the user know when the image build is completed. This could happen:

1) in the UI somewhere next to/near the "Launch notebook" button
2) in the notebook service (if image isn't in the registry, redirect to a "waiting for the image" page)
3) check the job status in the `SpawnerMixin`

cc @ciyer @jirikuncar @leafty @ableuler

leafty:
This can be a waiting page from the `Launch` button, similar to binderhub.

rokroskar:
Yes exactly. In addition there could be something indicating the status of the build in the ui.

leafty:
For ui implementation, 2 steps required on the gitlab api:
- hit `GET /projects/:id/repository/commits/:sha`: https://docs.gitlab.com/ce/api/commits.html#get-a-single-commit
- hit `GET /projects/:id/pipelines/:pipeline_id`: https://docs.gitlab.com/ce/api/pipelines.html#get-a-single-pipeline

rokroskar:
actually for this you just care about the image build so you want to look at 

`GET /projects/:id/jobs/` and filter on commit ID and the stage name (at this point it's `image_build`)

leafty:
true :)
but then you have to handle pagination

PR #278, opened 2018-06-20
jh: spawner checks for image_build jobs

rokroskar:
addresses #263 by checking for the `image_build` job in the spawner.

### DIFF
--- a/services/jupyterhub/spawners.py
+++ b/services/jupyterhub/spawners.py
@@ -86,7 +86,8 @@ class SpawnerMixin():
         options = self.user_options
         namespace = options.get('namespace')
         project = options.get('project')
-        commit_sha_7 = options.get('commit_sha')[:7]
+        commit_sha = options.get('commit_sha')
+        commit_sha_7 = commit_sha[:7]
 
         url = os.getenv('GITLAB_URL', 'http://gitlab.renku.build')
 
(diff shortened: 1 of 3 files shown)


services/jupyterhub/spawners.py
```diff
@@ -107,15 +108,55 @@ def start(self, *args, **kwargs):
...
-        self.image = '{image_registry}'\
-                     '/{namespace}'\
-                     '/{project}'\
-                     ':{commit_sha_7}'.format(
-                        image_registry=os.getenv('IMAGE_REGISTRY'),
-                        commit_sha_7=commit_sha_7,
-                        **options
+        # set default image
+        self.image = os.getenv(
+            'JUPYTERHUB_NOTEBOOK_IMAGE', 'jupyterhub/singleuser:0.8.1'
```
jirikuncar: IMHO the default is already in the config.

```python
if 'JUPYTERHUB_NOTEBOOK_IMAGE' in os.environ:
    self.image = os.getenv(...)
```

rokroskar: isn't that essentially the same as what I wrote?

jirikuncar: But you are probably experiencing a problem with `self` being global object that is shared by multiple requests.

jirikuncar:
The failed tests looks like an issue with renku-ui 😢

rokroskar:
yeah maybe, except that they pass on my machine...

services/jupyterhub/spawners.py
rokroskar: ah that might very well be. I'm not sure how you can reliably avoid this since the base `KubeSpawner` takes `self.singleuser_image_spec` when building the pod manifest. I tried serving multiple requests at once but I agree that this needs a proper test.

rokroskar:
good call @jirikuncar , I was using an old version of the ui image! 🥇

rokroskar:
@jirikuncar @leafty tests fixed.

PR #283, opened 2018-06-22
New way to build renku

leafty:
*Work in progress*

Will be inspired by: https://gitlab.com/renku/stage-renku/

Uses `pipenv`.

How to use it:
```bash
$ minikube start [options]
$ helm init [options]
$ helm upgrade --install nginx-ingress --namespace kube-system \
        --set controller.hostNetwork=true \
        --set tcp.2222=renku/renku-gitlab:22 \
        stable/nginx-ingress
$ make minikube-deploy
```


Makefile
```diff
@@ -1,261 +1,28 @@
-# -*- coding: utf-8 -*-
-#
-# Copyright 2017, 2018 - Swiss Data Science Center (SDSC)
```
jirikuncar: the license header should stay.

Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
```
rokroskar: If we're going to include `Pipfile` and `Pipfile.lock` (which I think is a good idea) they should probably include all the dependencies, including those in `docs/requirements.txt`?

leafty: Sure that can be added to the `[dev-packages]` section.

leafty: Done with f210f28edf7d4d980117f42988227d1390f8aa32

leafty:
I've been using this for a while, it's quite satisfying to deploy codes changes (e.g. https://github.com/SwissDataScienceCenter/renku-notebooks/pull/20) very quickly.

Need help to clean up where necessary, and to adapt the travis test and build.

leafty:
Yes, the reason is that I believe this is what you want as a Renku developper. Say you are working on a new feature for `renku-notebooks`, you will already be on another branch and modifying code. From there, you can simply issue `make minikube-deploy` in `renku` or more concisely `make -C ../renku minikube-deploy`.

What might be useful is a way to print the current branch and status of all repos from the makefile.

rokroskar:
Yes for the most part probably. Except when you're reviewing other people's pr's then it's nice to just specify the ref and have it deployed. Not a deal breaker though.

leafty:
@rok yes, what you said the other day about using this for staging is a good idea, but I'll refactor later to support that. The best for me would be that you create an issue with the desired workflow for you (as you are the one doing this the most). E.g. I need to know how you want the script to behave with unclean git repos.

rokroskar:
👍 sorry for repeating myself ;)

rokroskar:
@leafty neither of the README files now say anything about `make minikube-deploy` -- do you want to a) either include a short kubernetes section in the main readme under "starting the platform" or b) modify the README in the `charts` directory?

Makefile
```diff
@@ -263,3 +265,13 @@ test: .env demo
 
 wipe: .env
 	@./scripts/renku-wipe.sh
+
+.PHONY: minikube-deploy
+minikube-deploy:
```
rokroskar: shouldn't this depend on `clone`? i.e. clone a repo if it isn't there?

leafty: true

PR #293, opened 2018-06-29
postgresql as dep in charts

fgeorgatos:
factorize charts deps - the 1st blood

addresses #208


charts/renku/templates/configmap.yaml
```diff
@@ -63,3 +63,84 @@ data:
...
+        *.sql)    echo "$0: running $f"; "${psql[@]}" -f "$f"; echo ;;
+        *.sql.gz) echo "$0: running $f"; gunzip -c "$f" | "${psql[@]}"; echo ;;
+        *)        echo "$0: ignoring $f" ;;
+      esac
+      echo
+    done
+
+  init-keycloak-db.sh: |-
+    #!/bin/bash
+    set -e
```
leafty: TODO: `set -ex`

charts/renku/templates/post-install-job.yaml
```diff
@@ -8,12 +8,12 @@ metadata:
...
   annotations:
     "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-weight": "20"
     "helm.sh/hook-delete-policy": "before-hook-creation"
 spec:
   template:
     metadata:
-      name: "{{.Release.Name}}"
+      name: "{{.Release.Name}}-post-install"
```
leafty: TODO: rename to `{{ ... }}-post-install-gitlab`

fgeorgatos:
@leafty : re-review?

charts/renku/templates/post-install-job.yaml
```diff
@@ -8,12 +8,12 @@ metadata:
...
   annotations:
     "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-weight": "20"
     "helm.sh/hook-delete-policy": "before-hook-creation"
 spec:
   template:
     metadata:
-      name: "{{.Release.Name}}"
+      name: "{{.Release.Name}}-post-install-gitlab"
```
leafty: make it match with the name above

charts/renku/templates/configmap.yaml
```diff
@@ -63,3 +64,84 @@ data:
     # configure the logout redirect
     curl -f -is -X PUT -H "Private-token: ${GITLAB_SUDO_TOKEN}" \
       ${GITLAB_SERVICE_URL}/api/v4/application/settings?after_sign_out_path={{ template "http" . }}://{{ .Values.global.renku.domain }}/auth/realms/Renku/protocol/openid-connect/logout?redirect_uri=${GITLAB_URL}
+
+  # Init scripts that populate /docker-entrypoint-initdb.d
+
+  # Unashamedly copied from: https://github.com/docker-library/postgres/blob/master/9.6/docker-entrypoint.sh
+  init-postgres.sh: |-
```
rokroskar: This is going to keep crashing and recreating new pods to run the job until the server is available -- a better solution might be to wait for the server to be up before continuing. See the other post-install script: https://github.com/SwissDataScienceCenter/renku/pull/293/files#diff-0f15bad1c1882fee30ebc792c5a38350R24

leafty: That job is doing a curl to wait. Would need an equivalent psql command to ping.

rokroskar: yes I wasn't suggesting that what the other script does is ideal ;)  
But IMO the script should do the waiting, not rely on k8s to just keep relaunching the pod.

rokroskar: You could also install curl in the postgres container?

rokroskar:
Please cleanup the commits (either squash all at the end or selectively by hand and force-push)

charts/renku/templates/configmap.yaml
fgeorgatos: @rokroskar , @leafty : eh, what's the action point here? r we settled on a curl trick?

rokroskar: @fgeorgatos it was fixed in 864f2a0

PR #297, opened 2018-07-03
Fix postgresql script fail

leafty:
Job will error if postgres databases already exist


rokroskar:
I thought this was the case already before? was it changed in the postgres PR?

leafty:
yeah we added `set -ex` in the scripts

PR #298, opened 2018-07-04
GitLab LFS config

leafty:
This PR changes the GitLab LFS configuration to allow any configuration for lfs objects.

The exact configuration will depend on the backend, as GitLab is using [fog](https://github.com/fog) to perform LFS operations.

Available backend can be deduced by GitLab's Gemfile: https://github.com/gitlabhq/gitlabhq/blob/v10.8.5/Gemfile#L111-L118


rokroskar:
@leafty can you please test the LFS object migration on staging before we merge this?

PR #301, opened 2018-07-09
User private image

### LINKED ISSUE #276: jh: add registry credentials to pods
When using a private registry, the pod needs to be configured with the `imagePullSecret` otherwise images from private projects cannot be used. 

A workaround for now is to only allow public projects.

leafty:
Closes #276 

This PR adds the possibility to use the generated docker images from private projects.

This works by having the `renku-notebooks` service send the request to the spawner with a list of image pull secrets to include in the pod spec.


services/jupyterhub/Pipfile
```diff
@@ -0,0 +1,19 @@
+[[source]]
```
rokroskar: Are you including the `Pipfile` and `Pipfile.lock` in the commit on purpose? If so, don't you want to use them in teh `Dockerfile`?

leafty: Oops thanks. I did use pipenv to have the packages in pycharm.

PR #304, opened 2018-07-11
Added missing step to minikube first steps documentation.

ableuler:
After fresh install of helm, one also has to add the jupyterhub repository.


leafty:
Thanks, I added this to http://renku.readthedocs.io/en/latest/developer/minikube.html but not here.

PR #317, opened 2018-07-17
Use jupyterhub v0.7-dev

### LINKED ISSUE #299: update JupyterHub to v0.9
There have been quite some improvements in JupyterHub v0.9 -- notably, culling of idle servers should work for named servers (which is what we are using) so we shouldn't have pods lying around that are several days old anymore. 

With a bugfix release v0.9.1 now available, we should consider switching to this version. For the moment, this means using a development version of their charts -- see https://jupyterhub.github.io/helm-chart/ for available versions. I would start from jupyterhub-v0.7-f8dec3f since that is the commit that includes JH v0.9.1 (see the zero-to-jupyterhub repo commits: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commits/master). 

Note that to update the running instances, we may need to migrate the database manually with helm: http://zero-to-jupyterhub.readthedocs.io/en/latest/upgrading.html

leafty:
Closes #299


rokroskar:
@leafty I tested this just now, seems fine to me. Is there anything left to do? 

If not, could you please update the branch so we can merge? We also then need to update the singleuser image in #325 to use `jupyterhub/singleuser:0.9.1` as the base image.

leafty:
Test was not passing, don't know if the update broke something.

services/jupyterhub/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM jupyterhub/jupyterhub:0.8.1
+FROM jupyterhub/jupyterhub:0.9.1
```
rokroskar: the version needs to be changed in `jupyterhub-k8s.Dockerfile` as well...

rokroskar:
I'm having a problem with the idle server culling now -- getting a 400 for some reason when it tries to delete a server. Not sure why...

rokroskar:
the idle server cull error is a bug in the culling script in jupyterhub -- I'm fixing it in my fork now and will send a PR.

rokroskar:
This needs to be merged together with https://github.com/SwissDataScienceCenter/renku-notebooks/pull/25 and https://github.com/SwissDataScienceCenter/renku-ui/pull/266

services/jupyterhub/jupyterhub-k8s.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM jupyterhub/k8s-hub:v0.6
+FROM rrrrrok/k8s-hub:8aab19b
```
leafty: Switch back to official one?
Or still culling issues?

rokroskar: good point, the PR has been merged to master

leafty:
@rokroskar i think it's good here

PR #323, opened 2018-07-18
[Docs] Added pipenv install step to minikube setup

### LINKED ISSUE #315: [Docs] Minikube: add `pipenv install` step
this step is missing in the docs

leafty:
Closes #315


rokroskar:
The doc looks good, but I'm a bit confused by the previous part that is actually not being modified here... you say to include the SDSC helm chart registry, but isn't `make minikube-deploy` building and deploying all the charts for you? So it's not using the ones from the registry, correct?

leafty:
True, it's not required to add the renku helm repo.

PR #330, opened 2018-07-23
Use docker executor for gitlab runner

### LINKED ISSUE #327: [minikube] Runner -> docker executor
Remove runner chart and add instructions to start a runner with docker executor.

This will do:
1. Same setup as deployments
2. Use docker socket for image builds (will enable image caching).
3. Can launch notebook images. (Image already present from the build)

leafty:
Image building and notebook launch button work out of the box with `make minikube-deploy`.
The registry url uses the in cluster IP as a way to achieve this.

~~TODO:~~
~~- [ ] add doc for `.gitlab-ci.yml`~~

Closes #327 

~~Making the gitlab registry work properly is still not simple, so recommended to simply not use it in `.gitlab-ci.yml`:~~
```yaml
image_build:
  stage: build
  image: python:3.6
  before_script:
    - curl -sSL https://get.docker.com/ | sh
    # - docker login -u gitlab-ci-token -p $CI_JOB_TOKEN http://$CI_REGISTRY
  script:
    - CI_COMMIT_SHA_7=$(echo $CI_COMMIT_SHA | cut -c1-7)
    - docker build --tag $CI_REGISTRY_IMAGE:$CI_COMMIT_SHA_7 .
    # - docker push $CI_REGISTRY_IMAGE:$CI_COMMIT_SHA_7
  tags:
    - image-build
```

~~As the image is built in minikube, the image is available for JH.~~

### DIFF
--- a/charts/renku/charts/gitlab/templates/registry-service.yaml
+++ b/charts/renku/charts/gitlab/templates/registry-service.yaml
@@ -10,6 +10,7 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
+  clusterIP: 10.100.123.45
   type: NodePort
   ports:
     - port: 8105
--- a/charts/renku/requirements.yaml
+++ b/charts/renku/requirements.yaml
@@ -7,11 +7,6 @@ dependencies:
   alias: notebooks
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
   version: "0.1.0-3a7db4d"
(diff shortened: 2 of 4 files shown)


rokroskar:
@leafty hmm that will work but it will end up being quite annoying to have to change those lines all the time. Specifically, the automatic demo script + test won't work.

leafty:
these two could be adapted for minikube.

leafty:
@rokroskar I have found a potential workaround.

PR #336, opened 2018-07-25
use official keycloak chart

ebiiii:
contributes to #208


charts/renku/values.yaml
```diff
@@ -70,31 +70,48 @@ keycloak:
...
+      dbVendor: postgres
+
+      ## The following values only apply if "deployPostgres" is set to "false"
+
+      dbName: keycloak
+      dbHost: renku-postgresql
+      dbPort: 5432 # 5432 is PostgreSQL's default port. For MySQL it would be 3306
+      dbUser: keycloak
+
+      existingSecret: renku
```
leafty: TODO: Document the following helm template caveat (at least in `values.yaml`).

Convention in helm chart is to define name of resources using the following template:
```
{{- define "postgresql.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 20 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 20 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 20 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

This makes the resource name depend on the Helm release name.

The values of `dbHost` and `existingSecret` work when the release name is `renku`. As @ebiiii experienced, when deploying to a different release name, then at least `dbHost` and `existingSecret` need to be adjusted.

rokroskar: is there no way to define namespace-resolvable aliases that would make this work in general? Like network-specific aliases in `docker-compose`.

rokroskar:
I tried out the PR on staging and it looks good, thanks @ebiiii -- modulo @leafty's request for the extra line of docs, this can be approved and merged.

charts/renku/values.yaml
leafty: We could do by using `fullnameOverride`. If we want to have a preference for that, we can investigate it in a separate issue.

rokroskar:
what is this PR waiting for?

leafty:
@ebiiii need to manually set some values in https://github.com/SwissDataScienceCenter/renku/blob/master/scripts/travis-build.sh

PR #337, opened 2018-07-26
Use official redis charts

leafty:


charts/renku/values.yaml
```diff
@@ -450,32 +467,17 @@ gitlab:
...
-  ## Horizontal scaling of pods
-  # replicaCount: 1
-  # antiAffinity: soft
+  usePassword: false
 
-  ## Resource requests/limits for Redis
-  # resources: {}
+  master:
+    persistence:
+      enabled: false
```
leafty: TODO: Network policy.

rokroskar:
@leafty can you please resolve conflicts here so it can be merged?

leafty:
It's a good idea to have it

charts/renku/requirements.yaml
```diff
@@ -1,29 +1,30 @@
 dependencies:
 - name: renku-ui
   alias: ui
-  version: "0.1.0-c776f4e"
+  version: "0.1.0-0bc7625"
```
leafty: These are upgrades, not downgrades?

charts/renku/values.yaml
```diff
@@ -493,32 +493,20 @@ gitlab:
...
-  # service:
-  #   port: 6379
-  #   type: ClusterIP
+  master:
+    persistence:
+      enabled: false
+      
+  networkPolicy:
+    enabled: true
+    allowExternal: false
```
leafty: Works as intended (only GitLab container can contact redis)

charts/renku/requirements.yaml
rokroskar: yes -- were they supposed to be downgrades?

leafty: nah, i just didn't check the git logs to verify that

PR #341, opened 2018-07-30
Use sub claim for token to identify users in GitLab

leafty:


leafty:
Let's not forget to publish the chart

PR #342, opened 2018-07-30
docs: use patched sphinx until upstream PR is merged

rokroskar:
This builds the doc against my patch of the sphinx graphviz extension, which enables relative paths in custom node images (i.e. the icons actually show up). See https://renku.readthedocs.io/en/fix-324/user/reproducibility.html

The patch has been submitted as a PR to sphinx.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -29,7 +29,7 @@ requests==2.19.1
 simplegeneric==0.8.1
 six==1.11.0
 snowballstemmer==1.2.1
-sphinx==1.7.5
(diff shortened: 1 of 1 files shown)


rokroskar:
@leafty I think `Pipfile.lock` by definition always uses a specific commit hash. I've pushed a cherry-picked change from #347 that updates both `Pipfile` and `Pipfile.lock`.

leafty:
@rokroskar when I delete the lock file on my end, it recreates it without the hash for the editable sphinx.

rokroskar:
@leafty just with the branch ref?

leafty:
@rokroskar yup that's how it behaves on my side

PR #345, opened 2018-07-31
Fix demo user mapping in gitlab after #341

leafty:
Fixes [minikube] deployments after #341

### DIFF
--- a/charts/renku/templates/configmap.yaml
+++ b/charts/renku/templates/configmap.yaml
@@ -53,7 +53,7 @@ data:
         -d  '{"username": "demo",
         "email": "demo@datascience.ch",
         "name": "John Doe",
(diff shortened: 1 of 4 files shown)


charts/renku/templates/configmap.yaml
```diff
@@ -53,7 +53,7 @@ data:
         -d  '{"username": "demo",
         "email": "demo@datascience.ch",
         "name": "John Doe",
-        "extern_uid": "demo",
+        "extern_uid": "e144b235-793b-4e2e-bb1f-1f8baccc321f",
```
rokroskar: this is always the same value?

leafty: yes, it's hard-coded in the realm template

PR #365, opened 2018-08-13
Change keycloak theme

ebiiii:
following instructions from : https://github.com/helm/charts/tree/master/stable/keycloak#providing-a-custom-theme

~~I still need to figure out why it started crashing my minikube and how to add the pointers to the theme directly in the imported realm.~~

Update: if the first time the init container crashes and then you update the chart to fix it, keycloak will not set the admin user of its backend!
Update2: if building the theme image using an old version of busybox, the cp command crashes silently Oo
Update3: to enable the new theme, we need to manually select it in the admin console and add: `<div id="renku-logo-wrapper"></div>` as HTML display name.


charts/renku/values.yaml
```diff
@@ -88,7 +88,7 @@ keycloak:
           - -c
           - |
             echo "Copying theme..."
-            cp -R /renku-theme/* /theme
+            cp -R /renku_theme/* /theme
```
leafty: :boom:

ebiiii: it seems consistent with https://github.com/SwissDataScienceCenter/keycloak-theme/blob/master/Dockerfile#L2

ebiiii:
It is now fixed.
Note1: first time the init container is run, it is with user root, subsequent times it is run it is with user 1000 !!! (I just chown'd the folder to 1000 and it works for both)
Note2: I tried to integrate the theme directly in the renku repository, but chartpress cannot set the version of the image, as is it in the middle of a multiline text and not a value. If you have suggestions...

rokroskar:
@ebiiii this rings a bell -- I think we had the same problem with the init containers we use for the notebook servers. 

Regarding your second point, unfortunately I'm not sure there is a good solution atm. The keycloak chart uses the `tpl` function to inject the init containers which I suppose requires a string (https://github.com/helm/charts/blob/e6ac09ea3ea00f65c924da62deb264e242511d39/stable/keycloak/templates/statefulset.yaml#L47). I guess this is much easier than iterating (is that even possible with templating functions?)

rokroskar:
actually @ebiiii you mentioned on Monday you would inject the fix for the broken `<div>` tag so it didn't have to be done by hand -- did you manage to do this?

rokroskar:
tested once more on staging, setup looks ok!

PR #366, opened 2018-08-14
singleuser: Make jovyan sudoer

leafty:
Because at times, some python packages have system dependencies.

Example:
```bash
$ pip install opencv-python
[...]
Successfully installed numpy-1.15.0 opencv-python-3.4.2.17
$ python -c "import cv2"
[...]
ImportError: libgthread-2.0.so.0: cannot open shared object file: No such file or directory
```

Need to run:
```bash
$ sudo apt update && sudo apt install -y python-opencv
```

Of course, we'd still recommend to add the `apt install` to the Dockerfile, but it's not nice to have to fully
cycle a jupyter server and wait for an image build to install things.

### DIFF
--- a/services/singleuser/Dockerfile
+++ b/services/singleuser/Dockerfile
@@ -53,6 +53,8 @@ COPY git-config.bashrc /home/jovyan/
 RUN cat /home/jovyan/git-config.bashrc >> /home/jovyan/.bashrc && rm /home/jovyan/git-config.bashrc
 
 # Add user `jovyan` to sudoers (passwordless)
+USER root
 RUN echo "jovyan ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+USER jovyan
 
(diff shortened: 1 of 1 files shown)


rokroskar:
I think the scope of sudo for Jovan has to be limited narrowly

leafty:
@rokroskar why? you can make and run your own Dockerfile on renku anyway, so I find it quite silly to put artificial restrictions here.

rokroskar:
That's true I suppose, but by default that should not be the case. In the future we could add checks to prevent as much as possible people running things as root in the notebook containers.

leafty:
@rokroskar The containers are running in cgroups, so I don't see why this would be an issue, except if there are vulnerabilities in the container execution model.

It is very simple to put the setuid on any binary and run things as root, even interpreted languages (shell, python, ...)

PR #368, opened 2018-08-14
jh: make pod culling time 24 hours

### LINKED ISSUE #367: increase JH culling timeout
it's 20 minutes at the moment -- make it 24 hours?

rokroskar:
closes #367

### DIFF
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -622,7 +622,7 @@ jupyterhub:
       enabled: false
   cull:
     enabled: true
-    timeout: 600
+    timeout: 86400
     every: 60
 apispec:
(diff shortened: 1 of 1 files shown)


leafty:
@rokroskar mmmh informing the user when there is no recourse... That is so great :smiling_imp:

rokroskar:
yeah I guess once a pod termination is initiated there is no way to stop it, but we could increase the termination grace period to something reasonable.

leafty:
Yes, but I don't even know if it could help. I think the process receive a SIGINT, so the grace period is meant to give it some time to stop.

I don't think you could even try to save stuff once that is triggered.

PR #369, opened 2018-08-14
Charts: Update requirements.yaml

leafty:
Bumping dependencies values

Taking today's latest.


leafty:
@rokroskar Of course if I delete `dependencies` at the top, it won't work

rokroskar:
@leafty oops didn't notice

PR #371, opened 2018-08-14
jh: use jupyterlab

### LINKED ISSUE #370: set JH to open JupyterLab
related to https://github.com/SwissDataScienceCenter/renku-ui/issues/273  and https://github.com/SwissDataScienceCenter/renku-notebooks/issues/27

rokroskar:
closes #370  and https://github.com/SwissDataScienceCenter/renku-notebooks/issues/27 -- however, beware that this brings up new problems. For example jupyterlab introduces the idea of workspaces, which seems to conflict a bit with our concept of tying a server to a git commit. Right now, if you try to launch two different jupyterlab sessions for two different commits in your history you will be asked to switch to a different workspace -- at which point you will redirect to an incorrect URL (doesn't take into account the `base_url` from jupyterhub -- bug report?)

### DIFF
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -543,7 +543,7 @@ jupyterhub:
   hub:
     image:
       name: renku/jupyterhub-k8s
-      tag: latest
+      tag: 0.0-unclean-1a673b8
     services:
       notebooks:
         url: http://renku-notebooks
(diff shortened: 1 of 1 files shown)


charts/renku/values.yaml
```diff
@@ -543,7 +543,7 @@ jupyterhub:
   hub:
     image:
       name: renku/jupyterhub-k8s
-      tag: latest
+      tag: 0.0-unclean-1a673b8
```
leafty: Got tricked by chartpress

rokroskar: Doh!

PR #376, opened 2018-08-16
Fix integration test

ableuler:
Adapt test to renaming of notebooks in UI.


rokroskar:
this was not the reason for the failed test... 

https://travis-ci.org/SwissDataScienceCenter/renku/builds/416801034?utm_source=github_status&utm_medium=notification

PR #380, opened 2018-08-21
[WIP] docs: restructure intro pages

rokroskar:
closes #351

PR #381, opened 2018-08-21
docs: restructure intro pages

### LINKED ISSUE #351: Restructure documentation
The documentation should be restructured with two main objectives in mind:
- Incorporate the concepts section (reproducibility, version control, environment) into the introduction part.
- Clearly distinguish between status-quo and vision.

rokroskar:
reopening #380 from a non-forked branch...

closes #351


scripts/run-tests.sh
```diff
@@ -28,6 +28,7 @@ set -ex
 # FIXME: make helm lint work
 # helm lint charts/renku -f charts/minikube-values.yaml
```
ableuler: hmm, ok - slowly passing builds are better than quickly failing ones I gues...

leafty: The problem is that when the chart ends, gitlab is not fully ready. The rest api endpoints sometime take a little bit longer to be up.

rokroskar: so maybe the readiness probe should hit `/api/v4` instead?

PR #382, opened 2018-08-21
Cleanup main repo

rokroskar:
Removing things we are not using anymore, mostly with regards to our (now obsolete) docker-compose setup and even some other relics from the distant past. 

extra bonus: the `run-tests` script can now again be used to test things locally 

Note that I removed `shellcheck` -- we can re-introduce it to check the scripts inside `./scripts` but I would then make a separate issue/PR for fixing up all the shell scripts as it looks to be a bit of work.


leafty:
Is it still a WIP?

rokroskar:
@leafty I don't think so -- I'll make sure it deploys ok still but the tests pass (I was expecting more of a struggle there, hence the `[WIP]`)

docs/developer/minikube.rst
```diff
@@ -100,17 +100,7 @@ For more on the Renku helm charts, go to ``charts/renku/README.rst``.
...
-.. code-block:: console
-
-    $ kubectl -n renku run renku-demo -it \
-        --env="GITLAB_URL=http://$(minikube ip)/gitlab" \
-        --env="KEYCLOAK_URL=http://$(minikube ip)" \
-        --image=renku/renku-demo:latest \
-        --restart=Never
-
-Then we can run the tests:
+Run the tests with ``helm``:
```
leafty: No need to run the demo script now?

leafty:
I'm glad I can read the `Makefile` again!

docs/developer/minikube.rst
rokroskar: I moved it to the test init container.

rokroskar:
@leafty actually I think we could remove the makefile completely. That requires adding some more instructions to the various docs so the lazy option is to leave it be for now.

PR #384, opened 2018-08-22
improve the 1-liner definition

fgeorgatos:


rokroskar:
I tried to simplify the wording so I removed the federated part -- plus we don't have any federation atm so to me it's a bit misleading.

rokroskar:
I'm closing this for now -- we will need to sync the 1-liner again anyway when we have a website for renku.

PR #385, opened 2018-08-22
include a statement about technologies used in intro

fgeorgatos:


docs/introduction/index.rst
```diff
@@ -115,6 +120,8 @@ accidental data duplication as much as possible.
 History
 -------
 
+- 2018.08.01. **Renku** Component Architecture is solidifying now to a more long-term design
```
rokroskar: I wand to add a line here when we actually do a release, so I think we can remove for now

docs/introduction/index.rst
```diff
@@ -19,7 +19,12 @@ Features
 --------
 
 Renku consists of a collection of services, including a web-based user interface
-and a command-line client.
+and a command-line client, exploiting in a coherent setup the joint features of:
+* GitLab - version control
+* JupyterHub - Interactive Notebooks
+* Kubernetes - Container Orchestration
+* Keycloak - Identity and Access Management
+* Python - Scripting, glue logic and much more...
```
rokroskar: remove "and much more"

docs/introduction/index.rst
```diff
@@ -19,7 +19,12 @@ Features
 --------
 
 Renku consists of a collection of services, including a web-based user interface
-and a command-line client.
+and a command-line client, exploiting in a coherent setup the joint features of:
+* GitLab - version control
```
rokroskar: gitlab is actually for repository management -- version control is handled by git. Could we just say `GitLab - repository management`?

rokroskar: need a blank line to get the list

docs/introduction/index.rst
```diff
@@ -19,7 +19,13 @@ Features
...
 
 Renku consists of a collection of services, including a web-based user interface
-and a command-line client.
+and a command-line client, exploiting in a coherent setup the joint features of:
+
+* GitLab -  repository management
+* JupyterHub - Interactive Notebooks
+* Kubernetes - Container Orchestration
+* Keycloak - Identity and Access Management
+* Python - Scripting & glue logic
```
rokroskar: hmm I don't know about including python here -- it's not really a technology. It's clearly different than the other things on the list so I would remove it. Maybe also add Common Workflow Language and hyperlinks to the web pages for these components?

fgeorgatos:
@rok: thanks for the markup improvement - merged! I feel for you, with the markup format fighting :)

PR #386, opened 2018-08-23
Spawners: run `git lfs install --local` in init container

leafty:
So that users don't need to do it.

addresses https://github.com/SwissDataScienceCenter/renku-python/issues/213


leafty:
Tested

PR #391, opened 2018-08-23
jh: improve the notebook server setup

rokroskar:
* set the working directory to be the project name
* set the push behavior to squelch warning message

Note that this change means that in jupyterlab the terminal opens directly in the project directory (which I think is convenient and the preferred behavior). Before, the file browser and the terminal opened in different directories. In addition, the directory is actually the repo (project) name instead of the generic `work`.


services/jupyterhub/spawners.py
```diff
@@ -371,6 +376,7 @@ def get_pod_manifest(self):
                     'git submodule init && git submodule update && '
                     'git reset --hard {commit_sha} &&'
                     'git lfs install --local &&'
+                    'git config push.default simple &&'
```
leafty: Annoyed by the wall of text,heh? :wink:

PR #403, opened 2018-08-29
Update gateway section in developer docs

ableuler:
Add sequence diagram and more information to gateway section in developer docs.

### DIFF
--- a/docs/developer/services/api_gateway.rst
+++ b/docs/developer/services/api_gateway.rst
@@ -3,9 +3,52 @@
 API gateway
 ===========
 
-.. contents::
-    :depth: 1
-    :local:
+The API gateway acts as a middle layer between the Renku clients and the
+different backend services (GitLab, JupyterHub, Renku storage service, ...).
+More specifically, the API gateway makes life easier for client developers in
+the following ways:
 
-The API gateway serves as a middle layer between the Renku clients and the
-different backend services.
+- **Translation layer**:
+  In cases where the data model used in the backend service
+  does not match the data model of Renku, the gateway performs the necessary parsing
+  of the resources and thus prevents re-implementation of very similar code throughout
+  all clients.
+
+- **Resource specific routing**:
+  Notebook servers are stored in JupyterHub, project details in GitLab and datasets
+  have to be fetched through the storage service - this must not be the worry of
+  each client (developer), instead it is handled by the gateway.
+
+- **Token handling**:
+  Even with all backend services offering the possibility to rely on an OpenID-connect
+  provider (OIDC) such as Keycloak, the API of those services do not currently accept
+  access tokens issued by the OIDC providers. The gateway stores the access tokens
+  for the different services, therefore allowing clients to access all resources
+  using the same credentials.
+
+- **Confidential client**:
+  The gateway serves as the confidential part of the web UI, which is a single
+  page application and can therefore not store client secrets. This allows
+  using the `OAuth2 authorization code grant`__ for user login.
+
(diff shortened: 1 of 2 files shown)


docs/developer/services/api_gateway.rst
```diff
@@ -7,5 +7,51 @@ API gateway
     :depth: 1
     :local:
 
-The API gateway serves as a middle layer between the Renku clients and the
-different backend services.
+The API gateway acts as a middle layer between the Renku clients and the
+different backend services (GitLab, Jupyterhub, Renku storage service, ...),
+thereby providing a single uniform interface to all backend services. More specifically,
```
rokroskar: maybe  the rest of the sentence starting at "thereby" could be dropped? Would be simpler and avoid repetition

docs/developer/services/api_gateway.rst
```diff
@@ -7,5 +7,51 @@ API gateway
...
+  of the resources and thus prevents re-implementation of very similar code throughout
+  all clients.
+
+- **Resource specific routing**:
+  Notebook servers are stored in JupyterHub, project details in GitLab and datasets
+  have to be fetched through the storage service - this must not be the worry of
+  each client (developer), instead it is handled by the gateway.
+
+- **Token handling**:
+  Even with all backend services offering the possibility to rely on a OpenID-connect
```
rokroskar: rely on a -> rely on an

docs/developer/services/api_gateway.rst
```diff
@@ -7,5 +7,51 @@ API gateway
...
+  provider (OIDC) such as Keycloak, the API of those services do currently not accept
```
rokroskar: do currently not -> do not currently

docs/developer/services/api_gateway.rst
```diff
@@ -7,5 +7,51 @@ API gateway
...
+Current state of implementation
+-------------------------------
+
+Not all the features mentioned above are already implemented. Currently, the
+API gateway is handling login functionalities for the Renku web UI and acts as
+a proxy between the UI and the GitLab API. The gateway does not yet handle
+user-specific access tokens. Instead it checks the validity of a users access
+token and forwards requests as root on behalf of the authenticated user.
+Furthermore, GitLab resources are simply forwarded to the UI without any
+modifications.
```
rokroskar: since the work to rectify this situation will soon begin, maybe say something about the timeline here?

docs/developer/services/api_gateway.rst
```diff
@@ -7,5 +7,51 @@ API gateway
...
+
+The diagram below illustrates the sequence of events that currently take place
+in the API gateway.
+.. _fig-uml_gateway_service:
```
rokroskar: need a blank line between the paragraph and the anchor text

rokroskar:
the sequence diagram looks good! In the "on startup" section, does't the gateway make a request for the public key?

docs/developer/services/api_gateway.rst
ableuler: removed.

ableuler:
> the sequence diagram looks good! In the "on startup" section, does't the gateway make a request for the public key?
Yes it does. To me this diagram shows the flow of information rather than an exactly displaying all http requests and responses. I've dropped some of the request/response arrows in other places too, for example a redirect in the OAuth2 flow is an http response followed by a request to the new url. I find it to be more understandable like this.

rokroskar:
Ok makes sense, as long as we're consistent in the sequence diagrams. The one for the notebook service may not be very diligent in this regard.

PR #405, opened 2018-08-29
docs: update firststeps to include refactoring

rokroskar:
Revise the first-steps tutorial to include a refactoring step in order to run the code as a standard python script rather than via `papermill`.


3C111:
General comment: to copy in jupyterlab you need something more than ctrl+shift+v. I have middle-mouse-click for copying and that works after clicking several times.

leafty:
@3C111 I'll check if there's an issue about that in jupyter, but you can paste fine with ctrl shift v.

You can't copy (the key combination is bound to a lab thing) so it's annoying. Using the mouse copy/paste buffer works too but it might not be available on all Linux distros

rokroskar:
On Mac you copy with `Cmd+C` as normal

3C111:
@leafty Maybe it is a firefox thing? I am just saying, it didn't work for me. However, this is more of a Jupyterlab issue, not a Renku issue.

3C111:
BTW: Is there a reason why we don't directly make the 'data' and 'notebooks' directories? It feels a bit artificial to have the users explicitly make the notebooks folder, that way it seems like it is optional.

leafty:
@3C111 -> https://github.com/SwissDataScienceCenter/renku-project-template/issues/6

3C111:
In the Ku section, Copy/paste of this command 'Let's explore the dataset! Here is what we know:

Exploration notebook' Didn't show the notebook in the ku itself. Did anyone else encounter this? If I copy the last line in a command to the Ku I do get to the notebook.

docs/user/firststeps.rst
```diff
@@ -228,9 +235,78 @@ For example, if you want to keep the new notebook(s), run the following.
...
+    mkdir -p notebooks
+    wget -O "notebooks/zhbikes-notebook.ipynb" https://raw.githubusercontent.com/SwissDataScienceCenter/renku/master/docs/_static/zhbikes/ZHBikes.ipynb
+    git add notebooks
+    git commit -m"Added zuerich bike notebook"
+    git push
+
+
+Refactor the notebook
+"""""""""""""""""""""
+
```
3C111: Is it possible to say how you did this? Is there a special command for that? I can imagine it is useful for users to know!

docs/user/firststeps.rst
```diff
@@ -228,9 +235,78 @@ For example, if you want to keep the new notebook(s), run the following.
...
+Here, we have already done the refactoring work for you - to get the scripts,
+run:
+
+.. code-block:: console
+
+    mkdir -p src
+    wget -O "src/clean_data.py" https://raw.githubusercontent.com/SwissDataScienceCenter/renku/master/docs/_static/zhbikes/clean_data.py
+    wget -O "src/plot_data.py" https://raw.githubusercontent.com/SwissDataScienceCenter/renku/master/docs/_static/zhbikes/plot_data.py
+
```
3C111: screenshot_cleandata
Computer says no. The file is not properly downloaded and there is no explanation. The other file downloaded fine. No error messages..I tried again a few times with the same result. Ideas?

leafty: I guess: replace `master` with `revise-firststeps`

3C111: Please note that I have done that already.

rokroskar: I don't understand, is the link incorrect?

docs/user/firststeps.rst
rokroskar: You mean refactoring? Did it by hand...

docs/user/firststeps.rst
3C111: I copied the link by hand from the repo so I don't think so. I don't know what is going on, but what happened is that the file starts to download and then stops. Result: the file downloaded partially. I am not sure what the problem was here. I tried several times yesterday and kept getting a partially broken file. This morning I tried and the file was downloaded properly. Maybe something in the environment? 
For now: moving on.

docs/user/firststeps.rst
3C111: Ah I thought maybe there was a secret python trick or so :]

leafty: I remember @rokroskar complaining that copy/pasting from a notebook is not a thing

rokroskar: Well it is a painful cell-by-cell thing

docs/user/firststeps.rst
rokroskar: Ok thanks, I'll try as well

docs/_static/zhbikes/plot_data.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # check if the figures exist already. If so, remove them
+    remove_if_exists('figs/grid_plot.png')
+    plt.savefig('figs/grid_plot.png')
+
+    sns.set(rc={'figure.figsize': (15, 10)})
+    sns.lineplot(
+        x="datetime", y="velo_all", hue="counting_station", data=df_weekly)
+    remove_if_exists('figs/cumulative.png')
+    plt.savefig('figs/cumulative.png')
```
leafty: For me, the second plot retains the grid from the first one.

rokroskar: oops -- should be fixed now

rokroskar: tl;dr?

leafty: it worked ;)

docs/user/firststeps.rst
rokroskar: `wget https://raw.githubusercontent.com/SwissDataScienceCenter/renku/revise-firststeps/docs/_static/zhbikes/clean_data.py` worked for me

leafty: I think it was a network/cdn fluke yesterday

PR #412, opened 2018-09-04
add possibility to configure users for tests

### LINKED ISSUE #411: use a configmap for renku test users
The renku test uses "real" users -- we should use a configmap in the renku test chart to use robot users instead.

fgeorgatos:
in follow up to https://github.com/SwissDataScienceCenter/renku-demo/pull/20 

closes #411


rokroskar:
see my comment in renku-demo -- I would prefer to leave the users there untouched and instead inject `users.json` into the test container here via a configmap

rokroskar:
You can make that change a part of this PR

charts/renku/values.yaml
```diff
@@ -588,6 +588,23 @@ tests:
...
     repository: renku/tests
     tag: '0.2.0-rc4'
+  users_json: |
+    [
+      {
+        "username": "bwayne",
+        "firstName": "Bruce",
+        "lastName": "Wayne",
+        "email": "bwayne@example.com",
+        "password": "IamBatman"
```
rokroskar: I suppose if these go in the `values.yaml` we should either a) make sure that the users are deleted once the test is done or b) provide a different `users.json` for deployments.

leafty: I think we may want to tell folks to be careful if they want to run `helm test` in production

rokroskar: or template the password? We can leave it out of the default `values.yaml` and use the `required` function when evaluating in https://github.com/SwissDataScienceCenter/renku/pull/412/files#diff-3f630e831a60063e0ab2d4a02e70c2acR28

charts/renku/values.yaml
```diff
@@ -588,6 +588,17 @@ tests:
   image:
     repository: renku/tests
     tag: '0.2.0-rc4'
+  users_json:
+    - username: bwayne
+      firstName: Bruce
+      lastName: Wayne
+      email: bwayne@example.com
+      password: IamBatman
```
rokroskar: Leave the passwords out of the defaults

tests/users.json
```diff
@@ -0,0 +1,16 @@
+[
+  {
+    "username": "cramakri",
```
rokroskar: Isn't the whole point to use robot users?

leafty: > see my comment in renku-demo -- I would prefer to leave the users there untouched and instead inject users.json into the test container here via a configmap

There is coupling with `renku-demo`. The default has to be the same here and there.

charts/renku/values.yaml
leafty: what to do then? it could be overly complicated if you want to generate one and insert it using the helm template functions

rokroskar: All you need to do is something like:

```
test-users:
  - user1:
     username: bwayne
     firstName: Bruce
     lastName: Wayne
     email: bwayne@example.com
  - user2:
     username: ckent
     firstName: Clark
     lastName: Kent
     email: ckent@example.com
```

And in the template just put in the password.

leafty: I don't think I follow. Did you check the template? It's using `toJson` so it's not possible

rokroskar: right... the question then is why do we need the test users in `values.yaml`? They can just be in `secrets.yaml` with the JSON written out and the password field templated.

leafty: I still don't see what is the problem with the current situation. We can comment the example values.

rokroskar: Ok -- for the time being we can add a comment to `values.yaml` -- one easy way out is to change the `users.json` to yaml in `renku-demo` which would make it easier to template.

leafty:
@rokroskar for me it should be OK

PR #414, opened 2018-09-05
Set renku version to 'latest' for minikube deployments

ableuler:
Set the renku version (affecting the template used in project creation) to latest for minikube deployments. Does currently not affect anything than project creation.

### DIFF
--- a/charts/minikube-values.yaml
+++ b/charts/minikube-values.yaml
@@ -13,6 +13,8 @@ ingress:
 global:
   gitlab:
     urlPrefix: /gitlab
+  renku:
+    version: 'latest'
 
 gitlab:
   ssh:
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -40,6 +40,10 @@ global:
   renku:
     ## Domain name for the deployed instance of renku
     domain: example.local
+    ## Renku version to be used (currently for project creation in
+    ## the UI) If not set explicitly the version will be picked up
+    ## from the respective renku (sub)chart. 
+    ## version: '0.0.0'
   ## Set to true if using https
   useHTTPS: false
 


leafty:
Does it affect other things? If not, maybe add a comment just above to clarify this.

ableuler:
@leafty nope and yes - comment added.

charts/minikube-values.yaml
```diff
@@ -13,6 +13,8 @@ ingress:
 global:
   gitlab:
     urlPrefix: /gitlab
+  renku:
+    version: 'latest'
```
leafty: I would rather put this in https://github.com/SwissDataScienceCenter/renku/blob/master/charts/renku/values.yaml,
with a proper doc comment

ableuler: My reasoning was that by setting `latest` in `values.yaml`, any deployment would use 'latest' unless it's explicitly overwritten. In the suggested setup it would pick up the chart version from the renku-ui chart by default which I would see as the expected behaviour. But maybe there should be a commented-out hint to the possibility of fixing the renku version in `values.yaml`?

leafty: The idea is to put `0.2.0` in the default values for the chart with the same tag, and `latest` otherwise. And doing this in the default values, not just minikube values

rokroskar: The other reason to put it in `values.yaml` is to document what it does -- not at all obvious if it's hidden in `minikube-values`

ableuler: @rokroskar I agree that indicating this possibility in `values.yaml` is a good idea.

But don't you think that hard-coding this value to '0.2.0' in `values.yaml` is very error prone? I see us forgetting to manually update this value in the future when we have a new release...

rokroskar: well that's part of the release. when we are in post-0.2.0 mode we can set it to latest but for the 0.2.0 release it needs to be pinned. Either way it should be in `values.yaml` and documented.

ableuler: I don't want it to be latest! I want it to pick up the version of the chart unless explicitly set to latest for dev purposes, therefore minikube!

rokroskar: Ok! I get it! I misunderstood! Is it templated then???? :)

ableuler: The following env var for the ui is set in `deployment.yaml`
```
            - name: RENKU_VERSION
               value: {{ .Values.global.renku.version | default ( .Chart.Version ) | quote }}
```

rokroskar: ok but still add it to `values.yaml` just commented out and with a description of what it does.

rokroskar: The issue with this setup is that we need to continue releasing the template along with the rest. I suppose that's fine but we need to automate the process

PR #415, opened 2018-09-05
0.2.0 rc4

rokroskar:
* update requirements
~~* add CI deploy on tags~~

### DIFF
--- a/charts/renku/requirements.yaml
+++ b/charts/renku/requirements.yaml
@@ -1,18 +1,18 @@
 dependencies:
 - name: renku-ui
   alias: ui
-  version: "0.2.0-rc3"
+  version: "0.2.0-rc4"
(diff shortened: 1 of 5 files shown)


charts/renku/requirements.yaml
```diff
@@ -1,20 +1,28 @@
...
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
   condition: gitlab.enabled
+  tags:
+    - production
+- name: gitlab
+  version: "~0.0-unclean-12345"
+  repository: "file://../gitlab"
+  condition: gitlab.enabled
+  tags:
+    - development
```
leafty: nice idea

rokroskar: well it doesn't work yet... :/

leafty:
@rokroskar you can remove my two commits since it doesn't help with the tests

PR #416, opened 2018-09-12
Update sub-chart dependencies in minikube-deploy

ableuler:
Update the dependencies of the sub-charts (currently only relevant for notebooks service) before packaging the renku chart in the `minikube-deploy` script.

Solves the problem that the jupyterhub service was not included in the minikube deployment unless the renku-notebooks dependencies had been added manually at some point before.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.6
 
 # RUN apk add --update openssl
-RUN wget https://github.com/mozilla/geckodriver/releases/download/v0.11.1/geckodriver-v0.11.1-linux64.tar.gz && \
+RUN wget https://github.com/mozilla/geckodriver/releases/download/v0.21.0/geckodriver-v0.21.0-linux64.tar.gz && \
     tar -xvzf geckodriver* && \
     chmod +x geckodriver && \
     mv geckodriver /usr/local/bin && \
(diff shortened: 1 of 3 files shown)


rokroskar:
The upshot is that now the tests fail even one step earlier... but it has nothing to do with this PR :)

ableuler:
@rokroskar tests are passing 🎉 , please reapprove!

tests/test_renku.py
```diff
@@ -44,8 +45,19 @@ def test_notebook_launch(browser):
     browser.fill('username', 'cramakri')
     browser.fill('password', 'cramakri')
     browser.find_by_id('kc-login').click()
+
+    # wait a bit for the page to load, helps to avoid test failures
+    time.sleep(2)
```
rokroskar: I don't think this is needed

tests/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM python:3.6
 
 # RUN apk add --update openssl
-RUN wget https://github.com/mozilla/geckodriver/releases/download/v0.11.1/geckodriver-v0.11.1-linux64.tar.gz && \
+RUN wget https://github.com/mozilla/geckodriver/releases/download/v0.21.0/geckodriver-v0.21.0-linux64.tar.gz && \
```
leafty: :champagne: :tada:

tests/test_renku.py
fgeorgatos: any guesses how this `time.sleep()` affects the behaviour of the above code? just saying...

rokroskar: @fgeorgatos see the comment from @ableuler below

PR #422, opened 2018-09-18
Mary had a little bug, little bug...

fgeorgatos:


docs/user/firststeps.rst
```diff
@@ -232,7 +232,7 @@ For example, if you want to keep the new notebook(s), run the following.
 .. code-block:: console
 
     git add notebooks # track everything inside the notebooks folder
-    git commit -m "Added some notebooks"
+    git commit -m "Added some notebooks" -a
```
rokroskar: why?

rokroskar:
I don't understand what problem this PR is solving, can you clarify please?

docs/user/firststeps.rst
fgeorgatos: Well, I was forced to use it to make it advance, but now that I see it again, wait before merging:
* let's try reproduce it elsewhere, to rule out an artifact on my end - or is it prereq for the dir?
(interestingly, I didn't need to use it a bit later on in the 1st steps instructions, in another git op)

rokroskar: what do you mean "to make it advance"? Are you sure you didn't forget the line above (`git add notebooks`)?

fgeorgatos: @rokroskar : I'm sure I did not forget that line - in fact, that was a prerequisite for `-a` to work!

rokroskar:
Can you please give an example of the error you were getting?

leafty:
Closing as a stale PR

PR #428, opened 2018-09-24
Allow for having multiple redis instances

ebiiii:
by resolving redis.fullname in a sort of scoped namespace. Once the related helm issue (https://github.com/helm/helm/issues/4535) is solved, we can also clean the other re-definitions of fullnames.

### DIFF
--- a/charts/renku/requirements.lock
+++ b/charts/renku/requirements.lock
@@ -10,13 +10,10 @@ dependencies:
   version: 0.2.0-rc4
 - name: gitlab
   repository: https://swissdatasciencecenter.github.io/helm-charts/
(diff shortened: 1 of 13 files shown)


charts/renku/requirements.lock
```diff
@@ -10,18 +10,15 @@ dependencies:
   version: 0.2.0-rc4
 - name: gitlab
   repository: https://swissdatasciencecenter.github.io/helm-charts/
-  version: 0.2.0-rc4
+  version: 0.0-072ab7e
```
rokroskar: Is this committed on purpose?

PR #430, opened 2018-09-25
charts: use 0.2.0 renku dependencies

rokroskar:
was too quick earlier, forgot to update requirements...


leafty:
i was too quick too

PR #440, opened 2018-10-10
Setting parameters for the gateway client

ebiiii:
Needed for: https://github.com/SwissDataScienceCenter/renku-gateway/pull/58


rokroskar:
Ideally those would be handled through k8s secrets.

ableuler:
I agree with @rokroskar but we do the same in many other places, so I'd prefer to move on with this PR while keeping track of this in a new issue #447.

PR #442, opened 2018-10-12
charts: update values for notebooks pr#56

rokroskar:


rokroskar:
can't be quicker than Travis... :)

PR #448, opened 2018-10-17
charts: adapt for notebook server options

rokroskar:
compatibility with
https://github.com/SwissDataScienceCenter/renku-notebooks/pull/68

### DIFF
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -60,10 +60,10 @@ ingress:
   ## Annotations for the created ingress
   annotations:
     ## The ingress class
-    kubernetes.io/ingress.class: 'nginx'
+    kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: 'true'
     nginx.ingress.kubernetes.io/proxy-body-size: '0' # Adjust to a reasonable value for production to avoid DOS attacks.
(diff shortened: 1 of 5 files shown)


charts/renku/values.yaml
```diff
@@ -60,10 +60,10 @@ ingress:
   ## Annotations for the created ingress
   annotations:
     ## The ingress class
-    kubernetes.io/ingress.class: 'nginx'
+    kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: 'true'
     nginx.ingress.kubernetes.io/proxy-body-size: '0' # Adjust to a reasonable value for production to avoid DOS attacks.
-    nginx.ingress.kubernetes.io/proxy-request-buffering: 'off' # Needed if GitLab is behind this ingress
+    nginx.ingress.kubernetes.io/proxy-request-buffering: off   # Needed if GitLab is behind this ingress
```
leafty: :/

rokroskar: Aaaaaaah!

leafty: But the problem remains that chartpress will mess up at publish time -_-'

rokroskar: but the charts we are using now have been published the same way and deployed without errors...

rokroskar:
continued as #451

PR #451, opened 2018-10-18
charts: adapt for notebook server options

rokroskar:
compatibility with
https://github.com/SwissDataScienceCenter/renku-notebooks/pull/68

was #448


charts/renku/requirements.lock
```diff
@@ -4,7 +4,7 @@ dependencies:
   version: 0.2.0
 - name: renku-notebooks
   repository: https://swissdatasciencecenter.github.io/helm-charts/
-  version: 0.2.0
+  version: 0.2.1-b0c93bb
```
ableuler: Fix before merging
```suggestion
  version: 0.2.1
```

charts/renku/requirements.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
 - name: renku-notebooks
   alias: notebooks
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: "0.2.0"
+  version: "0.2.1-b0c93bb"
```
ableuler: Same here...

ableuler:
```suggestion
  version: "0.2.1"
```

charts/renku/requirements.lock
rokroskar: well this was actually on purpose -- because this PR was for a specific version of the notebooks service that introduced a breaking change

ableuler: Ah, I see

ableuler:
I think it should be merged already because it contains changes that are needed to work with the current master branch on renku-notebooks. A new PR could be created to match the ongoing development in SwissDataScienceCenter/renku-notebooks#76.

charts/renku/values.yaml
```diff
@@ -60,10 +60,10 @@ ingress:
   ## Annotations for the created ingress
   annotations:
     ## The ingress class
-    kubernetes.io/ingress.class: 'nginx'
+    kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: 'true'
     nginx.ingress.kubernetes.io/proxy-body-size: '0' # Adjust to a reasonable value for production to avoid DOS attacks.
-    nginx.ingress.kubernetes.io/proxy-request-buffering: 'off' # Needed if GitLab is behind this ingress
+    nginx.ingress.kubernetes.io/proxy-request-buffering: off   # Needed if GitLab is behind this ingress
```
leafty: to be corrected next time

PR #454, opened 2018-10-22
Add values for the integration of jupyterhub in the gateway

ebiiii:
needed for https://github.com/SwissDataScienceCenter/renku-gateway/pull/63


charts/minikube-values.yaml
```diff
@@ -47,6 +47,9 @@ notebooks:
       services:
         notebooks:
           api_token: notebookstoken
+        gateway:
+          oauth_client_id: gateway
+          apiToken: dummy-secret
```
leafty: `changeme-openssl-rand-hex-32`

ebiiii: changed

PR #455, opened 2018-10-22
Rework charts secrets

### LINKED ISSUE #447: Use k8s secrets everywhere
We currently read secrets like passwords, tokens etc from values files. We should switch to using k8s secrets for this purpose.

See also https://github.com/SwissDataScienceCenter/renku-gateway/issues/44

### LINKED ISSUE #453: Make it possible to check secrets
The idea is to make it so that there is a reliable way to check if some secrets are left to their default value.

leafty:
Closes #447, closes #453

* Reworked gitlab secrets
* Reworked renku secrets

~~After this PR, all secrets should have a default value starting with `changeme` so that they are easily tracked.~~

This will probably need to be revisited after making sure subcharts follow the same principles.

### DIFF
--- a/charts/renku/templates/secrets.yaml
+++ b/charts/renku/templates/secrets.yaml
@@ -18,12 +18,27 @@ data:
   {{- end }}
   gitlab-postgres-password: {{ .Values.global.gitlab.postgresPassword | b64enc | quote }}
 {{- end }}
+
 {{- if .Values.keycloak.enabled }}
   keycloak-password: {{ .Values.keycloak.keycloak.password | b64enc | quote }}
   keycloak-postgres-password: {{ .Values.global.keycloak.postgresPassword | b64enc | quote }}
   renku-realm.json: {{ include "renku.keycloak.realm" . | indent 4  | b64enc | quote }}
 {{- end }}
+
   jupyterhub-postgres-password: {{ .Values.global.jupyterhub.postgresPassword | b64enc | quote }}
+
 {{- if .Values.tests.users_json }}
   users.json: {{ .Values.tests.users_json | toJson | b64enc | quote }}
+{{- end }}
+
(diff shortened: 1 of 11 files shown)


charts/renku/values.yaml
```diff
@@ -9,15 +9,15 @@ global:
     ## Client secret, set this to a random string.
     ## To change this value after deploying, first generate a new secret using the Keycloak admin console
     ## then paste the generated secret here
-    clientSecret: open_secret
+    clientSecret: changeme-openssl-rand-hex-32
```
rokroskar: use the `required` template function for these?

charts/renku/templates/secrets.yaml
```diff
@@ -18,12 +18,27 @@ data:
...
 {{- end }}
+
   jupyterhub-postgres-password: {{ .Values.global.jupyterhub.postgresPassword | b64enc | quote }}
+
 {{- if .Values.tests.users_json }}
   users.json: {{ .Values.tests.users_json | toJson | b64enc | quote }}
 {{- end }}
+
+{{- if .Values.notebooks.jupyterhub.auth.gitlab.clientSecret }}
+  jupyterhub-auth-gitlab-client-secret: {{ .Values.notebooks.jupyterhub.auth.gitlab.clientSecret | b64enc | quote }}
```
rokroskar: use the `required` function and do not generate a random string automatically

leafty:
Wow, Travis failed for a good reason:
```
Error: render error in "renku/templates/secrets.yaml": template: renku/templates/secrets.yaml:14:24: executing "renku/templates/secrets.yaml" at <required "Fill in .V...>: error calling required: Fill in .Values.global.gitlab.sudoToken with `openssl rand -hex 32`
```

rokroskar:
works as intended then!

leafty:
@rok up

PR #457, opened 2018-10-23
charts: set requirements to new versions

rokroskar:


leafty:
Note: tests not passing is probably due to change of UI

charts/renku/requirements.yaml
```diff
@@ -1,19 +1,19 @@
...
 - name: renku-notebooks
   alias: notebooks
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: "0.2.1-b0c93bb"
+  version: "0.2.1-3702f3b"
 - name: renku-gateway
   alias: gateway
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: "0.2.0"
+  version: "0.2.0-9094770"
```
leafty: can we do something about this not being `0.2.1-x`?

rokroskar: sure, and the UI as well I suppose

leafty: yup -> https://github.com/SwissDataScienceCenter/renku-ui/pull/344

charts/renku/requirements.lock
```diff
@@ -1,16 +1,16 @@
 dependencies:
 - name: renku-ui
   repository: https://swissdatasciencecenter.github.io/helm-charts/
-  version: 0.2.0
+  version: 0.2.1-9fc6260
 - name: renku-notebooks
   repository: https://swissdatasciencecenter.github.io/helm-charts/
-  version: 0.2.1-b0c93bb
+  version: 0.2.1-3702f3b
```
leafty: wait for update

PR #458, opened 2018-10-24
Hopefully a version bump fixes keycloak bugs (WIP)

fgeorgatos:
Hopefully a version bump will fix `../k8s-switch-limited/issues/7`

### DIFF
--- a/charts/renku/requirements.yaml
+++ b/charts/renku/requirements.yaml
@@ -19,7 +19,7 @@ dependencies:
   version: 0.14.4
   repository: "@stable"
 - name: keycloak
-  version: 3.2.0
+  version: 4.0.1
   repository: "@stable"
   condition: keycloak.enabled
 - name: minio


fgeorgatos:
marked as (wip) since config changes might be required...

rokroskar:
Which issues?

charts/renku/requirements.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   version: 0.14.4
   repository: "@stable"
 - name: keycloak
-  version: 3.2.0
+  version: 4.0.1
```
leafty: Stage it for test

rokroskar:
@fgeorgatos what is the status here?

charts/renku/requirements.yaml
fgeorgatos: done; we've arrived here now, it seems that v4 has a few changes to consider here:
https://stackoverflow.com/questions/45352880/keycloak-invalid-parameter-redirect-uri

fgeorgatos:
> @fgeorgatos what is the status here?

ongoing, tbc.

charts/renku/requirements.yaml
ebiiii: Try with 3.4.2 and we can tag this PR with renku 0.3.0. This one shouldn't break things.

leafty:
Postponed due to change of default signing method in newer version of keycloak

PR #459, opened 2018-10-25
Fix lfs docs

leafty:
From lfs help:
```
$ g lfs pull -h
git lfs pull [options] [<remote>]
```


rokroskar:
Is it possible this is version dependent? I needed to specify a remote if I recall correctly

PR #463, opened 2018-10-31
Fix minikube-values.yaml

ableuler:
Fix the postgres password in the url used by jupyterhub to connect to postgres.

Note: This  works, however it's not very elegant to just copy paste a value from another line in `minikube-values.yaml`. On the other hand, values files are not templateable themselves... Is there a better way to do this?


leafty:
Nope, blame the fact that the JH chart does offer only the db URL and does not break it in parts

rokroskar:
Could make a contribution to the chart in zero-to-jupyterhub...

PR #467, opened 2018-11-12
Update gateway documentation

ebiiii:


scripts/travis-before-script.sh
```diff
@@ -36,6 +36,7 @@ done
 
 helm init --wait
 helm upgrade --install nginx-ingress --namespace kube-system \
+    --version 0.30.0 \
```
leafty: We can do with latest, just need to stop builds then purge nginx

PR #480, opened 2018-11-27
charts: set version to 0.3.0 everywhere

rokroskar:
wait for all of the subcharts with v0.3.0 to be available to set the `requirements.lock`


rokroskar:
@leafty tests fixed -- I think it's gtg.

leafty:
Great, I'll just try to deploy on minikube to check.

PR #483, opened 2018-12-10
[chart] increase timeout for keycloak readiness probe

leafty:
It was too short, so keycloak mey not have the time to start


leafty:
@rokroskar sorry, forgot the liveness probe :/

PR #491, opened 2018-12-14
Add graph components to main chart

rokroskar:
Webhook and triples service works with the helm command found in `charts/README.md`. 

~~ingress for Jena doesn't work because Jena is not running on a sub-path. Unclear at the moment how to fix this, presumably there is some way to set http server options.~~

We aren't exposing Jena via ingress for the time being. Services talk to Jena via cluster routing. 

contingent on https://github.com/SwissDataScienceCenter/renku-graph/pull/26
uses https://github.com/SwissDataScienceCenter/renku-graph/pull/32


charts/renku/requirements.lock
```diff
@@ -11,6 +11,9 @@ dependencies:
 - name: gitlab
   repository: https://swissdatasciencecenter.github.io/helm-charts/
   version: 0.3.0
+- name: renku-graph
+  repository: file://../../../renku-graph/helm-chart/renku-graph
```
leafty: ?

charts/renku/requirements.yaml
```diff
@@ -15,6 +15,10 @@ dependencies:
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
   version: "0.3.0"
   condition: gitlab.enabled
+- name: renku-graph
+  repository: file://../../../renku-graph/helm-chart/renku-graph
```
leafty: ?

scripts/minikube_deploy.py
```diff
@@ -64,49 +72,89 @@ def main():
...
-            '--set', 'notebooks.jupyterhub.hub.extraEnv.GITLAB_URL=http://{mip}/gitlab'.format(mip=minikube_ip()),
-            '--set', 'notebooks.jupyterhub.hub.services.gateway.oauth_redirect_uri=http://{mip}/api/auth/jupyterhub/token'.format(mip=minikube_ip()),
-            '--set', 'notebooks.gitlab.registry.host=10.100.123.45:8105',
-            '--set', 'gitlab.registry.externalUrl=http://10.100.123.45:8105/',
-            '--timeout', '1800',
+            '--namespace',
+            namespace,
+            '-f',
+            os.path.join(renku_chartpress_dir, 'minikube-values.yaml'),
+            '--set',
```
leafty: if we're expanding this, let's use `set-string`

jachro:
With my minimal knowledge of helm I can say that all looks good.

leafty:
For jena, it might work to just route to a single dataset endpoint like `jena-renku`

rokroskar:
@leafty:
> For jena, it might work to just route to a single dataset endpoint like jena-renku

yes, that's what I was looking at as well. But I think there is a way to specify the base URL in the global config. We may end up with more datasets in the future which would make this a pain.

charts/renku/requirements.lock
rokroskar: it is, I forgot to put back the registry URL

leafty:
From what I got from the fuseki docs is that there are a bunch of RDF files for configuration :roll_eyes:

rokroskar:
yes, it's rather opaque for me atm. `prefix` means rdf prefix but also (I think) URL prefix...

scripts/minikube_deploy.py
rokroskar: I didn't realize that was an option -- fixed in c7d0cf2

charts/minikube-values.yaml
```diff
@@ -41,6 +41,7 @@ gitlab:
   demoUserIsAdmin: true
 
 notebooks:
+
```
jirikuncar:
```suggestion
```

PR #494, opened 2018-12-21
WIP Example configuration: renkulab.io backend.

ciyer:


docs/developer/example-configurations/renkulab.rst
```diff
@@ -0,0 +1,98 @@
...
+For each created project you will have to modify the :code:`.gitlab-ci.yaml`
+in the repository for the image build to work. In the image_build job
+
+- remove the :code:`image-build` tag
+- add the following entry:
+
+.. code-block:: console
+
+  services:
+    - docker:dind
```
rokroskar: does this actually work with the runner configured on renkulab.io?

docs/developer/example-configurations/renkulab.rst
```diff
@@ -0,0 +1,98 @@
...
+| Client ID         | [Application Id from Step 1]               |
++-------------------+--------------------------------------------+
+| Client Secret     | [Secret from Step 1]                       |
++-------------------+--------------------------------------------+
+
+
+Save the identity provider and logout from the Keycloak admin panel.
+
+Step 4
+--------
```
ableuler: On renkulab Step 4 should not be necessary. If I remember correctly the runners worked "out of the box" when I tried with renkulab.

docs/developer/example-configurations/renkulab.rst
ciyer: Need to test -- I had copied this section from the gitlab.com configuration documentation,

docs/developer/example-configurations/renkulab.rst
ciyer: I've tested this and you are correct. I have updated the documentation.

docs/developer/example-configurations/renkulab.rst
```diff
@@ -0,0 +1,91 @@
...
+| Client ID         | [Application Id from Step 1]               |
++-------------------+--------------------------------------------+
+| Client Secret     | [Secret from Step 1]                       |
++-------------------+--------------------------------------------+
+
+
+Save the identity provider and logout from the Keycloak admin panel.
+
+Use
+--------
```
rokroskar: sphinx usually complains about underlines that are too long/short

charts/example-configurations/minikube-values-renkulab-template.yaml
```diff
@@ -0,0 +1,82 @@
...
+        cryptoKey: 86282f03d9886a64a46ee38f946e9ee27a600df4559584eeb90d0bfbd5a3dc0e
+      type: gitlab
+      gitlab:
+        clientId: # Put Application Id here!
+        clientSecret: # Put Application Secret here!
+        callbackUrl: '""' # Forces the default callback url
+
+gateway:
+  secretKey: 5495ef917488a19d1d4106452c899f5824972acd87edce11fd65e7956bea3bb0
+  gatewayEnv: development
```
ableuler: Actually you also have to update the gateway version in the requirements file (for example to `0.3.0-e6dc1b1`). Otherwise the `gatewayEnv` will not be picked up.

rokroskar:
@ciyer can we wrap this up?

charts/example-configurations/minikube-values-renkulab-template.yaml
ciyer: Added a note to the documentation saying that the gateway version should be >= 0.3.1.

ciyer:
@rokroskar I think I have address all outstanding issues now. Please review.

PR #498, opened 2019-01-16
Developer setup documentation update

jachro:
addresses #501

### DIFF
--- a/docs/developer/python_setup.rst
+++ b/docs/developer/python_setup.rst
@@ -1,0 +1,43 @@
+.. _python_setup:
+
+Python setup
+============
+
+**Renku** requires ``python`` 3.6+.
+
+
+Installation (MacOS users)
+--------------------------
+
+This setup assumes you've got ``Homebrew`` installed in your computer. If not please look `here <https://brew.sh>`_.
+To check which version of python you've got already installed you can use:
+
+.. code-block:: console
+
+    $ python -version
+
+If you have python 2 you can consider uninstalling it with
+
+.. code-block:: console
+
+    $ brew uninstall --force python@2
+
+Before installing Python 3 add a following export line to ``~/.profile``
+
+.. code-block:: console
+
+    $ export PATH="/usr/local/opt/python/libexec/bin:$PATH"
+
+and install ``xcode-select``
+
+.. code-block:: console
+
+    $ xcode-select --install
+
+To install Python:
+
+.. code-block:: console
+
+    $ brew install python
+
+To check if Python 3 gets installed correctly do ``python --version`` again.
--- a/docs/developer/python_setup.rst
+++ b/docs/developer/python_setup.rst
@@ -22,7 +22,7 @@ If you have python 2 you can consider removing it with
 
     $ brew uninstall --force python@2
 
-Before installing Python 3 add a following export line to ``~/.profile``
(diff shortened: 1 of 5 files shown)


docs/developer/python_setup.rst
```diff
@@ -0,0 +1,43 @@
...
+
+    $ python -version
+
+If you have python 2 you can consider uninstalling it with
+
+.. code-block:: console
+
+    $ brew uninstall --force python@2
+
+Before installing Python 3 add a following export line to ``~/.profile``
```
rokroskar:
```suggestion
Before installing Python 3 add the following line to ``~/.profile``
```

docs/developer/python_setup.rst
```diff
@@ -0,0 +1,43 @@
...
+
+.. code-block:: console
+
+    $ export PATH="/usr/local/opt/python/libexec/bin:$PATH"
```
rokroskar: I don't have this in my `PATH` -- why is it needed?

docs/developer/setup.rst
```diff
@@ -79,6 +84,25 @@ Once minikube is started, make sure you can access it by running:
...
+    $ minikube addons enable coredns
+    $ eval $(minikube docker-env)
+
+Installing Chartpress
+^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: console
+
+    $ pip install chartpress
+    $ chartpress --tag latest
```
rokroskar: This isn't about installing chartpress, but about building the images. So this is probably unnecessary here. But you may say two words at the beginning of this section about chartpress, e.g.

"We use the `chartpress` package to bundle together images and charts."

docs/developer/python_setup.rst
jachro: I'm not entirely sure but I think this is to override the default osx' python so the brew installed python is kind of hiding the system one.

docs/developer/setup.rst
jachro: Yes, that's true but I guess `pip install chartpress` should stay here maybe?

docs/developer/python_setup.rst
jirikuncar: IMHO it's not necessary. If there is a problem with Python binary installed by brew, one should just run `brew doctor`.

docs/developer/setup.rst
rokroskar: Actually, we use `pipenv` anyway with the provided `Pipfile` -- that's what the `make minikube-deploy` will use. So this section is not really necessary IMO.

jachro: I'm sure you know better how the Python machinery works here. I took the section out.

docs/developer/python_setup.rst
rokroskar: I think @jirikuncar's point is that brew should take care of making sure that the brew-installed versions are the ones that are actually used. But you might be right that it doesn't always work out that way...

jachro: My experience is that when it comes into binaries both provided by OSX natively and installed with `brew`, the default is that the system ones are winning as they are on the `PATH` and the additional ones not necessarily. And the problem is that for `brew` everything is fine as from its perfectly and installation wise all is good. But maybe I'm doing something wrong.

rokroskar:
Thanks @jachro! Please squash before merging.

PR #506, opened 2019-02-05
Fixed naming of the webhook-service's event consuming endpoint

jachro:
I reminded a suggestion that `webhook-service`'s endpoints names are a bit broken. I fixed that as I think it's easier to do it now.


rokroskar:
Note that this will break existing web hooks...

jachro:
I know but because `webhook-service`'s endpoints are not integrated into UI, there should be no webhooks pointing to the old endpoint except the ones created manually.

PR #510, opened 2019-02-14
add some basic docs about the graph service(s)

rokroskar:
This PR adds some basic documentation including an architecture sketch. 

Check it out with 

```
cd docs
pipenv run make html
open _build/html/index.html
```


jachro:
Any ideas what could be the reason for such error when calling the `make`:
`xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun`

rokroskar:
@jachro: https://apple.stackexchange.com/questions/254380/why-am-i-getting-an-invalid-active-developer-path-when-attempting-to-use-git-a ?

rokroskar:
You need to `brew install graphviz` I imagine

jachro:
Thanks, that did the job!

On the `services_architecture.html` graph I don't know how much details we would like to have but there's also a link between `Gateway` and `Graph service` which is not present. 

On the `graph_service.html` `Gateway` is pointing to `Gitlab` for registering a hook but in real `Gateway` reaches `Gitlab` through `webhook-service`. Again, if we'd like to catch all the details we can change the graph but in general, the essence is caught.

Apart from these two, all look good!

rokroskar:
@jachro can you approve the PR if there are no more changes to be made?

PR #511, opened 2019-02-14
Added config for the token-repository service

jachro:


jachro:
OK. I'll add it once renku-graph is built

jachro:
`requirements.yaml` updated

PR #512, opened 2019-02-14
Added hookToken.secret value required during webhook-service deployment

jachro:
This secret I forgot the other day and effectively all the environments would be using the same secret for encrypting the hook tokens send from Gitlab


jachro:
Could someone please click the `Merge` button as it seems I don't have enough power to do that.

PR #522, opened 2019-03-07
Prepare 0.3.1 release

rokroskar:


charts/renku/requirements.yaml
```diff
@@ -19,6 +19,7 @@ dependencies:
   alias: graph
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
   version: 0.1.0-aa0b94b
```
ableuler: Does it make sense to include the graph at all before there has been a first graph release?

rokroskar: Yes because we're using it for development. Otherwise I have to make a release branch just for this.

ableuler: I agree, that would be a bit annoying.

jachro: Could we then go with version `0.1.0-36eedf7`? `0.1.0-aa0b94b` is pretty old and differs not just in functionality but also in the way dataset is and postgres db are created.

rokroskar:
Ok I filled in some fake values for the graph chart. This is the only way to make the chart render correctly without requiring that, if the graph is turned *off*, the user fills in those values. We may want to add a conditional in the graph chart for this in the future to make it a bit cleaner.

charts/renku/values.yaml
```diff
@@ -580,26 +580,37 @@ gateway:
...
-#       secret:
-
-#   triplesGenerator:
-#     play:
-#       secret:
-
-#  tokenRepository:
-#    tokenEncryption:
-#      secret:
+  ## Configuration for Apache Jena
```
jachro: @rokroskar why do we need to define these values at all? I thought that once `graph` is disabled we don't have to have any values for it and thus has the yaml files a bit less noisy. Unfortunately, there's this one mandatory postgres password value in the `global` section but my thinking was that's the only one.

rokroskar: Because it's defined as a dependency in `requirements.yaml` the chart gets pulled and placed in the charts directory. It's rendered with all the rest of the charts, it's just not deployed. So it fails in the rendering step if the values are missing.

rokroskar: We should put a conditional in the `secrets.yaml` definitions in the graph chart to skip them if `enable: false` is set. So this here is a place-holder solution without having to change the graph chart for the moment.

jachro: Right, in that case, we have no choice. It's a bit unfortunate, though.

rokroskar: I agree -- lets do it this way for now and when we iterate on the graph chart add in the conditional. Could you make an issue for it?

charts/renku/values.yaml
```diff
@@ -51,17 +51,17 @@ global:
...
-      ## Name of the postgres user to be used to access the db storing access tokens
-      postgresUser: tokenstorage
-      ## Postgres password to be used to access the db storing access tokens
-      ## Generated using: `openssl rand -hex 16`
-      postgresPassword:
-    jena:
-      ## specify which dataset should be used in Jena
-      dataset:
+    version: 0.3.1
+  # graph:
```
jsam: Should we clean this up or do we plan to plug it back in? :smile:

charts/renku/templates/configmap.yaml
```diff
@@ -156,6 +157,8 @@ data:
         grant all privileges on database "$TOKEN_REPOSITORY_DB_NAME" to "{{ .Values.global.graph.tokenRepository.postgresUser }}";
         grant all privileges on schema "public" to "{{ .Values.global.graph.tokenRepository.postgresUser }}";
       EOSQL
+  {{- end }}
+
```
jirikuncar:
```suggestion
```

charts/renku/values.yaml
rokroskar: ah thanks, I'll remove the comments

PR #525, opened 2019-03-12
Adapt to nginx-unprivileged base image.

ciyer:
Merge together with https://github.com/SwissDataScienceCenter/renku-ui/pull/399.


ciyer:
This PR is no longer necessary because of the exposed port in renku-ui does not change.

PR #528, opened 2019-03-13
git lfs migrate import

erbou:
Another useful git lfs command for fixing the repository after accidentally committing large data files


jirikuncar:
It looks like a section that should be also in `renku.cli.storage`.

rokroskar:
@jirikuncar the `renku doctor` command should check the output of `git lfs migrate info`

PR #535, opened 2019-03-28
Add pipx instructions

rokroskar:


docs/user/cli-installation.rst
```diff
@@ -17,19 +17,53 @@ It is easiest to install the CLI using `homebrew <https://brew.sh/>`_:
...
+  i.e. the default environment you use when you launch a JupyterLab session
+  via the renku browser interface.
+
+`Install pipx <https://github.com/pipxproject/pipx#install-pipx>`_
+and make sure that the ``$PATH`` is correctly configured.
+
+::
+
+    $ python3 -m pip install --user pipx
+    $ pipx ensurepath
```
jsam: This part is not gonna work on every environment. After installation with command as follows, pipx is not on the path for example on alpine and it will result in following error: ```/bin/sh: pipx: not found```. 

To make things a bit easier to follow for non-unix experts let's add the whole output of pipx ensure path to the docs.

```
/root/.local/bin/pipx ensurepath

Note:
To finish installation, '/root/.local/bin' must be added to your PATH
environemnt variable.

To do this, add the following line to your shell
config file (such as ~/.bashrc if using bash):

    export PATH=/root/.local/bin:$PATH
```

rokroskar: ? most people aren't `root`

docs/user/cli-installation.rst
```diff
@@ -17,19 +17,53 @@ It is easiest to install the CLI using `homebrew <https://brew.sh/>`_:
...
+
+Once ``pipx`` is installed use following command to install ``renku``.
+
+::
+
+    $ pipx install renku
```
jsam: On alpine this command results in 

```
/ # pipx install renku
Command "/root/.local/pipx/venvs/renku/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-zar2sap7/lxml/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-record-tdez0xs_/install-record.txt --single-version-externally-managed --compile --install-headers /root/.local/pipx/venvs/renku/include/site/python3.6/lxml" failed with error code 1 in /tmp/pip-install-zar2sap7/lxml/
'/root/.local/pipx/venvs/renku/bin/python -m pip install renku -q' failed
```

Any idea why?

jsam: (Following the docs) On alpine this command fails with:

```
/ # pipx install renku
Command "/root/.local/pipx/venvs/renku/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-zar2sap7/lxml/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-record-tdez0xs_/install-record.txt --single-version-externally-managed --compile --install-headers /root/.local/pipx/venvs/renku/include/site/python3.6/lxml" failed with error code 1 in /tmp/pip-install-zar2sap7/lxml/
'/root/.local/pipx/venvs/renku/bin/python -m pip install renku -q' failed
```

Should we add information about which binary packages are needed by renku?

rokroskar: I think we need a section for troubleshooting and this can go there - can you make an issue for it?

PR #554, opened 2019-05-03
docs: remove outdated information

ableuler:


docs/developer/general/json_web_tokens.rst
```diff
@@ -35,28 +34,15 @@ Relevant documentation
...
-------------
 
-In the scope of the platform, JWTs are used as bearers of proof. In simple
-terms, a valid token holds the proof of the claims it contains.
+Usage in Renku
+--------------
 
-`Access Tokens <https://tools.ietf.org/html/rfc6749#section-1.4>`_ from the
-identity manager (keycloak) are used to prove:
+Keycloak serves as an identity provider to GitLab. Personal `access
```
rokroskar: I would say that it serves as an ID provider to *renku* not just gitlab (although I realize gitlab is the only place it's used at the moment)

docs/developer/general/json_web_tokens.rst
```diff
@@ -35,28 +34,15 @@ Relevant documentation
...
+<https://tools.ietf.org/html/rfc6749#section-1.4>`_- and `identity tokens
+<http://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken>`_ from
+Keycloak for a user are collected during the log-in process. However, there is
+currently no Renku backend service which uses the Keycloak access token for
```
rokroskar: Is it necessary to say that no service is using jwt's?

PR #560, opened 2019-05-07
docs: enable live reload during editing

ableuler:
The question asked by @cchoirat reminded me that the setup I use for editing docs might be useful for others too:

```
rerun --ignore _build 'make html' &
httpwatcher --root ./_build/html/ 
```
gives you live updates in your browser when editing the docs. If people think it's an overkill I can keep this for myself, otherwise I will add a bit of documentation to it.


rokroskar:
I usually just rebuild the docs and reload the file... so this will update as you type?

ableuler:
> I usually just rebuild the docs and reload the file... so this will update as you type?

Depending on your editor you might have to save, but essentially yes.

ableuler:
> actually I take it back - could you add some words to this effect in the developer docs?

sure, that was the plan.

docs/developer/contributing/documentation.rst
```diff
@@ -0,0 +1,52 @@
+.. _documentation:
+
+Contributing to Documentation
```
rokroskar:
```suggestion
Contributing to the Documentation
```

rokroskar: Oops sorry about the underline...

PR #565, opened 2019-05-09
nginx-ingress: avoid `proxy-request-buffering: off`

fgeorgatos:
This PR lives in a grey zone; please pause to discuss merits before reject/accept.

Let's refrain to explain *why* `proxy-request-buffering` is there - that is NOT the issue.


rokroskar:
I would be in favor of closing this PR since it's a yaml processing issue.

fgeorgatos:
agreed, once the below PR got in, we should be able to avoid commenting out the line, closing:
https://github.com/SwissDataScienceCenter/chartpress/pull/1/files

PR #573, opened 2019-05-22
docs: add a very basic doc about the KG

### LINKED ISSUE #539: document privacy issues for publishing graph data
Add documentation to the lineage page of the main docs that explain what is shared if a private project decides to export its lineage to the knowledge graph.

rokroskar:
closes #539


docs/user/knowledge-graph.rst
```diff
@@ -0,0 +1,37 @@
...
+
+In short: if you enable the Knowledge Graph integration on a private project,
+the *metadata* about your project will become publicly accessible. Important:
+your *data* is in no way affected by this - the contents of your project are
+still private. What is metadata then? Things like filenames, paths within the
+project, people working on the project, project directory structure etc.  This
+is especially true if your private project references a public project - a
+lineage query will reveal the workflow steps taken within your private project
+using the public data. If sharing this level of metadata is not acceptable to
+you, then please do not turn on the Knowledge Graph integration at this stage.
```
lorenzo-cavazzi: At the moment the user can choose to **not** activate the lineage for old projects, but this is not possible for new projects. This sentence is misleading since the user would probably think that he can always opt out

rokroskar: can't opt out on new private projects?

lorenzo-cavazzi: No. Should the user have this option?

rokroskar: I assumed that was the case since it's opt-in for old projects as well.

lorenzo-cavazzi: I believe we spoke about this and we decided to simply add a warning for private projects. My understanding was that KG is a main feature of Renku and it should be enabled, otherwise we may lose important info when lineage will be cross-projects.

Anyway adding an option to opt-out from the KG is easy, but the user experience for non-indexed projects should probably change, otherwise an annoying popup keeps showing every now and then during the navigation.

docs/user/knowledge-graph.rst
```diff
@@ -0,0 +1,48 @@
...
+project, people working on the project, project directory structure etc.  This
+is especially true if your private project references a public project - a
+lineage query will reveal the workflow steps taken within your private project
+using the public data. If sharing this level of metadata is not acceptable to
+you, then please disable the Knowledge Graph integration for now.
+We are working on providing comprehensive access control on the graph queries
+that traverse projects, but at this stage they are not yet implemented. 
+While the Knowledge Graph integration is by default enabled on all new projects 
+regardless of privacy settings, it can be disabled by deleting the webhook 
+under ``Integrations`` in a project'ssettings in Gitlab.
```
lorenzo-cavazzi: missing space in `project'ssettings`

rokroskar: I probably shouldn't change prs on my phone...

lorenzo-cavazzi:
This adds the possibility to opt-out for new private projects SwissDataScienceCenter/renku-ui#469
At this point the previous version of this PR was probably more accurate, sorry for the mess :sweat_smile:

rokroskar:
no worries! thanks for changing that so quickly

rokroskar:
You're right, I thought it was enough that it's implicit but I can make it explicit.

docs/user/knowledge-graph.rst
```diff
@@ -0,0 +1,68 @@
...
+In short: if you enable the Knowledge Graph integration on a private project,
+the *metadata* about your project will become publicly accessible.
+**Important**: your *data* is in no way affected by this - the contents of your
+project are still private. What is metadata then? Things like filenames, paths
+within the project, people working on the project, project directory structure
+etc.  This is especially true if your private project references a public
+project - a lineage query will reveal the workflow steps taken within your
+private project using the public data. If sharing this level of metadata is not
+acceptable to you, then please do not enable the Knowledge Graph integration for
+now. For new private (or internal) projects, you will be given an option when
```
lorenzo-cavazzi: It's possible to opt-out only for private projects. Should internal projects also be included? Discussing with @ciyer it seemed a reasonable choice not to include them -- data are already not private

rokroskar: yes this is fine IMO.

lorenzo-cavazzi: I would then remove `(or internal)` at line 62.

PR #577, opened 2019-05-27
build(ci): deploy tags

rokroskar:
One more fix for new travis behavior...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,9 +75,15 @@ script:
   - ./scripts/run-tests.sh
 
 deploy:
-  - provider: script
+  - &script
+    provider: script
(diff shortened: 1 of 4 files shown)


rokroskar:
We do - except for here apparently. :)

rokroskar:
btw a tagged build is running right now... if it completes without issue I'll merge

ableuler:
> We do - except for here apparently. :)

Ah true - has been there for only 6 months 🙈

rokroskar:
@ableuler yes I put this in everywhere except for this repo because of the annoying circular dependency on our gitlab chart. We should factor out the gitlab chart as well and then everything will be fine.

PR #578, opened 2019-06-05
minor(helm readme): use star instead of the exact version

jkremser:
this way it can be copy&pasted w/o changing the string


rokroskar:
won't you get in trouble if you happen to fetch more than one version?

jkremser:
wow, that was quick :) good point
well, I guess something like `ls renku-*.tgz | sort | head -1` would be safer, but also more complicated

```
helm upgrade --install renku --namespace renku \
        -f minikube-values.yaml \
        --set global.renku.domain=$(minikube ip) \
        --set ui.gitlabUrl=http://$(minikube ip)/gitlab \
        --set jupyterhub.hub.extraEnv.GITLAB_HOST=http://$(minikube ip)/gitlab \
        `ls renku-*.tgz | sort | head -1`
```

btw. I am reading through the communication in your slack channel and trying to catch up everything.

rokroskar:
Actually this doc should be changed completely I think - fetching the charts and deploying is probably not the way to go. A better way is to do something like:

```bash
helm repo update
helm upgrade --install renku renku/renku --namespace renku \
        -f minikube-values.yaml \
        --set global.renku.domain=$(minikube ip) \
        --set ui.gitlabUrl=http://$(minikube ip)/gitlab \
        --set jupyterhub.hub.extraEnv.GITLAB_HOST=http://$(minikube ip)/gitlab \
        --version renku-0.X.Y-12345
```

jkremser:
right, I was able to use helm chart on the current working directory with 

```
helm template --set env.crd=true . | kubectl apply -f -
```
so it's another way to consider..

rokroskar:
ha! yes I guess that will work, but then you are doing the job of `helm` with unix pipes :)

It would be great to update this doc to be more in-line with how we actually expect people to deploy renku from published charts -- so I think the `--version` flag should be the way to go.

rokroskar:
Sorry, I hijacked this PR and made some more changes to the README. Please have a look @jkremser, I hope it makes sense?

rokroskar:
Oh actually I just realized this will never pass CI because it can only use encrypted variables from PRs that come from members of the org. I will close this PR and make a new one.

PR #579, opened 2019-06-06
Use new postgres chart and cleanup values

rokroskar:
The postgres chart has changed significantly - we should consider upgrading. 

Benefits:

* HA mode
* doesn't run as root
* easier init configuration 

*need to test upgrading from old chart first*!!!


ableuler:
Closing in favour of #1285. Changes of this PR to the values file are reflected there.

PR #580, opened 2019-06-06
Remove registry secret

rokroskar:
Move the registry secret to the renku-notebooks repo - see https://github.com/SwissDataScienceCenter/renku-notebooks/pull/170 and merge together. 

This also removes the code that adds the gitlab sudo token and the demo user to gitlab. Note that in order to use the sudo token functionality, one must do the following:

0. disable automatic oauth login in the `gitlab.rb` configmap
1. log in to gitlab as `root`
2. create the sudo token
3. copy the sudo token to `global.gitlab.sudoToken`
4. redeploy


charts/renku/values.yaml
```diff
@@ -542,7 +542,7 @@ notebooks:
     #   gitlab:
     #     clientId: jupyterhub
     #     clientSecret: dummy-secret
-    #     callbackUrl: '""' # Forces the default callback url
+    #     callbackUrl:  # Null value forces the default callback url
```
ableuler: Not sure you wanted this in this change in this commit/PR, but it's a useful fix!

charts/renku/requirements.lock
```diff
@@ -24,4 +24,4 @@ dependencies:
   repository: https://kubernetes-charts.storage.googleapis.com
   version: 1.6.0
 digest: sha256:a71bb5d25af250ea5280765b3a433cfeb63dd077ca58ae0816ad07e5bd2157c7
-generated: "2019-05-27T15:27:30.285351+02:00"
+generated: "2019-06-07T00:51:36.835977+02:00"
```
ableuler: I think this doesn't need to be committed.

rokroskar:
@ableuler yes I agree, I wanted to change it to use api calls. However, it's not possible to create a trusted application with the API call and we have no way of extracting the application secret after it's created using the API. I think the solution is to remove all this automatic github configuration altogether and instead document very clearly the steps that need to be taken. Alternatively we can keep it here and offer some sort of "auto-provision" flag that does this shady stuff.

charts/renku/values.yaml
rokroskar: I think I'll leave it here - it causes many headaches.

charts/renku/requirements.lock
rokroskar: removed

charts/renku/values.yaml
ableuler: I agree, I have the same change in some half-baked chart PR too. Good to have it in!

charts/example-configurations/minikube-values-gitlab.yaml
```diff
@@ -72,35 +72,18 @@ notebooks:
...
 graph:
   jena:
     admin:
-      password: 94a635b40afc2cba
+      password: &sparql-password 94a635b40afc2cba
 
   ## if running gitlab on minikube, set on cmd line with --set graph.gitlab.url
-  gitlab:
-    url: https://gitlab.com
+  # gitlab:
```
ableuler: Why this change? I'd assume that `https://gitlab.com` is the correct value here.

rokroskar: oops thanks for catching that. Probably a left-over from some experiment. I also apparently removed some other values which I've now put back.

PR #628, opened 2019-08-20
chore: create a "new release" template

rokroskar:
This adds a "new release" template with a checklist for making a new renku release.


fgeorgatos:
lgtm, but branch is out-of-sync with base

rokroskar:
actually I think I'd like to create other templates so that users aren't confused

PR #647, opened 2019-09-03
Adds secrets generation using pre-install helm hooks

### LINKED ISSUE #637: charts: explore the possibility to use helm-hooks for secret creation
It is described [here](https://medium.com/nuvo-group-tech/move-your-certs-to-helm-4f5f61338aca) how [chart hooks](https://github.com/helm/helm/blob/master/docs/charts_hooks.md) can be used to create and inject secrets on the first install of a chart (thanks to @Panaetius for pointing this out). This could be an alternative to storing a plethora of encrypted automatically created secrets in deployment-repos.

Panaetius:
Closes #637 

Changes the current `charts/renku/templates/secrets.yaml` file to only create the secret object when installing the chart.

Creation of the secret can be force by setting `.Values.global.recreateSecrets` to true.

Secrets can be accessed by running e.g. `kubectl get secret --context switch-dev --namespace $DEV_NAMESPACE ${DEV_NAMESPACE}-renku  -o=custom-columns=Value:data.keycloak-password` for the `data.keycloak-password` value.

Caveats:
- Setting `.Values.global.recreateSecrets` will currently recreate ALL secrets. To allow more fine-grained control we'd need individual `kind: Secret` files for each secret.
- Setting the value of a password that was already generated on install won't have any effect. We could have 2 passwords, one set in `values.yaml` and one auto-generated and only use the auto-generated one if it wasn't set manually. Though that'd mean a bit of redesign of the values.
- When adding new secrets values, e.g. credentials for newly added subcharts, they have to be placed in new `kind: Secret` files and run with an override similar to `.Values.global.recreateSecrets` to get executed on upgrade, otherwise new values would only get added when newly installing the chart.


rokroskar:
I like this solution though I haven't had a chance to use it yet. One simple change that will make it work without adding more flags is to put the secret values that can't be updated easily into a separate secret template and keep the previous warnings about the values being required (e.g. keycloak).

Panaetius:
> I like this solution though I haven't had a chance to use it yet. One simple change that will make it work without adding more flags is to put the secret values that can't be updated easily into a separate secret template and keep the previous warnings about the values being required (e.g. keycloak).

Do you mean when upgrading just show a warning that it's missing? The 2nd caveat in the PR description? Or could you elaborate please

rokroskar:
> Do you mean when upgrading just show a warning that it's missing? The 2nd caveat in the PR description? 

Hmm not really. I was referring more to the `{{ if .Values.global.recreateSecrets -}}` conditional which applies to the whole secret file. The secrets that we don't want to automatically generate/regenerate should be in separate files. This would actually be slightly nicer anyway so we don't have all the secrets lumped together. 

However, your question brings up another point. I don't understand the statement 

> only use the auto-generated one if it wasn't set manually

This is already the case isn't it?

Panaetius:
> The secrets that we don't want to automatically generate/regenerate should be in separate files. This would actually be slightly nicer anyway so we don't have all the secrets lumped together.

Agree

> However, your question brings up another point. I don't understand the statement
> 
> > only use the auto-generated one if it wasn't set manually
> 
> This is already the case isn't it?

Not necessarily. If I have a Secret that is generated by a hook on install and didn't manually specify it anywhere, then it uses the autogenerated one. If later I fill in the value in `values.yaml` manually and do an update, it'd still use the autogenerated one, not the one I set, as the secret is not regenerated and the value in `values.yaml` therefore irrelevant. So it silently fails to accept the new, manually set credential. All solutions to this that I can think of are quite hacky.

Also if a secret is missing (e.g. newly added Secret that wasn't created because we ran helm upgrade not install), we should write something like `Secret XYZ is required, create it once ba running helm upgrade with --set create-xyz=true` to prompt the user. 

I also looked into other ways of autogenerating secrets for helm/k8s and while there are many solutions to storing secrets (helm-secrets, other solutions based on SOPS, I didn't really find anything that'd fit our use-case, short of writing our own helm plugin.

rokroskar:
Right, I didn't think about manually changing it and effectively having no clear way of enforcing that the change goes through unless you force the regeneration of all secrets. OTOH for the secrets that can be changed dynamically, forcing a regeneration doesn't really matter. For the others, we could have component-specific switches? Like `keycloak.secret.generate: true` or something similar. It would be nice if we could issue a warning if a secret set through a value is being ignored because of the helm hooks, however.

Panaetius:
> For the others, we could have component-specific switches? Like `keycloak.secret.generate: true` or something similar. It would be nice if we could issue a warning if a secret set through a value is being ignored because of the helm hooks, however.

Yes we can, though the switches can only be at the level of the Secret, not of individual data.* entries, so we would have to have one Secret per concern (e.g. keycloak) or even one Secret per value.

Panaetius:
I added a new `keycloak-secrets.yaml` that contains the keycloak screts and is genereted on install, if `.Values.keycloak.secret.generate == true` or if `.Values.keycloak.password` or `postgresPassword` are set.

It doesn't seem as if there is a way in helm or K8s to test if a secret exists or not, so warning for the case that it's not created doesn't work and K8s will just repeatedly try to access the secret, waiting for it to be created :cry:

fgeorgatos:
just a question, what should we expect to happen upon roll-back/roll-forward?

fyi. note that helm 3 will come with json schema validation, so depending on how it is enforced it could help in certain situations

Panaetius:
> just a question, what should we expect to happen upon roll-back/roll-forward?

That's a good question. 

Without `pre-rollback` hook set, it does nothing to a secret on rollback. With `pre-rollback` set it will return to the previous value on rollback. I've added `pre-rollback` to the template so it returns to previous values, but I think it'll depend on the value in question. 
If the secret is for some external API, we don't want it to change on rollback, if it to some value that is specific to the release (e.g. communication password for 2 pods), we probably want to roll it back.

In a similar vein, without the `pre-delete` hook set, the secret won't get deleted on delete of a release, with it set it will be deleted. This could be useful to keep secrets when deleting and reinstalling an release.

It is something we need to think about for each value.

rokroskar:
Where do we stand with this PR? Is there a deployment up on dev with this setup?

rokroskar:
Is this mergeable with #662? Would be great to merge these together asap.

Panaetius:
It should be mergeable with #662 since `realm.json` isn't touched in this PR

charts/minikube-values.yaml
```diff
@@ -15,23 +15,29 @@ ingress:
 global:
   gitlab:
     clientSecret: e3ecd740-0892-4610-ad69-c7f4886f4533
-    postgresPassword: 3b0aa4bae4fbbe167143f583e35eda18
```
ableuler: It would be much nicer to have all the secrets in this file auto-created. Having all those random strings in here has been bothering me anyway and leaves the impression of Renku being a pain to deploy.

charts/ralf-dev.values.yaml
```diff
@@ -0,0 +1,253 @@
+global:
```
ableuler: I think this file made it into the PR by accident ;)

charts/gitlab/templates/secret.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
 type: Opaque
 data:
   gitlab-password: {{ required "Fill in .Values.password with `openssl rand -hex 16`" .Values.password | b64enc | quote }}
-  postgres-password: {{ required "Fill in .Values.global.gitlab.postgresPassword with `openssl rand -hex 16`" .Values.global.gitlab.postgresPassword | b64enc | quote }}
+  #postgres-password: {{ required "Fill in .Values.global.gitlab.postgresPassword with `openssl rand -hex 16`" .Values.global.gitlab.postgresPassword | b64enc | quote }}
```
ableuler: Do you want to keep this despite it's commented out?

charts/renku/values.yaml
```diff
@@ -58,20 +66,27 @@ global:
...
+        value:
+        overwriteOnHelmUpgrade: false
+      existingSecret: '{{ template "renku.fullname" . }}-token-postgres'
     jena:
       ## specify which dataset should be used in Jena
       dataset:
   ## Set to true if using https
   useHTTPS: false
 
+  ## Set to true to recreate secrets (credentials etc.) when updating the chart
```
ableuler: Looks like this is not used as this option can now be set on each secret individually.

charts/renku/values.yaml
```diff
@@ -563,6 +580,20 @@ notebooks:
...
         #  apiToken: # use `openssl rand -hex 32`
         #  oauth_redirect_uri: http://<gateway>/api/auth/jupyterhub/token
+        ## Configure gitlab and postgres:
+        ## Values need to be set to get jupyterhub to work
+       # extraEnv:
+       #  - name: GITLAB_URL
+       #    value: <gitlab url>
+       #  - name: DEBUG
+       #    value: "1"
+       #  - name: JUPYTERHUB_SPAWNER_CLASS
```
ableuler: Uncommenting this one to make it a default would lead to problem with merging the `extraEnv` from two different values files, right?

charts/gitlab/templates/secret.yaml
Panaetius: I honestly don't remember where this is from, `charts/gitlab/` shouldn't be changed here.

charts/minikube-values.yaml
```diff
@@ -15,23 +15,29 @@ ingress:
...
-    sudoToken: 214981ee49479ecbee56359b70ad25dc87edd0e9ea569b7b8c522c3bf979446a
+    postgresPassword:
```
ableuler: Wait, why are the hardcoded values back here? Didn't you remove those @Panaetius?

Panaetius: I'd removed them from the files in `example-configurations`, not this one (must have missed it). Though maybe this should be in `example-configurations` as well?

ableuler: The hardcoded values have been removed from all example configurations by Pamela's recent PR anyway. The example configurations for deployments against existing GitLab are now much more minimal and just contain the which _differ_ from other deployments.

pameladelgado: >Though maybe this should be in example-configurations as well?
not needed. Those files are good as they are and they cover the minimal set of values to overwrite minikube-values.yaml values.
I was actually going to comment on that, we should remove as many hardcoded passwords as we can from the minikube configuration.

charts/minikube-values.yaml
```diff
@@ -65,14 +53,26 @@ notebooks:
...
       services:
         notebooks:
           apiToken: notebookstoken
         gateway:
           admin: true
           oauth_client_id: &gatewayClient gateway
           apiToken: &gatewaySecret 7ed9992425bbdcf03f323310205b2fea34bd9e1a255346f99066800d9440d6f7
+      extraEnv:
+        - name: GITLAB_URL
+          value: https://dev.renku.ch/gitlab
```
pameladelgado: why is this here? this is wrong, we should delete this extraEnv.
The correct value for minikube is set as part of the upgrade command (see README):
`--set notebooks.jupyterhub.hub.extraEnv.GITLAB_URL=http://$(minikube ip)/gitlab \`

Panaetius: Basically:
- To not have the postgres password in the connection string/url, we need to set `PGPASSWORD` in extraEnv
- In order to be able to use a secret value there, we need to set `valueFrom` with a `secretKeyRef`
- `valueFrom` is only available when using `map`representation (`name: <key> value: <value>` instead of `<key>: <value>`) for `extraEnv`, not `list` representation.
- Helm can't merge `map` representations, as it can with `list`, so the highest level that `extraEnv` is defined at HAS to include all the definition of the lower levels, otherwise they'd get lost (overwritten). Which sucks, but is a limitation of helm as far as I can tell.

I don't know enough about our minikube deployment to know the correct values, but we'd need to set it to `http://$(minikube ip)/gitlab` in the `values.yaml` for it to work or use `--set notebooks.jupyterhub.hub.extraEnv[0].value=http://$(minikube ip)/gitlab` on the commandline to modify the single value. But we still need all the `extraEnv`values in every `values.yaml`,  which is a bit annoying.

ableuler: Thanks @pameladelgado for spotting this. Unfortunately, `http://$(minikube ip)/gitlab` in a values file will not work, so `--set notebooks.jupyterhub.hub.extraEnv[0].value=http://$(minikube ip)/gitlab` is the way to go. I adapted the deployment command accordingly.

ableuler:
I found another small problem in this PR. @Panaetius can you confirm that the changes to 
charts/renku/templates/post-install-job-gitlab.yaml
charts/gitlab/templates/secret.yaml
charts/gitlab/templates/deployment.yaml
introduced in 
https://github.com/SwissDataScienceCenter/renku/commit/81b69d4026e68e88b337a3fc24584b1e133eab6f 
are consistent with the rest of this PR?

Panaetius:
> I found another small problem in this PR. @Panaetius can you confirm that the changes to
> charts/renku/templates/post-install-job-gitlab.yaml
> charts/gitlab/templates/secret.yaml
> charts/gitlab/templates/deployment.yaml
> introduced in
> [81b69d4](https://github.com/SwissDataScienceCenter/renku/commit/81b69d4026e68e88b337a3fc24584b1e133eab6f)
> are consistent with the rest of this PR?

I wouldn't know about the gitlab chart version, but the rest of it looks good! Thank you :smile:

PR #652, opened 2019-09-05
Graph services to be deployed separately

jachro:
There was a requirement of Graph Services to have separate deployments. There is a PR on graph-services https://github.com/SwissDataScienceCenter/renku-graph/pull/128 which does the majority of the changes. This PR fixes the ingress routing to the webhook-service's events endpoint.

There's also a tiny PR on `renku-gateway` to adjust the `webhook-service` hostname: https://github.com/SwissDataScienceCenter/renku-gateway/pull/167


jachro:
@ableuler great catch! I wasn't even aware of that config value on the gateway.

PR #662, opened 2019-09-16
214 remove realm template

### LINKED ISSUE #214: keycloak: use API to create new clients and users

ableuler:
This PR removes the full realm export from the helm templates and instead uses the keycloak API to create the necessary clients and users. Most importantly, the PR does two things:

1. It adds a script which can be used standalone to initialize a keycloak instance with the necessary clients and users.

2. This script is called as a part of a post-install k8s job.

Remarks:
- Letting keycloak create application secrets itself and injecting into a secret through the k8s API would be desirable but should be tackled in a [separate issue](https://github.com/SwissDataScienceCenter/renku/issues/665).

- The admin documentation [should be adapted](https://github.com/SwissDataScienceCenter/renku/issues/664).

Merge together with https://github.com/SwissDataScienceCenter/sdsc-dev-docs/pull/45

closes #214


pameladelgado:
some comments:
1. need to delete some lines from values.yaml, I will commit this change later
2. there's the `Register` button missing on the keycloak screen
3. after the gitlab screen where I'm asked to Authorize deployment  I'm getting `Unexpected error when authenticating with identity provider`
It can be checked here https://keycloak-214.dev.renku.ch/

pameladelgado:
I really :heart:  the deletion of https://github.com/SwissDataScienceCenter/renku/pull/662/files#diff-2f4fc8fd4bdff77fa97c27b9f78d6604 :smiley:

ableuler:
Thanks @pameladelgado for catching the registration problem. I've turned on user registration per default. Can you verify that this solves your problem?

pameladelgado:
@lorenzo-cavazzi  yes, you can try to use this version of the template instead: 
https://github.com/SwissDataScienceCenter/sdsc-dev-docs/tree/214-remove-realm-template
or alternatively, just manually delete lines 56 to 61 of your values.yaml file.

But I think this PR is not yet ready for a re-review, right @ableuler ?

lorenzo-cavazzi:
btw a deployment is available at the following link if someone wants to try it: https://lorenzotest.dev.renku.ch

pameladelgado:
> btw a deployment is available at the following link if someone wants to try it:  https://lorenzotest.dev.renku.ch

I'm getting a Bad gateway screen ... I tried in a private windows as well

lorenzo-cavazzi:
@pameladelgado I just tried again through `dev-renku` and it works with `GitHub` and by manually registering a new user. How did you try to log in? Could you try again now?

pameladelgado:
> @pameladelgado I just tried again through `dev-renku` and it works with `GitHub` and by manually registering a new user. How did you try to log in? Could you try again now?

It seems that it was some transient error. Now I can login with dev-core. Thanks @lorenzo-cavazzi !

I will redeploy this again as well because I still want to:
1. compare this keycloak realm with the ones we normally use like in renkulab, there might be some other default settings we want to change
2. make a test of upgrading from the old deployment to this (**important** for our current deployments)
3. test how updates would work

rokroskar:
@lorenzo-cavazzi it looks like the "email as username" setting is not on: 

image

charts/renku/templates/_keycloak-clients-users.tpl
```diff
@@ -0,0 +1,47 @@
...
+  {{- end -}}
+]
+{{- end -}}
+
+{{- define "renku.keycloak.users" -}}
+[
+  {{- if .Values.keycloak.createDemoUser -}}
+  {
+    "username": "demo",
+    "password": "demo",
```
rokroskar: maybe generate a random password and provide the `kubectl` command needed to retrieve it?

ableuler: Since the demo user is not created by default I thought that an "all-open" demo/demo user would be preferable, but if there are strong concerns about this, I'm happy to follow your suggestion.

rokroskar: it makes debugging more annoying, but the user is not removed even it `createDemoUser=false` in subsequent deployments so it's likely to just stay there until someone actively disables/removes it.

ableuler:
> @lorenzo-cavazzi it looks like the "email as username" setting is not on:
> 
> image

I have set "email as username" to `True` per default when I create the Realm.

ableuler:
> 1. compare this keycloak realm with the ones we normally use like in renkulab, there might be some other default settings we want to change

I've done this and adapted a handful of settings to the values that they had in the exported Realm template. If you see a setting that still seems odd to you, let me know.

ableuler:
> Thanks, not it works! 🎉
> 
> Just 2 minor details:
> 
> * After the deploy, the first instance of `lorenzotest-renku-post-install-keycloak` fails and it stays in `Error` status -- a second one succeeds after. Since it's probably related to the slowness of the keylock pod, maybe raising the timeout to 3 minutes could be enough to prevent this.
> 
> ```
> NAME                                                   READY   STATUS      RESTARTS   AGE
> lorenzotest-renku-post-install-keycloak-9vlrn          0/1     Error       0          2m26s
> lorenzotest-renku-post-install-keycloak-tsqkt          0/1     Completed   0          59s
> ```
> 

I let the job try for 5 minutes now before failing.

> * When I try to login through `dev-renku` and I select the GitHub account, I get asked for a username, email, first name, and last name. In the previous configuration, I was asked only for the last 3. It is not particularly relevant, but I guess this behavior is piloted by a setting somewhere. Maybe it's an easy change.
> 
> Screenshot from 2019-09-24 14-33-37

Can you confirm that the "email as username" setting has re-established the previous behaviour?

charts/renku/templates/_keycloak-clients-users.tpl
ableuler: Ok, I did as you suggested. Note that for the moment the correct password for the demo user can only be retrieved using `kubectl` after the initial install. On upgrade the corresponding secret will be overwritten, but not the password of the demo user. This will be fixed when resolving https://github.com/SwissDataScienceCenter/renku/pull/647 which is already in progress.

ableuler:
@pameladelgado 
> Waiting for the addition of the mapper needed for newer versions of keycloak to also test:
> 

done!

rokroskar:
what else is this waiting for? Seems like all issues have been addressed?

pameladelgado:
> what else is this waiting for? Seems like all issues have been addressed?

@rokroskar I just tried quickly the last modification for the new version of keycloak and it seems that the mapper is not correctly created for some reason

ableuler:
> that the mapper is not correctly created for some reason. 

I tried to be careful not to overwrite too many existing settings in keycloak on update. Therefore, the mapper was only created in absence of any mappers. Now I've changed this and I also add the necessary mapper to the already existing ones.

PR #666, opened 2019-09-19
chore: modify values example

lorenzo-cavazzi:
Update the examples in `values.yaml` about `notebooks.serverOptions` according to the changes here https://github.com/SwissDataScienceCenter/renku-notebooks/pull/213/files

### DIFF
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -524,14 +524,20 @@ notebooks:
   #     type: enum
   #     default: 0.5
   #     options: [0.5, 1.0]
-  #   gpu_request:
+  #   mem_request:
   #     order: 3
-  #     displayName: Number of GPUs
-  #     type: int
-  #     default: 0
-  #     options: [0, 4]
-  #   lfs_auto_fetch:
+  #     displayName: Amount of Memory
+  #     type: enum
+  #     default: 1G
+  #     options: [1G, 2G]
+  #   gpu_request:
   #     order: 4
+  #     displayName: Number of GPUs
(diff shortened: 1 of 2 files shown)


charts/renku/values.yaml
```diff
@@ -510,25 +510,31 @@ ui:
...
-  ## serverOptions allow for modifications to the notebook pod resource requests
+  ## serverOptions allow for modifications to the notebook pod resource requests and the UI rendering
   # serverOptions:
-  #   resources:
-  #     cpu_request:
-  #       displayName: Number of CPUs
-  #       type: integer
-  #       default: 1
-  #       range: [1, 4]
-  #     mem_request:
```
Panaetius: Is amount of memory not needed anymore?

lorenzo-cavazzi: Here I added just a few examples to show how the parameters are provided, not sure if I should copy the whole configuration from `notebooks` where the values are provided (non of these parameters are strictly needed for renku to work properly).

rokroskar: hmm but what is the harm in leaving the `mem_request` in there? To figure out how to change it, someone deploying the platform needs to look inside the notebooks service and they might not even be aware that it is possible to modify it.

lorenzo-cavazzi:
@rokroskar true, I can copy all the code defined in "notebooks" if you think it's useful. I intended this as an example but maybe it can be confusing for users not to find here `mem_request` since it's defined somewhere else.

lorenzo-cavazzi:
can't merge by myself since I'm unauthorized :slightly_frowning_face:

Issue #693, opened 2019-10-31
Prep new release 0.5.0


rokroskar:
To release a new version of Renku:

- [x] make sure images in https://github.com/SwissDataScienceCenter/renku-jupyter are using the desired version of renku-python; if so, make a new tag
- [x] update renku project templates to use the newly tagged images
- [x] update the template tag in `values.yaml`
- [x] update `requirements.yaml` with all the relevant versions and run `helm dep update renku` to update `requirements.lock`
- [x] test the deployment with these dependencies in a dev environment
- [x] tag release candidate
- [x] verify release candidate chart with a dev deployment
- [x] make PR with changes
- [ ] after merging the PR, tag and make github release

PR #700, opened 2019-11-06
Prep new release 0.5.0

pameladelgado:
Current status, waiting for new renku-graph chart to do:
- [x] update `requirements.yaml` with all the relevant versions and run `helm dep update renku` to update `requirements.lock`
- [x] test the deployment with these dependencies in a dev environment
- [x] tag release candidate
- [x] verify release candidate chart with a dev deployment
- [x] make PR with changes
- [ ] after merging the PR, tag and make github release

@rokroskar I added a changelog file with a draft for the release notes, could you please check it and suggest better structure, wording, more accurate descriptions and any other thing you think can improve? 

Addresses #693


charts/values.yaml.changelog.md
```diff
@@ -19,7 +20,7 @@ Please follow this convention when adding a new row
 * EDIT - *global.graph.dbEventLog.postgresPassword*: now split into `value` and `overwriteOnHelmUpgrade`, set the latter to true to generate a password or set value to use an existing one.
 * EDIT - *global.graph.tokenRepository.postgresPassword*: now split into `value` and `overwriteOnHelmUpgrade`, set the latter to true to generate a password or set value to use an existing one.
 * EDIT - *notebooks.jupyterhub.hub.db.url*: remove the password (between ":" and "@" from the URL).
-* EDIT - *notebooks.jupyterhub.hub.db.extraEnv*: due to the inability of helm to merge maps, all extra environment variables must be explicitly provided after the introduction of secrets auto-generation:
+* EDIT - *notebooks.jupyterhub.hub.db.extraEnv*: due to the inability of helm to merge maps, all extra environment variables must be explicitly provided after the introduction of :
```
rokroskar: why delete this?

rokroskar:
Thanks @pameladelgado it's looking good! I just made a few minor edits.

charts/values.yaml.changelog.md
pameladelgado: oops that must have been lost with the merge. Fixed it now.

pameladelgado:
The dependencies are deployed in https://helm-hooks.dev.renku.ch/ and the release candidate with `0.5.0-rc1` in https://stable.dev.renku.ch
This PR should be ready to review/merge

CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
...
+* Simplified deployment with automatic secrets generation
+
+
+### Individual components
+
+For changes to individual components, check:
+* renku ui [0.6.4](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.6.4), [0.7.0](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.7.0) and [0.7.1](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.7.1)
+* renku-gateway [0.6.0](https://github.com/SwissDataScienceCenter/renku-gateway/releases/tag/0.6.0)
+* renku-python [0.7.1](https://github.com/SwissDataScienceCenter/renku-notebooks/releases/tag/0.7.1)
+* renku-graph [0.24.3](https://github.com/SwissDataScienceCenter/renku-graph/releases/tag/0.24.3)
```
lorenzo-cavazzi: `0.24.6`

CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
...
+* Images and project templates now use Renku [0.7.1](https://github.com/SwissDataScienceCenter/renku-python/releases)
+* User profile redirects to Keycloak profile
+* Simplified deployment with automatic secrets generation
+
+
+### Individual components
+
+For changes to individual components, check:
+* renku ui [0.6.4](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.6.4), [0.7.0](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.7.0) and [0.7.1](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.7.1)
+* renku-gateway [0.6.0](https://github.com/SwissDataScienceCenter/renku-gateway/releases/tag/0.6.0)
```
lorenzo-cavazzi: `0.6.1`

pameladelgado: renku-gateway is correct: `0.6.0`, I guess you got confused with renku-notebooks which is `0.6.1` and is not there...
I will add the notebooks to the list of components in the CHANGELOG :+1:

CHANGELOG.md
pameladelgado: right, I guess now it will be `0.24.7`.
Will wait for that release to update both CHANGELOG and requirements.

charts/values.yaml.changelog.md
```diff
@@ -7,6 +7,7 @@ Please follow this convention when adding a new row
 ----
 
 ## Changes on top of Renku 0.4.3
+* ADD - *ui.templatesRepository.ref*: added `0.1.3` to update project templates reference. Temporary value that can be deleted for next release.
```
lorenzo-cavazzi: I believe the change here is an "EDIT" and it's about commenting the values instead of pinning a specific version (that happened in the ui).

> EDIT - \*ui.templatesRepository\*: values are commented out and left for reference in case someone wants to override the values now coming from the `ui` repo

Sorry for not adding it in the relevant PR :sweat_smile: I'll add here a commit with this, feel free to further modify the text

pameladelgado: Sorry @lorenzo-cavazzi , we actually need to revert to what was there before, I will make another commit for this since I see you already pushed the changes yesterday. 
The modification you made is true only for dev deployments. The rest of the deployments don't have those lines defined so it's an ADD. Furthermore, the value `0.1.3` was updated after the renku-ui release so if we dont add that line the default value is `0.1.2` which is incorrect.

lorenzo-cavazzi: Ok sorry for pushing a wrong change, I should have limited to the comment.
I thought the changelog file is meant to record the changes done to charts file (and I should have included them here https://github.com/SwissDataScienceCenter/renku/pull/702/files) but now I see you are using it to record the actions to be done in specific deployments.

Anyway, I am not sure why this change should bring to an `ADD` anywhere. I think inheriting the correct value from the `renku-ui` values file is the safest thing, I would instead `REMOVE` any pinned version from custom values files so that we won't have similar problems in the future.

pameladelgado: The changelog is not meant to be used to record actions to be done in specific deployments, I'm sorry if I wasn't clear.
In this particular case we need to put version `0.1.3` of the templates there because is not part of renku-ui `0.7.1` which will take `0.1.2`.  In the changelog for next release we want to delete this value to take the default (or leave to what the admin wants to use).
However I do think the content in changelog could be more generic depending on whether a value is there or not. For this change, in case a value is there it would be an EDIT but it would be editing the ref to `0.1.3` not really leaving it commented out.

pameladelgado: oh and we might have to change `0.1.3` for `0.1.4`, apparently there are things to fix still 😬

lorenzo-cavazzi: I think this is getting a bit confusing. Let's start from a different perspective.

At the moment the template version in use is *tightly bound* to the ui version, it's generally not advisable to define a different version of the template somewhere else because it may be incompatible with the ui and lead to errors (as in the preview release). This is why in #702 I commented out the values for `ui.templatesRepository` -- and to be on the safe side, this was done also for custom values file SwissDataScienceCenter/sdsc-dev-docs#51

If you want to include a newer template version, the appropriate version of the UI should be used (re-tagged if needed).

> renku-ui-0.7.1 --> template 0.1.2
> renku-ui-0.7.2-0de296d --> template 0.1.3
> renku-ui-0.7.2-b51b22f  --> template 0.1.4

Since in the specific case of upgrading the templates from `0.1.2` and `0.1.3` nothing bad happens in the UI, it would be possible to use `ui.templatesRepository.ref` here as a shortcut, but doing the same for the version `0.1.4` would break the ui again.
To avoid further confusion, I strongly advise to *not* fix the template version with `ui.templatesRepository.ref` here and simply chose the appropriate UI version for the release. This will guarantee that everything works as expected and we will be sure there won't be any further confusion in the future release.

pameladelgado: yes, making a special renku-ui release is another option. I discussed with @ciyer about what to do for this before and we agreed that a new release was not needed and it was simpler to just modify the ref in the values. 
We do need to use the latest version of renku-python afaik so we probably need to prepare special renku-ui releases to include the templates we want. Keep in mind that you also need to make sure the right set of changes are included in each renku-ui version.
I think the UI team should agree on how to proceed in these cases, I dont mind either solution (modifying temporarily the values file or making a special ui release).

lorenzo-cavazzi: I believe @ciyer agreed on that for version `0.1.3` instead of `0.1.2` since it doesn't introduce any breaking change and the relevant PR in the ui accidentally slipped out from the release and was merged just after.

If you overwrite `ui.templatesRepository.ref` in any values file, please remember to remove/adjust it in the following releases. It's safe to assume that it will be kept up to date in the `renku-ui` repo until the template handling will change :slightly_smiling_face:

rokroskar:
The commit that says it changes the images on "First Steps" doesn't actually include new images - was this somehow overwritten in the force-push?

charts/values.yaml.changelog.md
pameladelgado: >I believe @ciyer agreed on that for version 0.1.3 

Cool.
>If you overwrite ui.templatesRepository.ref in any values file, please remember to remove/adjust it in the following releases.

This is exactly why I have the "Temporary value that can be deleted for next release." comment on that line

pameladelgado:
> The commit that says it changes the images on "First Steps" doesn't actually include new images - was this somehow overwritten in the force-push?

@rokroskar  yes, you are right. Thanks for spotting this! We will fix it shortly.

CHANGELOG.md
```diff
@@ -0,0 +1,39 @@
...
+
+* Changed project URLs to show namespace and name instead of project ID
+* Reworked collaboration view with issues list and collapsing issue pane 👥
+* Enabled search by username and group 🔍
+* Fork functionality now allows changing the name 🍴
+* Better tools to get information about interactive environments 🕹
+* Better consistency with project and interactive environment URLs 🎯
+
+### Miscellaneous
+* Commit time is local timezone aware 🕖
```
fgeorgatos: minor stylistic: add space above

CHANGELOG.md
```diff
@@ -0,0 +1,39 @@
...
+
+For changes to individual components, check:
+* renku ui [0.6.4](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.6.4), [0.7.0](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.7.0) and [0.7.1](https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/0.7.1)
+* renku-gateway [0.6.0](https://github.com/SwissDataScienceCenter/renku-gateway/releases/tag/0.6.0)
+* renku-python [0.7.1](https://github.com/SwissDataScienceCenter/renku-notebooks/releases/tag/0.7.1)
+* renku-graph [0.24.7](https://github.com/SwissDataScienceCenter/renku-graph/releases/tag/0.24.7)
+* renku-notebooks [0.6.1](https://github.com/SwissDataScienceCenter/renku-notebooks/releases/tag/0.6.1)
+
+## Bug fixes
+* Lineage visualization bugs addressed 🐞
```
fgeorgatos: ditto here and after L38

lorenzo-cavazzi:
Updated releases
* renku-notebooks: `0.6.2`
* renku-ui: `0.7.2`

PR #701, opened 2019-11-06
chore: Fix chart version after hotfix release

ableuler:


rokroskar:
The next release is being prepped already and it's `0.5.0` - why is this needed?

ableuler:
> The next release is being prepped already and it's `0.5.0` - why is this needed?

Happy to put it to `0.5.0` already. It's just that after the recent [hotfix](https://github.com/SwissDataScienceCenter/renku/commit/20db63ff5a59f72079cd529e5478150bb43e6720) the current head of master is very different from the released `0.4.3` chart. Some dev deployments are failing and it would be nice if the chart version gave an idea about what people have actually deployed.

PR #702, opened 2019-11-07
fix: comment templates values to inherit them from the ui chart

lorenzo-cavazzi:


charts/renku/values.yaml
```diff
@@ -494,9 +494,9 @@ ui:
   #   tag: latest
 
   ## Where to fetch default templates from
-  templatesRepository:
-    url: https://github.com/SwissDataScienceCenter/renku-project-template
-    ref: a7d01246fdbe247752a99b14c0b6e92666910b8c
+  # templatesRepository:
+  #   url: https://github.com/SwissDataScienceCenter/renku-project-template
+  #   ref: a7d01246fdbe247752a99b14c0b6e92666910b8c
```
pameladelgado:
```suggestion
  #   ref: 0.1.3
```

lorenzo-cavazzi: ok, I guess it makes sense to provide a tag here to be coherent with what we do in the ui chart even if the user can choose any valid reference

PR #703, opened 2019-11-07
chore: updated version of renku-graph

jachro:
This PR raises a version of renku-graph to 0.24.6.


pameladelgado:
Thanks for doing this, but this change is already part of #700 
see https://github.com/SwissDataScienceCenter/renku/pull/700/files#diff-da51da01c5466ce2b38c0105ebacfaebR16 , so if you are OK with this we can close this PR.
Also, for next time, you also need to run `helm dep update renku` to get an update on `requirements.lock` as well.

PR #714, opened 2019-12-03
Release fix: GitLab version

pameladelgado:
Upgraded GitLab version to enable renaming a forked project.


pameladelgado:
This chart 0.5.1-rc1 has been deployed on limited (modified to take the default values from renku) and it has the right version of GitLab.

rokroskar:
you should do a manual push of the charts so that you get a rendered chart with the appropriate gitlab version. Then change the requirement for gitlab in the renku `requirements.yaml` to that version.

PR #715, opened 2019-12-03
Updated the api-guidelines

jachro:
This PR updates API guidelines for the developers according to the agreement made on the Renku meeting.

### DIFF
--- a/docs/developer/contributing/api-guidelines.rst
+++ b/docs/developer/contributing/api-guidelines.rst
@@ -204,31 +204,38 @@ each API endpoint.
 
 A generic ``sort`` parameter **must** be used to describe sorting rules. If
 there are multiple fields that a collection can be sorted by, the ``sort``
-parameter **must** accept a list of comma separated fields, indicating
-descending sort priority. A minus sign implies descending sort order in the
-subsequent field.
+parameter **must** accept a list of comma separated fields. The format of the
+``sort`` parameter values has to follow: ``field:ASC|DESC``.
 
 - Retrieving all files, ordered alphabetically by file type and by descending
   file size:
-  ``GET /files?sort=type,-size``
+  ``GET /files?sort=type:ASC,size:DESC``
 
 6.3 Pagination
 ~~~~~~~~~~~~~~
 
 Pagination **must** be supported by all endpoints returning more than just
-a single resource. Pagination **must** implemented using the query
-parameters ``page`` and ``per_page``. Link headers **must** be included in the
-response indicating:
+a single resource. Pagination **must** be implemented using the query
+parameters ``page`` and ``per_page``. Both the parameters have to accept
+any positive integer value and default to 1 and 20 accordingly. Non-positive
+values for both parameters causes endpoint to return ``BAD_REQUEST (400)`` response.
+
+Link headers **must** be included in the response indicating:
 
 - first page ``rel="first"``;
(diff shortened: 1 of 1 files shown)


docs/developer/contributing/api-guidelines.rst
```diff
@@ -204,31 +204,38 @@ each API endpoint.
...
 6.3 Pagination
 ~~~~~~~~~~~~~~
 
 Pagination **must** be supported by all endpoints returning more than just
-a single resource. Pagination **must** implemented using the query
-parameters ``page`` and ``per_page``. Link headers **must** be included in the
-response indicating:
+a single resource. Pagination **must** be implemented using the query
+parameters ``page`` and ``per_page``. Both the parameters have to accept
+any positive integer value and default to 1 and 20 accordingly. Non-positive
```
ableuler: Should we leave the default page size open so that it can be set specifically for each case?

jachro: Yes, why not. Let me tweak that.

docs/developer/contributing/api-guidelines.rst
```diff
@@ -204,31 +204,40 @@ each API endpoint.
...
 Link headers are described by `RFC 8288
 <https://tools.ietf.org/html/rfc8288>`_.
 
-In cases where the total number of resources is not known in advance, the
-parameters ``limit`` and ``offset`` **should** be used instead. In this case,
-the link for the last page is omitted in the response header.
+Additional headers **must** be included in the response to present information
+about pagination:
+
+- ``Total``	for the total number of items;
```
ableuler: Oh, and another thing (sorry for doing this one-by-one). It is fine for me to always require the pagination headers, even when the link headers are present. But I think the case where the total number of result is not known is actually realistic, so it should be possible to omit this header. 
For example, a call to `https://gitlab.com/api/v4/projects?page=2&per_page=10` won't give you the total either number of pages either.

docs/developer/contributing/api-guidelines.rst
```diff
@@ -204,31 +204,40 @@ each API endpoint.
...
+- ``Total-Pages``	for the total number of pages;
```
ableuler: Oh, and another thing (sorry for doing this one-by-one). I think it is fine to always require the pagination headers, even if the pagination links are present in the header too. But the case where the total number of results is not known in advance seems realistic. For example gitlab won't give you a total number of pages when querying the projects endpoint either `https://gitlab.com/api/v4/projects?page=2&per_page=10`.

jachro: Not a problem you do it one by one.
I'm not sure if I got your thought. Do you mean the `Total` and `Total-Pages` to be returned always regardless of the link headers? If yes, that was my intention too. Effectively, you should be getting all the link headers and the, let's say additional headers with every response to the paged endpoint.

ableuler: > I'm not sure if I got your thought. Do you mean the Total and Total-Pages to be returned always regardless of the link headers?

I agree with that in general. But I would simply add the for cases where the total number of resources is not known, the `Total-Pages` header and also the link to the last page can be omitted.

jachro: Oh, I see what you mean. Of course, I'll change that.

PR #722, opened 2020-01-13
WIP: update renku-ui version to 0.7.4

ciyer:


charts/renku/requirements.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
 - name: renku-ui
   alias: ui
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: "0.7.2"
+  version: "0.7.4"
```
lorenzo-cavazzi: I think it should be `0.7.3` here

lorenzo-cavazzi: Ops wait, I just saw #721 
I guess that is the proper version and this PR should be here yet, right?

ciyer: Yes, this should come after the update to 0.7.3, but I guess it should not be merged until the release is made.

lorenzo-cavazzi: I am a bit confused :thinking: 
The current renku release would contain renku-ui `0.7.3`, right? Then it makes sense to immediately update the ui version to `0.7.4` on the renku-ui chart, but not here where it should stay like this until we release renku-ui `0.7.4` at some point in the future. Does this make sense?

rokroskar:
Closing this as it will not be necessary anymore with automatic chart version updates.

PR #723, opened 2020-01-14
docs: update templates for project level environment options

lorenzo-cavazzi:
It's now possible to set project-level notebook options through the .renku/renku.ini file. The
.renku/metadata.yml file is not part of the templates anymore since those data are set by the `renku
init` command.

It makes sense to approve also this at the same time SwissDataScienceCenter/renku-project-template#49

### DIFF
--- a/docs/user/templates.rst
+++ b/docs/user/templates.rst
@@ -129,11 +129,12 @@ Here is the list of properties that can be customized in a standard Renkulab
 deployment:
 
 * ``default_url [string]``: URL to use when starting a new interactive
-  environment
-* ``cpu_request [float]``: CPUs quota
-* ``mem_request [float]``: memory quota (in GBs)
-* ``gpu_request [int]``: GPU quota
+  environment (``/lab``, ``/tree``, ...)
+* ``cpu_request [float]``: CPUs quota (``0.5``, ``1``, ...)
+* ``mem_request [string]``: memory quota (``1G``, ``2G``, ...)
+* ``gpu_request [int]``: GPU quota (``0``, ``1``, ...)
 * ``lfs_auto_fetch [bool]``: whether to automatically fetch lfs files or not
+  (``true``, ``false``)
 
 .. note::
 
--- a/docs/user/templates.rst
+++ b/docs/user/templates.rst
@@ -145,7 +145,7 @@ deployment:
 
       > renku config --local interactive.default_url "/rstudio"
 
-    Verify that your renku.ini file looks like the following.
+    Verify that your ``renku.ini`` file looks like the following.
 
     .. code-block:: console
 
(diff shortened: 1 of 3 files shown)


docs/user/templates.rst
```diff
@@ -71,44 +74,84 @@ here any files that you don't want to be tracked.
...
+allows you to specify which files to ignore in a docker build. If you aren't
+making changes to the docker build, you can .dockerignore this file.
 
 ``README.md``
 """""""""""""
 
-The ``README.md`` file is shown on a project's home page. It's good to have at least
-the name of the project and a brief overview of the project for your intended
-audience.
+The ``README.md`` file is shown on a project's home page. It's good to have at
```
ciyer: Proposed change to this section:

```
``.renku``
""""""""""

The ``.renku`` directory includes a ``renku.ini`` file which contains 
project-level configuration for renku, stored using the 
`INI format <https://en.wikipedia.org/wiki/INI_file>`_. Currently, it 
can be used to specify defaults values for launching interactive environments.

**Interactive Environments**

If your project has specific resources requirements to run, or if it should default
to RStudio or anything other than JupyterLab, then you will want to provide a
configuration for the interactive environments.

Although the file may be modified manually, it is recommended to use the
``renku config --local interactive.<property> <value>`` command.

Here is the list of properties that can be customized in a standard Renkulab
deployment:

* ``default_url [string]``: url to use when starting a new interactive
  environment
* ``cpu_request [float]``: CPUs quota
* ``mem_request [float]``: memory quota (in GBs)
* ``gpu_request [int]``: GPU quota
* ``lfs_auto_fetch [bool]``: whether to automatically fetch lfs files or not

.. note::

    We use JupyterLab as the default web interface for interactive environments. If you work in R, you may prefer to
    have RStudio ad the default. This can be achieved by using `/rstudio` as the ``default_url`` instead of `/lab`.

    .. code-block:: console

      > renku config --local interactive.default_url "/rstudio"

    Verify that your renku.ini file looks like the following.

    .. code-block:: console

      [renku "interactive"]
      default_url = /rstudio

    If you ran this command locally, you will need push back to the renkulab server, e.g., 
    
    .. code-block:: console

      > git push
    
    before this change is available. (`renku config` automatically creates a commit).

    You can now start a new environment against the latest commit and you will have RStudio as the
    default web interface.
    

.. note::

    Using the same approach as above for RStudio, it is possible to switch the interface from JupyterLab to
    the classic Jupyter Notebook by using `/tree` as the ``default_url`` instead of `/lab`.

    .. code-block:: console

      > renku config --local interactive.default_url "/tree"
```

lorenzo-cavazzi:
Thanks for the rewording, it's more clear now. I made some mess with the commits, but the suggestions are applied now.

PR #732, opened 2020-01-28
Add acceptance test for project forking

ciyer:


tests/src/main/scala/ch/renku/acceptancetests/ReplShell.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import org.openqa.selenium.{By, WebDriver}
+import org.openqa.selenium.chrome.ChromeOptions
+import org.openqa.selenium.remote.RemoteWebDriver
+import org.scalatestplus.selenium.{Chrome, Driver, WebBrowser}
+
+/**
+  * A shell for interactively controlling the browser.
+  */
+object ReplShell {
```
jachro: Is this object just to try something out? If yes, maybe it doesn't have to go into the PR?

tests/src/test/scala/ch/renku/acceptancetests/ExistingProjectSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+    with RemoveProject
+    with Settings
+    with Fork {
+
+  scenario("User can fork a project") {
+    implicit val loginType: LoginType = logIntoRenku
+
+    existingProjectDetails match {
+      case Some(projectDetails) => forkTestCase(projectDetails, loginType)
+      case None                 =>
```
jachro: Is that intentional to skip silently the test if there are no project details available? Actually, in the results of the test, we'll see a green pass for it while effectively nothing was tested.

tests/src/test/scala/ch/renku/acceptancetests/HandsOnSpec.scala
```diff
@@ -29,6 +29,8 @@ class HandsOnSpec
 
     implicit val projectDetails: ProjectDetails = ProjectDetails.generateHandsOnProject(captureScreenshots)
 
+    implicit val docsScreenshots = new DocsScreenshots(this, captureScreenshots, browser)
```
jachro: I should maybe ask about that in some previous PR but only noticed that now. It's about the `captureScreenshots` val. What do you think if the value is read from the args/env vars as part of the `DocsScreenshots` instantiation? I'm wondering if that wouldn't be maybe more suitable place for it instead of the `HandsOnSpecData`. I guess if you would like to generate screenshots for other flows the `DocsScreenshots` could be a bit simpler to use.

tests/src/main/scala/ch/renku/acceptancetests/ReplShell.scala
ciyer: I improved the code and added some documentation to explain how this can be used. Please take a look again.

tests/src/test/scala/ch/renku/acceptancetests/ExistingProjectSpec.scala
ciyer: For development or quick verification, it is nice to be able to run a test against an existing project, without the overhead of creating a new project.

I have added some output to explain what happens when no project was provided to run against.

tests/src/test/scala/ch/renku/acceptancetests/HandsOnSpec.scala
ciyer: The DocsScreenshots class was introduced in this PR, so your suggestion would not have been possible in the last PR.

This is a good suggestion! I have implemented it.

PR #733, opened 2020-01-28
[WIP] Pulumi install

### LINKED ISSUE #671: helm alternatives
Helm is intended to be a package manager for K8s. But its use as a devops/deployment tool is quite lacking. A year ago there were not many alternatives, but a lot of projects have been created in the meantime, either building on top of helm or replacing it.

In light of this, we should reconsider our tooling and see if there's any alternative better suited to our use cases.

Shortcomings of helm (A bit nit-picky and not everything here has to be addressed by an alternative :wink:  ):
- YAML file aren't DRY and there's not a lot of support for reusing definitions, which leads to a lot of duplication and difficulties in maintaining the code. Plus it's very verbose and hard to read, especially for newcomers.
- The `values.yaml` cannot be templated. It is impossible to e.g. use `.Release.Name` in the `values.yaml`, making it difficult for resources (or resources in subcharts) to refer to other resources which have dynamic names. For instance, a helm secret with a name of `{{ template "renku.fullname" . }}-secret` can't be passed to the `existingSecret` value of the keycloak subchart.
- The previous point means that we are entirely dependent on subchart  developers to add support to their chart for customizing (Passing values through the `tpl` function or allowing `extraEnv` to be set). This means we can only change what is intended to be changed by the chart developer, i.e. what they thought of when creating the charts, making it difficult to customize things when they did not envision our use cases (other than copying their chart and maintaining it ourselves).
- You get a single `values.yaml` for all your needs (There are some options to use more than one file with cmd flags, but the implementation is clunky and hard to use for users), which in many cases is too restrictive for actual devops needs.
- The way subcharts are handled means that a single chart without dependencies might work well enough, but having subcharts, subsubcharts etc., it quickly becomes unwieldy and a dependency hell, it's not obvious what need to be set on a subchart, some breaking changes to a subchart might go unnoticed (e.g. a subchart renames a value entry, we don't rename it in our `values.yaml`, so the default value gets used, things break but it's not obvious why they break)
- Deployments generally don't depend on the current state of the application. There's no way to deploy depending on what is already deployed. In cases where a change to one resource necessitates a reload of another (but unchanged) resource, you need hacks to force the restart, making upgrades difficult to do  and leaving a lot of room for bugs to creep into existing deployments on upgrade.
- More generally, helm doesn't really support lifecycle management of complex deployments and different environments. Workarounds/hacks and manual workflows are needed for something that should ideally be handled by tooling instead. 
- Bugs/typos in charts are very hard to spot, with sometimes borderline unreadable error messages.
- Bad user interaction. You run the command, wait a bit and maybe something didn't work. There's no way to provide feedback to the user like printing warnings, prompting for missing values, offering the user choices, no native input validation (I can set `serverOptions.cpu_request.default` to -100 and helm will happily apply it, writing validations for this in helm charts gets quite clunky, especially dynamic validations). 
- Go tpl/sprig is quite limited in scope and writing more than simple `| b64enc | quote` style code gets unmaintainable quickly. It lacks all the basic tools that usual programming has (Debugging, tests, etc.) and its syntax is not that usable for complex scripting. In a similar vein, the scripting has no regards for the structure and semantics of yaml itself. It should not be up to the developer to make sure that the produced yaml is syntactically valid!. Basically it doesn't really separate representation and code. If resources were objects in a programming language instead, you could work with them in a safe way, a compiler would catch illegal assignments, and indentation would be a non-issue as it would be automatically done on rendering.

Things helm does well:
- Preexisting charts for many applications, making it easy to get basic deployments.
- Rancher (and other platforms) integrations that allow one-click deployments - which they tend to work OK overall, but sometimes fail specifically due to helm/yaml quirks.
- No programming experience needed (at least in the ideal case) 
- Works well for charts without variability, applications that don't need a lot of customization and have a single, one size fits all, configuration.
- It is popular, which is a good thing in any tool
- It's a very thin layer on top of k8s, making it easy to set up and get started with.
- Open-source and can be used by anyone* (*terms and cluster-limitations may apply)

Ideally, an alternative to helm would still allow us to use helm charts, allowing for gradual replacement, but at the same time give us more powerful tools to deal with our complex use cases and increased ability to customize existing charts. And an alternative should also be usable by anyone (for free) without too much hassle (Ideally, install a tool, clone a repo, run a command, done).

A non-exhaustive list of possible alternatives:

- terraform (k8s or helm provider?)
- pulumi (k8s or helm provider?)
- helmfile (+helm)
- kustomize (together with helm?)
- Operator Framework (k8s or helm operators?)
- Ship
- JSonnet
- Kapitan
- hm, helm 3 (as a minimum, it solves the sanity checking via JSON schema based validation, #655)

Feel free to look for others.

We should evaluate which, if any, of these tools fit our workflow and use cases and then make a decision on whether to keep using helm or switching to one of these.

### LINKED ISSUE #1079: Get pulumi deployment running without gitlab
A first pulumi-powered Renku deployment should use pulumi 2.0 and rely on an external GitLab instance. [These values](https://github.com/SwissDataScienceCenter/renku/blob/minikube-cleanup/charts/example-configurations/external-gitlab-values.yaml) contain the necessary additional configuration for connecting to an external gitlab instance.

Panaetius:
Ports over the `renku` and `renku-gateway` charts to pulumi deploys. Needs https://github.com/SwissDataScienceCenter/renku-graph/pull/242 to work.

Currently needs a pulumi stack .yaml with all chart values to work, ask me for an example file on slack if needed.

~~Closes #671~~ 
Closes #1079


Panaetius:
stack yaml might still be optimized a bit, but it's pretty clean already (80 lines vs. 244 lines previously).

There's a `renku:dev` boolean to enable a dev deploy, and a `renku:baseurl` string that will automatically generate relevant Url's for a deployment (instead of having to specify them).

What's still needed is adding functionality to migrate existing deployments.

renku_pulumi/Readme.md
```diff
@@ -0,0 +1,89 @@
...
+  renku:keycloak_enabled: "true"
+  renku:minio_enabled: "false"
+  renku:postgres_enabled: "true"
+  renku:dev: "true"
+  renku:baseurl: dev.renku.ch
+  kubernetes:namespace:
+```
+
+- Fill in the kubernetes namespace in the `kubernetes:namespace:` and `graph:values: sentry: environmentName:` fields, as well as adding the gitlab application
+  token id and secret in `global:values: gitlab: clientAppId:` and `global:values: gitlab: clientAppSecret:`.
```
rokroskar: Maybe these would be more readable like this:

```
global:values: 
  gitlab: 
    clientAppId:
```

renku_pulumi/Readme.md
```diff
@@ -0,0 +1,89 @@
...
+  with the `global:  gitlab:    clientSecret:` value from your previous deployment yaml
+
+- set the kubectl context to one approriate for the namespace you wish to deploy to (`kubectl config set-context <context>`)
```
rokroskar: maybe a less invasive strategy is to use `pulumi config set kubernetes:context`

renku_pulumi/chart_requirements.ini
```diff
@@ -0,0 +1,31 @@
+[graph]
```
rokroskar: Can we make this yaml or json for easier manipulation in scripts?

renku_pulumi/requirements.txt
```diff
@@ -0,0 +1,5 @@
+pulumi>=1.0.0
```
rokroskar: This should go in the `Pipfile` one level up

renku_pulumi/resources/main/ingress.py
```diff
@@ -0,0 +1,141 @@
...
+                                "serviceName": global_config["gateway"]["name"],
+                                "servicePort": global_config["gateway"]["service"][
+                                    "port"
+                                ],
+                            },
+                        },
+                        {
+                            "path": "/",
+                            "backend": {
+                                "serviceName": "{}-ui-renku-ui".format(stack),
```
rokroskar:
```suggestion
                                "serviceName": "{}-ui".format(stack),
```

renku_pulumi/resources/main/ingress.py
```diff
@@ -0,0 +1,141 @@
...
+                    ]
+                },
+            }
+
+            if config.get_bool("keycloak_enabled"):
+                h["http"]["paths"].append(
+                    {
+                        "path": "/auth",
+                        "backend": {
+                            "serviceName": "{}-keycloak-http".format(stack),
```
rokroskar: The service and pod name will not always have `keycloak` - sometimes it can be shortened/abbreviated.

renku_pulumi/__main__.py
```diff
@@ -0,0 +1,159 @@
...
+    nb = renku_notebooks(
+        config, global_config, chart_reqs, js, p, dependencies=[pg_job]
+    )
+
+    # gateway
+    redis_chart = redis(global_config, chart_reqs)
+    sc = gateway.secret(global_config, gateway_values)
+    cfg = gateway.configmaps(global_config, gateway_values)
+    gateway.ingress(global_config, gateway_values)
+    gd = gateway.gateway_deployment(global_config, gateway_values, sc, cfg)
```
rokroskar: can we make the naming here more verbose and less cryptic?  As someone coming to understand how the platform is deployed and what pieces there are, this is probably the most critical starting point and we should make it as clear as possible.

aledegano:
hey @Panaetius is this PR still relevant? Seems abandoned for a while...

Panaetius:
> hey @Panaetius is this PR still relevant? Seems abandoned for a while...

I have no clue, it's a nice programmatic way to bring up the platform and model all the dependencies, but then no one other than me really seemed interested in it.

It'd probably need a bit of work to catch up to changes happened since, and maybe Pulumi fixed some issues on their end since that'd allow us to get rid of workarounds.

But I'm not really the person to ask if this is still something we want, as I'm not really the target audience.

But I'm sure it can be instructive either way, as it models all the dependencies we have more clearly than if you just looked at the helm chart.

PR #734, opened 2020-01-29
Updates post 0.5.2 release

pameladelgado:
Travis allows tags labels that use `-rc` followed by one or more digits.
Updated changelog.


pameladelgado:
> Do we need to bump the chart version?

yes that too, but I think it should not be 0.5.2 since there are new versions of other components on master, which value do you propose we use? we should probably have a policy for setting up the chart versions...

rokroskar:
I would call it 0.6.0 for now because unless there are bug fixes the next version will probably include the new renku-core service.

pameladelgado:
That's the actual value now, I guess we will leave it as is in master.

rokroskar:
Ah indeed. Oops, sorry!

PR #737, opened 2020-02-04
ci/cd: restucture test and deploy pipeline

### LINKED ISSUE #691: build: move CI deployments to the dev cluster
Currently PRs in this repo are deployed to a VM running minikube using helm. Instead, we should deploy them to the dev cluster so they are easier to inspect in case of test failures. 

The idea would be to:

1. create a k8s config for the bot user on the dev cluster 
2. adapt the current deploy scripts to use this k8s config
3. create a deployment against the gitlab running on dev, minimizing the footprint of each deploy
4. run tests etc on every push
5. clean up after the PR is merged

The last one is tricky with travis afaik so this might be a good opportunity to test-drive github actions for this use-case.

rokroskar:
Deploys to the dev cluster via a github action. 

To-do:

- [x] deploy PRs (see e.g. [this PR's deployment](https://ci-03b9279-renku.dev.renku.ch))
- [x] teardown closed PRs
- [x] run acceptance tests on PRs
- [x] automatically create prs from `auto-update/` branches
- [x] deploy on changes to `charts/**` ([successful deploy](https://github.com/SwissDataScienceCenter/renku/runs/427908321?check_suite_focus=true))
- [x] publish chart and images ([tested here](https://github.com/SwissDataScienceCenter/renku/runs/431951472?check_suite_focus=true))
- [x] remove travis

closes #691


scripts/travis-install.sh
```diff
@@ -41,7 +41,7 @@ rm helm.tar.gz
 rm -r linux-amd64
 
 
-# install pipenv
-pip install pipenv
-# install required packages
-pipenv install --dev --system --deploy
+# # install pipenv
```
jachro: Maybe we can simply delete the whole file?

rokroskar: ah yes, I need to clean this up still

rokroskar:
Thanks @jachro! I need to test the change from @ciyer that should allow the tests to pass and clean up a bit then it's ready to go IMHO

rokroskar:
I think until https://github.com/SwissDataScienceCenter/renku/pull/746 is merged we can simply not require the `test-pr` job to pass before merging. This is essentially the same as the current behavior where we were only testing that the chart deployed (but not actually running the integration tests). Once #746 is merged we can require the `test-pr` job for merging as well.

rokroskar:
Ok so for the record, I will turn off the travis-ci required check and activate the "deploy_pr" status check

PR #746, opened 2020-02-07
Test improvements

ciyer:
* Support the register a new user workflow (this is done with the register flag, see docs)

### DIFF
--- a/tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
+++ b/tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
@@ -23,7 +23,7 @@ case object LoginPage extends RenkuPage {
   /**
     * When a renku instance uses another openID connect provider, the user may get a second login page.
     */
-  private case class ProviderLoginPage(userCredentials: UserCredentials)
+  case class ProviderLoginPage()
       extends selenium.Page
       with ScalatestMatchers
       with Eventually
@@ -31,7 +31,7 @@ case object LoginPage extends RenkuPage {
 
     final override lazy val url: String = "https://dev.renku.ch/auth/realms/Renku/protocol/openid-connect/auth"
 
-    def logIn(implicit webDriver: WebDriver): Unit = eventually {
+    def logInWith(userCredentials: UserCredentials)(implicit webDriver: WebDriver): Unit = eventually {
       usernameField.clear() sleep (1 second)
       usernameField.sendKeys(userCredentials.email.value) sleep (1 second)
 
@@ -54,7 +54,7 @@ case object LoginPage extends RenkuPage {
     }
   }
 
-  private case class UpdateAccountInfoPage(userCredentials: UserCredentials)
+  case class UpdateAccountInfoPage(userCredentials: UserCredentials)
       extends selenium.Page
       with ScalatestMatchers
       with Eventually
@@ -69,7 +69,7 @@ case object LoginPage extends RenkuPage {
 
       val nameComps = userCredentials.fullName.split(" ")
       val firstName = nameComps.head
-      val lastName = nameComps.last
+      val lastName  = nameComps.last
 
       firstNameField.clear() sleep (1 second)
       firstNameField.sendKeys(firstName) sleep (1 second)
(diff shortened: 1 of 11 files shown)


rokroskar:
Do the tests ever run out of order? Also what happens if I run the register flow twice with the same user?

rokroskar:
I think we need to support the case where the test is run with the same options multiple times. I suppose this means that we must somehow check if the user is already registered - one option is to simply fail gracefully and anticipate this screen:

20200207233232-LoginSpec

If we get here, we simply go back and log in.

rokroskar:
The tests work when it's the first time the user is registering: https://github.com/SwissDataScienceCenter/renku/runs/432969756?check_suite_focus=true

tests/src/test/scala/ch/renku/acceptancetests/workflows/Login.scala
```diff
@@ -4,23 +4,135 @@ import ch.renku.acceptancetests.pages._
...
-    Given("User is logged in on the Welcome page")
+    Given("user is not logged in")
     go to LandingPage sleep (1 second)
     verify browserAt LandingPage
-    click on LandingPage.loginButton
+
+    When("user clicks on the Login button")
+    try {
+      click on LandingPage.loginButton
+    } catch {
```
jachro: As it looks the `StaleElementReferenceException` occurs on the click operation, could we maybe build our own grammar for clicking a button which would do the re-try on that exception internally? I know, there's is this `click` object on the `WebBrowser` but it could be quite useful to have that mechanism in place for all the clicks everywhere. I'm not sure now how difficult it will be to do that but maybe it's worth trying out...

jachro: The other thing is that maybe it would be good to add some tiny delay before we try to click again. Otherwise, it just shoots like a machine gun :)

ciyer: 1. I think that is a good suggestion, but I would wait until we identify more places where this behavior is happening. Right now, I have only seen in on the login, so I'm ok with have a custom solution in this one location until we need a general solution.

2. Good point. Done.

tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
```diff
@@ -97,69 +97,96 @@ case object LoginPage extends RenkuPage {
...
+    usernameField.sendKeys(userCredentials.email.value) sleep (1 second)
+
+    passwordField.clear() sleep (1 second)
+    passwordField.sendKeys(userCredentials.password.value) sleep (1 second)
+
+    try {
+      logInButton.click()
+    } catch {
+      // Just try again
+      case e: StaleElementReferenceException => logInButton.click()
```
jachro: Maybe the `sleep` here, too?

ciyer: I think this catch is not necessary at all. I've removed it to test that hypothesis. If it is necessary, I will add it back and include a sleep.

jachro: 👍

tests/src/test/scala/ch/renku/acceptancetests/workflows/Login.scala
jachro: You know it better but isn't the `StaleElementReferenceException` being thrown here because we don't have a sleep after the previous step which is `verify browserAt LandingPage`?  I'm just wondering if what could be happening here is that the browser has already the `LandingPage` URL (so the `verify browserAt LandingPage` passes ) but the content is still being loaded while we try to click the LoginButton.

ciyer: Yes, I think there is a possibility that sleeping longer will fix this problem as well...

jachro: So which option do you reckon would be more robust? From the code simplicity extending the timout seems to be better but to be honest I'm not so sure...

ciyer: I'm increased the timeout. This seem sufficient.

rokroskar:
If @jachro is happy with this PR we can merge it, it will self-correct as soon as the renku-python version in the triples gen is updated.

PR #767, opened 2020-02-14
chore: add ui sentry info

lorenzo-cavazzi:
This PR adds basic info to `charts/renku/values.yaml` about the new `ui.sentry` settings. Related to SwissDataScienceCenter/renku-ui#761

### DIFF
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -528,6 +528,12 @@ ui:
   # templatesRepository:
     # url: https://github.com/SwissDataScienceCenter/renku-project-template
     # ref: 0.1.5
+  ## Configuration for Sentry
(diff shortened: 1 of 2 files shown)


charts/renku/values.yaml
```diff
@@ -528,6 +528,12 @@ ui:
   # templatesRepository:
     # url: https://github.com/SwissDataScienceCenter/renku-project-template
     # ref: 0.1.5
+  ## Configuration for Sentry
+  # sentry:
+  #   enabled: true
```
rokroskar: is it on by default? The commented values here should show the defaults.

lorenzo-cavazzi: oh I see, let me change it then

PR #773, opened 2020-02-14
docs: update first step tutorial

lorenzo-cavazzi:
Updating the first step tutorial


docs/user/firststeps.rst
```diff
@@ -520,30 +530,42 @@ and other commands.
...
+    Remove the untracked files and this time execute ``only`` the renku command
+
+    .. code-block:: console
+
+        rm data/output/*
+        renku run python src/00-FilterFlights.py data/201901_us_flights_1/2019-01-flights.csv.zip data/output/2019-01-flights-filtered.csv
+
+.. note::
+
+    Did you get an error like this instead?
```
rokroskar: Note - this should not happen with `renku>=0.9.0` - if it does, we need to reopen the appropriate issue

lorenzo-cavazzi: I was using renku 0.9.0 and this happened again :thinking: 

https://dev.renku.ch/projects/lorenzo.cavazzi.tech/flights-tutorial/

lorenzo-cavazzi: Oh wait, you are right, I tried to reproduce this and it's not happening anymore. I am wondering how it comes that this happened the first time I run through this

rokroskar: ah wait, actually this change doesn't make sense - the error should happen when the data is not pulled yet but in that case you won't get a "file not found" error, you'll get the "bad zip archive" error instead. No?

lorenzo-cavazzi: that is the error below, this one happens when accidentally executing also the first part of the code that is the "non-renku" way

rokroskar: oh I probably read it out of order 🤦‍♂

PR #786, opened 2020-02-19
build: add component update action

### LINKED ISSUE #784: create re-usable action for updating main chart
Ideally we could reuse the code to update the main chart across all of our projects.

rokroskar:
Adds an actions directory with generic re-usable actions. 

Two actions are added: 

* action to update a component version in the main renku chart
* action to build images and publish the chart with chartpress

closes #784

### DIFF
--- a/actions/update-component-version/update-upstream.sh
+++ b/actions/update-component-version/update-upstream.sh
@@ -24,6 +24,16 @@ then
   CHART_TAG="--tag $(echo ${GITHUB_REF} | cut -d/ -f3)"
 fi
 
+if [ -z $GIT_EMAIL ]
+then
+  echo "GIT_EMAIL must be set."
+fi
+
+if [ -z $GIT_USER ]
+then
+  echo "GIT_USER must be set."
+fi
+
 # install helm
 HELM_URL=${HELM_URL:=https://storage.googleapis.com/kubernetes-helm}
 HELM_TGZ=${HELM_TGZ:=helm-v2.14.3-linux-amd64.tar.gz}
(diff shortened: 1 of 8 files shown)


actions/update-component-version/update-upstream.sh
```diff
@@ -0,0 +1,73 @@
...
+# update the upstream repo
+cd upstream-repo
+
+# set up git
+git config --global user.email "$GIT_EMAIL"
+git config --global user.name "$GIT_USER"
+
+# update the chart requirements and push
+git checkout -b auto-update/${CHART_NAME}-${CHART_VERSION} master
+yq w -i charts/renku/requirements.yaml "dependencies.(name==${CHART_NAME}).version" $CHART_VERSION
```
jachro: I'm wondering if requiring UPSTREAM_REPO variable is something we should do assuming that we have to hardcode paths like charts/renku/requirements.yaml anyway?

rokroskar: oh yeah good point - I could get rid of that

rokroskar: or, alternatively, make the path to the requirements file configurable

jachro: Yes, you can try and then we'll see how complex that would become. Do you aim to end up with something others can use? To me, that's something very renku specific but of course I could be wrong.

rokroskar: yes, this one is probably pretty renku-specific - the other one (chartpress) is pretty generic

rokroskar: ok I simplified it a bit

rokroskar:
I think there's still a problem unfortunately - looks like 

```
if [[ $GITHUB_REF =~ "tags" ]]
```

is not valid shell syntax (needs bash)...

jachro:
Yes, I have `#!/bin/env bash` in my travis scripts.

rokroskar:
Right, I just need to install bash in the docker image.

jachro:
Does it work better with bash?

rokroskar:
yeah no more errors: https://github.com/SwissDataScienceCenter/renku-python/runs/460751439?check_suite_focus=true#step:6:10

PR #792, opened 2020-02-20
feat: add ingress rule for /entities

ableuler:
This PR adds an ingress rule which forwards traffic from `/entities` to the gateway.

Currently deployed together with https://github.com/swissdatasciencecenter/renku-gateway/issues/174 under https://andreas.dev.renku.ch

### DIFF
--- a/charts/renku/templates/ingress.yaml
+++ b/charts/renku/templates/ingress.yaml
@@ -66,6 +66,10 @@ spec:
         backend:
           serviceName: {{ $gatewayFullname }}
           servicePort: {{ $gatewayServicePort }}
+      - path: /entities
+        backend:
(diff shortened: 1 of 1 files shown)


rokroskar:
https://github.com/swissdatasciencecenter/renku-gateway/issues/174 is not really blocking this PR - it can be merged, th `/entities` endpoint will just not go anywhere.

ableuler:
> [SwissDataScienceCenter/renku-gateway#174](https://github.com/SwissDataScienceCenter/renku-gateway/issues/174) is not really blocking this PR - it can be merged, th `/entities` endpoint will just not go anywhere.

True that, not really a dependency but rather a related PR.

PR #827, opened 2020-02-27
Test robustness

ciyer:
Improvements to tests to make them more robust.


tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
```diff
@@ -201,6 +203,26 @@ class ProjectPage(projectDetails: ProjectDetails, userCredentials: UserCredentia
           .getOrElse(fail("Environments -> Running title not found"))
       }
 
+      def connectToJupyterLab(implicit webDriver: WebDriver, spec: AcceptanceSpec): Unit = eventually {
+        spec.And("tries to connect to JupyterLab")
```
jachro: What about inserting an `import spec._` as a first line of the method body. Then, we could take out all the `spec.` prefixes?

rokroskar:
Now lets see if the tests pass :)

PR #916, opened 2020-03-25
chore: comment out template values

lorenzo-cavazzi:
Comment out template repository version to allow inheriting the value from the UI values


lorenzo-cavazzi:
Integration tests cannot pass until the Docker image is available

PR #968, opened 2020-04-07
Modify Renku chart to enable anonymous notebook sessions (optional)

ableuler:
This PR adds the chart changes necessary to enable anonymous interactive environments. Only merge after

- [x] https://github.com/SwissDataScienceCenter/renku-ui/pull/893 

- [x] https://github.com/SwissDataScienceCenter/renku-notebooks/pull/285 

- [x] https://github.com/SwissDataScienceCenter/renku-gateway/pull/194

have been merged.

Deployments with all these changes are running under:
https://tmp-on.dev.renku.ch/ (anonymous sessions enabled)
https://tmp-off.dev.renku.ch/ (anonymous sessions disabled)

### DIFF
--- a/charts/renku/requirements.lock
+++ b/charts/renku/requirements.lock
@@ -4,10 +4,10 @@ dependencies:
   version: 0.10.3-a68dcb7
 - name: renku-ui
   repository: https://swissdatasciencecenter.github.io/helm-charts/
-  version: 0.9.2-8a4685a
+  version: 0.9.1-n008.ha2f5325
 - name: renku-notebooks
   repository: https://swissdatasciencecenter.github.io/helm-charts/
-  version: 0.7.1-0aec590
+  version: 0.7.2-e00ce47
 - name: renku-gateway
   repository: https://swissdatasciencecenter.github.io/helm-charts/
   version: 0.7.1-deb488b
@@ -26,5 +26,5 @@ dependencies:
 - name: minio
   repository: https://kubernetes-charts.storage.googleapis.com
   version: 1.6.0
-digest: sha256:b3d7d18fda8118a73ec27c805f110da5a4940d0c2d144562fecfb6fc43252f56
-generated: 2020-04-14T19:04:30.178207317Z
+digest: sha256:34c16bf7a89fb1217cf5af0708e2ca3b054e8ce5741ed5275c90370a2fcae2aa
+generated: "2020-04-14T18:25:32.231125+02:00"
--- a/charts/renku/templates/NOTES.txt
+++ b/charts/renku/templates/NOTES.txt
@@ -8,3 +8,28 @@ You have configured a demo user for the platform. The details (password) of this
 can be accessed using the following one-liner (you need to have jq installed).
 kubectl get secrets -n <k8s-namespace> {{ template "renku.fullname" . }} -o json | jq -r .data.users | base64 --decode
 {{- end -}}
+
+{{- if .Values.global.anonymousSessions.enabled -}}
+You have enabled anonymous interactive sessions. For this feature to work you
+will have to perform a few manual steps. These are the ready-made or almost ready
+commands that you will need. Check out the docs for more information:
+https://renku.readthedocs.io/en/latest/admin/index.html#enabling-notebooks-for-anonymous-users
+
+{{ if .Values.global.anonymousSessions.autoCreateTmpNamespace -}}
+A secondary namespace {{ .Release.Namespace }}-tmp has been created.
+{{ else }}
+Please create a secondary namespace using the following command:
+$ kubectl create namespace {{ .Release.Namespace }}-tmp
+{{- end }}
+
+You have to copy a secret into this secondary namespace using this command:
+$ kubectl get secret renku-jupyterhub-tmp-postgres -n {{ .Release.Namespace }} -o yaml | grep -v '^\s*namespace:\s' | kubectl apply -n {{ .Release.Namespace }}-tmp -f -
+
+The following two commands have to be executed from renku-notebooks/helm-chart.
+
+The command to create a values file for your secondary notebooks-service deployment:
+$ pipenv run ./make-values-tmp.py  --values-file <path-to-renku-values> --renku-namespace {{ .Release.Namespace }}
+
+Finally, the command for deploying the secondary notebooks service:
+$ helm upgrade --install {{ .Release.Namespace }}-tmp-notebooks --namespace {{ .Release.Namespace }}-tmp -f ./notebooks-tmp-values.yaml --timeout 1800  ./renku-notebooks
+{{ end }}
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -83,6 +83,20 @@ global:
       dataset:
   ## Set to true if using https
   useHTTPS: false
+  anonymousSessions:
+    ## Set to true to enable anonymous sessions through
+    ## a secondary Jupyterhub instance
+    enabled: false
+    ## Set the postgres password for you secondary JH deployment
+    ## or force the (re)generation of it on helm upgrade.
+    postgresPassword:
+      value:
+      overwriteOnHelmUpgrade: false
+    ## Set to false to avoid auotmatically creating the secondary
+    ## k8s namespace for the tmp notebooks-service deployment. This
+    ## can be necessary to avoid permission problems in renku.dev.ch
+    ## deployments.
+    autoCreateTmpNamespace: true
 
 ## Ingress configuration
 ## See: https://kubernetes.io/docs/concepts/services-networking/ingress/
@@ -575,13 +589,14 @@ notebooks:
     # rbac:
     #   enabled: true
     ## Set the notebooks service API token
-    # hub:
-    #   services:
+    hub:
+      services:
         # notebooks:
         #   api_token: notebookstoken
 
         ## define the client ID and secret for the gateway
-        # gateway:
+        gateway:
+          oauth_no_confirm: true
         #  oauth_client_id: gateway
         #  apiToken: # use `openssl rand -hex 32`
         #  oauth_redirect_uri: http://<gateway>/api/auth/jupyterhub/token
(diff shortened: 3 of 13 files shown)


lorenzo-cavazzi:
should we add `anonymousSessions` also [here](https://github.com/SwissDataScienceCenter/sdsc-dev-docs/blob/master/dev-deployments/dev.values.yaml.template), enabling it, to allow personal deployment to use the feature by default?

ableuler:
> should we add `anonymousSessions` also [here](https://github.com/SwissDataScienceCenter/sdsc-dev-docs/blob/master/dev-deployments/dev.values.yaml.template), enabling it, to allow personal deployment to use the feature by default?

I thought about it and decided not to. On one side i agree it would be good if people developed with this feature turned on most of the time. On the other hand it adds complexity to the dev deployment (basically 3 additional commands to execute) so I was a bit hesitant to impose this on people. But I'd be more than happy to put it there by default :)

ableuler:
Thanks @lorenzo-cavazzi for the updates
I have deployed them in two new deployments:
https://tmp-on.dev.renku.ch (anonymous sessions enabled)
https://tmp-off.dev.renku.ch (anonymous sessions disabled)

charts/renku/requirements.lock
```diff
@@ -3,14 +3,14 @@ dependencies:
   repository: https://swissdatasciencecenter.github.io/helm-charts/
   version: 0.10.3-3a30de0
 - name: renku-ui
-  repository: https://swissdatasciencecenter.github.io/helm-charts/
-  version: 0.9.2-4538959
+  repository: file:///Users/ableuler/renku/src/renku-ui/helm-chart/renku-ui/
```
lorenzo-cavazzi, line 6: I guess this was added by mistake :smile: 
I would remove both `requirements` files from this PR since the version numbers will automatically be updated when the repositories' PR are merged

ableuler: Oops, it was indeed. No I wanted to put it there to make testing easier for people who just want to check if deployment works with prebuilt charts. I'll fix it.

docs/admin/index.rst
```diff
@@ -122,6 +122,41 @@ For instance, make a user admin on GitLab.
...
+-------------------------------------
+
+Interactive sessions launched from public notebooks can be made available to logged-out users. However,
+this feature is deactivated by default. In order to activate it, you need a secondary deployment of the
+renku-notebooks chart (including a secondary Jupyterhub as a dependency) to handle the anonymous
+("temporary") sessions.
+
+1. Set :code:`.Values.global.anonymousSessions.enabled: true` in your values.yaml file.
+2. Deploy renkulab (or update your existing deployment) just as you would otherwise.
+3. Create an additional kubernetes namespace:
```
rokroskar, line 135: can't this be automatically done? You could configure the name of the namespace in the values file with some sensible default and provide a template that gets rendered only if the global toggle switch is on.

charts/renku/values.yaml
```diff
@@ -575,13 +582,14 @@ notebooks:
     # rbac:
     #   enabled: true
     ## Set the notebooks service API token
-    # hub:
-    #   services:
+    hub:
```
rokroskar, line 585: why is this uncommented? the gateway setting should be applied in the default values for the notebooks service.

docs/admin/index.rst
```diff
@@ -122,6 +122,41 @@ For instance, make a user admin on GitLab.
...
+
+.. code-block:: bash
+
+   kubectl namespace create <namespace-name>
+
+4. Copy a necessary secret from your primary to your secondary namespace:
+
+.. code-block:: bash
+
+   kubectl get secret renku-jupyterhub-tmp-postgres -n <primary-namespace> -o yaml | grep -v '^\s*namespace:\s' | kubectl apply  -f -
```
rokroskar, line 145: can't this be templated?

ableuler:
> This looks pretty good! I don't think it's the end of the world to require a few additional steps for this feature - though I have some questions about why certain things aren't being templated.

It's true, in order to get this into a functional state asap I took some "shortcuts" here and didn't automate everything that could be automated. Based on your feedback I would suggest the following improvements:
- The extra namespace is created through the primary chart with a default name derived from the primary release name. This name can be overwritten through values.yaml.
- The command for copying the secret is templated as part of NOTES.txt and thus rendered/displayed when Renku is deployed (of course only when the feature is enabled).

@rokroskar what do you think?

rokroskar:
@ableuler that makes sense to me! I think the self-documenting part will make it relatively painless.

charts/renku/values.yaml
ableuler: Because I removed this from the default values in the notebooks service for the secondary `renku-notebooks` deployment.

rokroskar: IMO we shouldn't set any defaults for the various services here if at all possible - the values are provided here just as documentation.

ableuler: Why? Renku uses a dependency (renku-notebooks) in a specific way. The same dependency is used differently in other places. I think this is the natural place to put such settings.

rokroskar: yes but that dependency is part of renku :) So if we want a default set for all to use we should just set it in the dependency itself.

ableuler: Agree, but now we have a use case where a child chart is used independently of the renku chart with different settings and then it starts to be more like a regular dependency. I thought this was the cleaner way to avoid bloat in the values file for the secondary notebooks deployment, but if you have a very strong opinion about this I think setting the `gateway` key to `Null` in the in the deployment values file for the secondary notebooks deployment should do the job too.

rokroskar:
How come CI isn't running on this PR?

lorenzo-cavazzi:
I could successfully upgrade an old environment following the steps suggested after the normal `helm upgrade`, apart from a minor issue I described in the `renku-notebooks` PR.

I needed to add this to my values files
```
global:
  anonymousSessions:
    enabled: true
    postgresPassword:
      value:
      overwriteOnHelmUpgrade: true  # can be turned to false after the first upgrade
```

If anyone want to try it, the deployment is https://lorenzotest.dev.renku.ch/ and it contains the current last version of every component as in the PRs linked at the beginning.

ableuler:
> I needed to add this to my values files

Thanks @lorenzo-cavazzi. Please check the documentation added through this PR an let me know if anything should be added.

ableuler:
> How come CI isn't running on this PR?

GitHub CI pipelines seem to be skipped for PRs with merge conflicts... https://github.community/t5/GitHub-Actions/Run-actions-on-Pull-Requests-with-merge-conflicts/td-p/40229

rokroskar:
Hmm not sure - many PRs that are automatically created when other repos merge to master have conflicts here and the CI always runs.

ableuler:
As soon as I rebased with master CI started to run, as soon as master evolved (resulting in a new conflict), CI stopped again.

docs/admin/index.rst
```diff
@@ -122,6 +122,42 @@ For instance, make a user admin on GitLab.
...
+--------------------------------------
+
+Interactive sessions launched from public notebooks can be made available to logged-out users. However,
+this feature is deactivated by default. In order to activate it, you need a secondary deployment of the
+renku-notebooks chart (including a secondary Jupyterhub as a dependency) to handle the anonymous
+("temporary") sessions.
+
+1. Set :code:`global.anonymousSessions.enabled: true` in your values.yaml file. If you are
+   updating an existing deployment you have to set :code:`global.anonymousSessions.postgresPassword.overwriteOnHelmUpgrade: true`
+   for the upgrade and then back to `false` for upcoming deployments.
```
lorenzo-cavazzi, line 135: I would also mention something about `anonymousSessions.autoCreateTmpNamespace`, otherwise it's undocumented

charts/renku/templates/NOTES.txt
```diff
@@ -8,3 +8,26 @@ You have configured a demo user for the platform. The details (password) of this
...
+{{ else }}
+Please create a secondary namespace using the following command:
+$ kubectl create namespace {{ .Release.Namespace }}-tmp
+{{- end }}
+
+You have to copy a secret into this secondary namespace using this command:
+$ kubectl get secret renku-jupyterhub-tmp-postgres -n {{ .Release.Namespace }} -o yaml | grep -v '^\s*namespace:\s' | kubectl apply -n {{ .Release.Namespace }}-tmp -f -
+
+The command to create a values file for your secondary notebooks-service deployment:
+$ pipenv run ./make-values-tmp.py  --values-file <path-to-renku-values> --renku-namespace {{ .Release.Namespace }}
```
lorenzo-cavazzi, line 29: This command (as well as the following one) should be run from `renku-notebooks/helm-chart`, right? It should be mentioned here and also on the admin docs page

charts/renku/templates/NOTES.txt
```diff
@@ -8,3 +8,26 @@ You have configured a demo user for the platform. The details (password) of this
...
+
+Finally, the command for deploying the secondary notebooks service:
+$ helm upgrade --install {{ .Release.Namespace }}-tmp-notebooks --namespace {{ .Release.Namespace }}-tmp -f ./notebooks-tmp-values.yaml --timeout 1800  ./renku-notebooks
```
lorenzo-cavazzi, line 32: This is a minor suggestion, feel free to ignore it if you think it's not useful.

After the `helm upgrade` command finish, it would be nice to check if `anonymousSessions.postgresPassword.overwriteOnHelmUpgrade` is `true` and display a reminder to set it back to `false`.

docs/admin/index.rst
ableuler: It's something that is only necessary because of our dev cluster setup with Rancher and it's documented in `values.yaml`, so I'd prefer to leave it the way it is.

charts/renku/templates/NOTES.txt
ableuler: This is a very good suggestion but it's something that we should do for all secrets which are handled following the same helm-hooks logic. If we will stick to the helm-hooks based approach to automatic secret generation very much deploys on the pulumi story which is still open, so I'd like to table this for the moment.

charts/renku/templates/NOTES.txt
ableuler: good point, adapted.

docs/admin/index.rst
ableuler: I made the comment in the `values.yaml` file a bit more clear.

PR #991, opened 2020-04-16
docs: split tutorial

### LINKED ISSUE #988: docs: split the first steps tutorial into pieces
it's better to divide the tutorial into small chunks on pages, with clickable "next" button, to
* not have to scroll a long time on the page
* include links to more details for each section
* more easily refer back to specific pieces of the tutorial when you're working on your own project

* make a folder for "first_steps" in the "tutorials" folder, and separate out each of the chunks
* make sure the link to the tutorial doesn't change, or if it does, let UI know to update the links on renkulab and everywhere else in the docs and README pages 🧷 
* if it's easy to add the "next" button, add the "next" button

**Alternatives**
* collapsible sections

emmjab:
**First Steps Tutorial**
* splits the tutorial into smaller, more manageable chunks by separating the content from one long page into a directory of smaller pages
* does not change content (there are other tickets open for this)
* DOES change the name of the first steps page from `tutorials/firststeps.html` to `tutorials/01_firststeps.html` -> this might be annoying for changing links; i did this because i used a glob for the toc tree and otherwise the pages are out of order; there's only 3 tutorials so far but it will grow and it can be annoying to keep track of the index pages, but i can switch it back on request)

**TOC Tree**
* fixes the TOC tree (as in, sets the conf param to keep it in place while the rest of the content moves)
* uses glob toc trees for directories of pages of content
* adds "next" and "previous" to the pages

**for review**
* make sure links to the first steps still work from other places within the docs (should be ok because the directive name didn't change
* make sure links from the first steps still work 

---

closes #988


ciyer:
Looks good! 

I noticed one strange thing, though. On smaller window sizes, the navigation in the old version was at the bottom:

image

Now it's at the top of the page:

image

Any idea why that might be? I think it is a little nicer at the bottom, but if it is not easy to do, I don't think it is a dealbreaker?

emmjab:
Thanks @ciyer !!

> Any idea why that might be? I think it is a little nicer at the bottom, but if it is not easy to do, I don't think it is a dealbreaker.

I didn't notice! (and it wasn't fully clear to me what the difference is in the screenshots -- also nothing i'm seeing is green, where does that come from 😄 )

rokroskar:
btw @ciyer @lorenzo-cavazzi @vfried this will break the links in the UI (and possibly other places) to the tutorial pages.

emmjab:
ah yeah @rokroskar -- we decided to hold off merging until the UI changes are ready to push this week; there's also links on discourse that I will update after merge!

PR #992, opened 2020-04-16
chart: Improve deployment process for anonymous notebooks service

ableuler:
This PR adds a script which will handle the entire secondary deployment. It contains the previous logic which creates a values file for the secondary deployment from the primary renku values file. 

It is invoked like this
```
pipenv run ./deploy-tmp-notebooks.py --release-name andreas-renku  --renku-namespace andreas  --extra-values ./extra-values.yaml
```
and has the following arguments:
```
  --release-name RELEASE_NAME
                        Name of the primary Renku helm release. REQUIRED.
  --renku-namespace RENKU_NAMESPACE
                        Namespace of the primary Renku deployment. REQUIRED.
  --tmp-namespace TMP_NAMESPACE
                        Namespace for the secondary renku-notebooks
                        deployment. OPTIONAL.
  --kube-context KUBE_CONTEXT
                        Kubernetes context to use. OPTIONAL.
  --extra-values EXTRA_VALUES
                        Path to an additional values file for the secondary
                        notebooks-deployment. These values will override the
                        ones automatically derived from primary Renku
                        deployment. OPTIONAL.
  --no-cleanup          Do not remove the generated values file after
                        deployment.
  --no-deploy           Only display the helm command for deployment, do not
                        execute it.
```


rokroskar:
Can you add this to the CI scripts?

ableuler:
> Can you add this to the CI scripts?

Yes, that's the plan and there's already an issue for it. But I wanted to merge this PR first as it's a self-contained improvement useful for anybody trying to deploy the new feature.

pameladelgado:
could we add the output file path an optional argument to the script?

also, I'm not sure I totally like the `tmp` prefix in all of the things related to this extra deployment. `tmp` to me means temporary, which is not the case here. Maybe make it `anon` or some shortened version of `anonymous sessions`

rokroskar:
> > Can you add this to the CI scripts?
> 
> Yes, that's the plan and there's already an issue for it. But I wanted to merge this PR first as it's a self-contained improvement useful for anybody trying to deploy the new feature.

Isn't it just adding one line to the CI script? Or is it more complicated than that?

ableuler:
> Isn't it just adding one line to the CI script? Or is it more complicated than that?

To be hones I've just never looked at the CI part so far. But it's true that one line should probably do. Where are the ci renku values taken from? (`CI_RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}`)

ableuler:
> could we add the output file path an optional argument to the script?
It would for sure be possible, but with the changes in this PR that file becomes temporary and is removed right after deployment. The foreseen way to modify settings on the secondary deployment (like resource options, etc) is through the `--extra-values` argument. Values passed in like that will always override the ones automatically inferred from the main Renku deployment values file.

> also, I'm not sure I totally like the `tmp` prefix in all of the things related to this extra deployment. `tmp` to me means that the extra deployment/namespace is temporary, which is not the case. Maybe make it `anon` or some shortened version of `anonymous sessions`
Yes, I see your point. The tmp comes from the fact that the whole thing is built around Jupyterhubs [tmpauthenticator](https://github.com/jupyterhub/tmpauthenticator). But I agree that this it's not the best naming. We can either rename things on the level of this script and the documentation as a part of this PR or create a new issue for renaming everything including the k8s resources. I have a tendency for the former variant... What do you think?

rokroskar:
> Where are the ci renku values taken from? (CI_RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }})

Yes, they are stored as a secret in the repo and passed into an environment variable.

rokroskar:
I like that idea @pameladelgado!

ableuler:
yeah, why not - at least until the the teardown works reliably that sounds like a good option. But I would probably add that label per default for upgrades of the `renku-notebooks` dependency.

pameladelgado:
do we need this for *every CI*? I would like to avoid having even more namespaces and more resources used when they are not absolutely needed.
Would it make sense to use a label for the PRs that want to include this feature and chose the appropriate CI_RENKU_VALUES and not do it by default?

.github/workflows/pull-request-test.yml
```diff
@@ -69,11 +78,13 @@ jobs:
       env:
         KUBECONFIG: ${{ github.workspace }}/renkubot-kube.config
         RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_KUBECONFIG }}
-        CI_RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
+        CI_RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES_ANON }}
```
pameladelgado, line 81: do we need this for every CI? I would like to avoid having even more namespaces and more resources used when they are not absolutely needed.
Would it make sense to use a label for the PRs that want to include this feature and chose the appropriate CI_RENKU_VALUES and not do it by default?

pameladelgado:
hahaha, sorry for the spam! I made the comment when github was down :D

ableuler:
> hahaha, sorry for the spam! I made the comment when github was down :D

haha, I already thought that you were really insisting on this point - like **really** 😄

pameladelgado:
> I would probably add that label per default for upgrades of the renku-notebooks dependency.

I agree @ableuler

ableuler:
As discussed I separated out the ci parts into another [draft PR](#1013) so that we can move on with this PR.

ableuler:
> I tested and the deployment is improved! thanks @ableuler !
> A few things to note:
> 
> * we need to include a Pipfile or some way of getting the right dependencies for executing the script in the release
> * the documentation is better as well, I would just move this new section to a child page and reference it in the main admin index page

Thanks for the review @pameladelgado. I have moved the new section to a dedicated sub-page. Also, I've added a remark that for the time being admins need to have the repo checked out and pipenv installed to enable the new feature.

PR #996, opened 2020-04-17
chart: remove renku requirements.lock from the repo

ableuler:
The requirements lockfile is generated when packaging the chart ([here I think](https://github.com/SwissDataScienceCenter/renku/blob/56e11abacb1e536fcb3619ae031d9fb784b3f914/.github/workflows/deploy.yml#L103)), it therefore does not need to be part of the repository. Removing it should help avoid conflicts when versions of dependencies are updated.


ableuler:
Tested the chart build, seems to work properly (see https://github.com/SwissDataScienceCenter/helm-charts/blob/gh-pages/renku-0.6.1-0e21b3f.tgz).

rokroskar:
Ironically, you now have a lock file conflict 😄

PR #1006, opened 2020-04-22
ci: do not add requirements.lock in update action

ableuler:


rokroskar:
could you check that the components which use this action are using `master` and not some tagged version?

ableuler:
I was wondering about this. Anything else besides renku-notebooks?

rokroskar:
renku-python is also using this action iirc

PR #1009, opened 2020-04-22
chart: update renku-ui dependency

ableuler:


lorenzo-cavazzi:
this was already automatically done by #1008 right?

ableuler:
true - that one is now coming through of course...

PR #1013, opened 2020-04-23
ci: activate anonymous notebooks in ci deployments

ableuler:
~~Work in progress, currently blocked by RBAC problems with Rancher-generated users.~~

PR now working and ready for review. Entire life-cycle - including teardown - seems to work.

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -52,6 +52,17 @@ jobs:
     needs: cleanup-previous-runs
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        sudo apt-get update -y && sudo apt-get install -y snapd
+        sudo snap install kubectl --channel=1.16/stable --classic
+        sudo snap install helm --channel=2.16/stable --classic
+        echo "::set-env name=PATH::/snap/bin:$PATH"
+        pip install --upgrade pip pipenv
+        pipenv install --deploy --dev --system
     - name: Configure environment
       run: |
         PR_NUMBER=${{ github.event.number }}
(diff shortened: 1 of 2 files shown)


rokroskar:
Nice! It looks like this is working?? It just needs the test configuration to be adapted.

pameladelgado:
> Nice! It looks like this is working?? It just needs the test configuration to be adapted.

Yes, this PR is no longer blocked by RBAC problems

ableuler:
> Nice! It looks like this is working?? It just needs the test configuration to be adapted.

@ciyer what config would have to be changed to test anonymous user sessions on ci-*** deployments?

rokroskar:
@ableuler presumably `RENKU_TEST_ANON_AVAILABLE` - see https://github.com/SwissDataScienceCenter/renku/tree/master/tests#running-1

ableuler:
Closing to test teardown.

ciyer:
Does this turn on anonymous notebooks in all CI deployments or just certain ones? It looks like all, in which case the best solution would be to update the logic in this method:

  https://github.com/SwissDataScienceCenter/renku/blob/master/tests/src/test/scala/ch/renku/acceptancetests/tooling/AnonEnv.scala#L31

ableuler:
> Does this turn on anonymous notebooks in all CI deployments or just certain ones? It looks like all, in which case the best solution would be to update the logic in this method:
> 
> https://github.com/SwissDataScienceCenter/renku/blob/master/tests/src/test/scala/ch/renku/acceptancetests/tooling/AnonEnv.scala#L31

I'd rather keep things explicit here. It took me quite some time to realize why the anonymous sessions are being tested at all on `dev.renku.ch` especially as the description of the defaults in the test README is gone.

.github/workflows/pull-request-test.yml
```diff
@@ -52,6 +52,17 @@ jobs:
...
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        sudo apt-get update -y && sudo apt-get install -y snapd
+        sudo snap install kubectl --channel=1.16/stable --classic
+        sudo snap install helm --channel=2.16/stable --classic
+        echo "::set-env name=PATH::/snap/bin:$PATH"
+        pip install --upgrade pip pipenv
+        pipenv install --deploy --dev --system
```
ableuler, line 65: As mentioned by @rokroskar, the primary namespace should be deleted too.

ableuler: done

PR #1015, opened 2020-04-23
wip on fixing teardown

ableuler:


.github/workflows/pull-request-teardown.yml
```diff
@@ -4,6 +4,15 @@ on:
...
 jobs:
+  github-env:
+    runs-on: ubuntu-latest
+    steps:
+    - name: get environment
+      run: env
+    - name: Dump GitHub context
+      env:
+        GITHUB_CONTEXT: ${{ toJson(github) }}
+      run: echo "$GITHUB_CONTEXT"
```
pameladelgado, line 15: if you want to get $GITHUB_CONTEXT variable written in the logs, instead of using `echo` you should use something like `echo "::warning file=app.js,line=1,col=5::Start configuration, github context: $GITHUB_CONTEXT"`

ableuler:
closing to test teardown.

.github/workflows/pull-request-teardown.yml
ableuler: sorry, should have labeled that as draft - my bad

pameladelgado:
sure, but I saw this `echo` in some other places where it does not lead to any variable and clearly is not printed in the github actions logs either. That's why I thought you might not be aware of this.

ableuler:
opened new PR as github does not seem to allow reopening a PR after force-pushing to the branch...? #1016

rokroskar:
No it doesn't - it's linked to a different ref - you can force push the other ref back, reopen, then force push the new changes again

rokroskar:
@pameladelgado what do you mean? it is printed in the logs

rokroskar:
as in here: https://github.com/SwissDataScienceCenter/renku/runs/611465930?check_suite_focus=true#step:3:1

pameladelgado:
@rokroskar 
aha! I remembered that when I was debugging inside a `stefanprodan/kube-tools@v1` action I needed to add the special debugging line otherwise it would not show up.
That, combined with the fact that the value is actually empty in some of the failed tear-down actions made me think `echo` does not work but it does, unless the output is redirected like in some github actions.

PR #1018, opened 2020-04-23
docs: Adds docs for new renku lfs features

Panaetius:
Updates docs for new renku LFS changes

Depends on https://github.com/SwissDataScienceCenter/renku-python/pull/1210


emmjab:
In case it's not clear -- the way to fix the docs build is to add `kb` to the list of spelling words in ` docs/spelling_wordlist.txt` :D

docs/user/lfs.rst
```diff
@@ -35,6 +35,24 @@ JupyterLab sessions by default. As a result, you do need to be aware of dealing
 with data stored in LFS if you want to use it efficiently in your work with
 renku.
 
+Renku LFS configuration
+-----------------------
+Renku by default only stores files larger than 100kb in LFS, to prevent
```
emmjab, line 40: Renku by default stores all files larger than 100kb in LFS to prevent slowing down git (and thus `renku`) with large files.

PR #1019, opened 2020-04-23
docs: updated docs for renku-graph

jachro:
This PR refreshes the docs for renku-graph services.


rokroskar:
This should now be merge-able, right @jachro ?

jachro:
Yes, only now as I've just updated the sequence diagrams to match with the API changes we agreed on.

rokroskar:
Thanks for updating these docs @jachro 

just a quick question: the Event Log service is listed as an actor in the webhook registration sequence diagram - is it actually doing anything there?

jachro:
It's not doing things directly but the webhook registration process kicks off the repository history loading process which creates data in the Event Log. So, as far as I remember, that was the reason I added it there. But maybe it makes the whole picture a bit confusing...

rokroskar:
Since it has no requests going to it, it looks like something is missing. I would remove it if it's involved only indirectly. I guess the same is true in the "missed commits synchronization job" - the TG picks up the events once they are in the event queue, but it isn't actually involved in the synchronization action.

PR #1029, opened 2020-04-24
fix: adapt tests to new dataset listing

ciyer:


rokroskar:
I guess you need to use the version of the UI that includes the changes?

ciyer:
Moved commit to #1021.

Issue #975, opened 2020-04-14
Add notebook launch test for logged-out user


ableuler:
The Renku integration tests should include a scenario where a logged-out user launches an interactive session.

Issue #976, opened 2020-04-14
Add notebook launch test for unprivileged user


ableuler:
There should be a test scenario where a logged-in user without developer access launches an interactive session.

PR #1037, opened 2020-04-27
tests for unprivileged environments

ciyer:


ableuler:
Do I interpret correctly that this will also test anonymous sessions iff they are enabled on the deployment which is being tested?

tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
```diff
@@ -189,7 +200,9 @@ class ProjectPage(projectDetails: ProjectDetails, userCredentials: UserCredentia
       }(waitUpTo(10 minutes), implicitly[source.Position])
 
     private def waitForImageToBeReady(implicit webDriver: WebDriver): Unit = {
-      find(cssSelector(".badge.badge-warning")) getOrElse fail("Image building info badges not found");
+      find(cssSelector(".badge.badge-warning")) getOrElse (findImageReadyBadge getOrElse fail(
```
jachro, line 203: I think you should be able to simplify that with:
`find(cssSelector(".badge.badge-warning")) orElse findImageReadyBadge getOrElse fail("Image building info badges not found")`

tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
```diff
@@ -205,16 +218,32 @@ class ProjectPage(projectDetails: ProjectDetails, userCredentials: UserCredentia
...
-      def connectToJupyterLab(implicit webDriver: WebDriver, spec: AcceptanceSpec): Unit = eventually {
+      def connectToJupyterLab(implicit webDriver: WebDriver, spec: AcceptanceSpec): Unit = {
+        connectToJupyterLab(s"a[href*='/jupyterhub/user/']")
+      }
+
+      def connectToAnonymousJupyterLab(implicit webDriver: WebDriver, spec: AcceptanceSpec): Unit = {
+        connectToJupyterLab(s"a[href*='/jupyterhub-tmp/user/']")
+      }
+
+      def connectButton(buttonSelector: String)(implicit webDriver: WebDriver): WebElement = eventually {
```
jachro, line 229: I'm guessing this is a private method...

rokroskar:
does this close #975?

tests/src/test/scala/ch/renku/acceptancetests/workflows/Environments.scala
```diff
@@ -0,0 +1,125 @@
...
+    And("the Environments tab")
+    click on projectPage.Environments.tab
+    And("they turn off the interactive session they started before")
+    click on projectPage.Environments.Running.connectDotButton
+    click on projectPage.Environments.Running.stopButton sleep (10 seconds)
+    Then("the session gets stopped and they can see the New Session link")
+    verify userCanSee projectPage.Environments.newLink
+  }
+
+  def launchAnonymousEnvironment(anonEnvConfig: AnonEnvConfig): Option[JupyterLabPage] = {
```
jachro, line 51: What do you think about making the `anonEnvConfig: AnonEnvConfig` `implicit`? It could make the spec a bit cleaner. Alternatively, as the `AnonEnv` is, I guess build for the `Environments`, we could potentially take it from the `MinimalFunctionalitySpec` and mix it in directly to the Environments. Perhaps, that could be even better than the implicit on the config. That would allow us to take the config args from all the `Environment` defs.

tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
```diff
@@ -205,16 +218,32 @@ class ProjectPage(projectDetails: ProjectDetails, userCredentials: UserCredentia
           .getOrElse(fail("Environments -> Running title not found"))
       }
 
-      def connectToJupyterLab(implicit webDriver: WebDriver, spec: AcceptanceSpec): Unit = eventually {
+      def connectToJupyterLab(implicit webDriver: WebDriver, spec: AcceptanceSpec): Unit = {
```
jachro, line 221: Tiny thing. You don't need the curlys here and in the next def.

tests/src/test/scala/ch/renku/acceptancetests/MinimalFunctionalitySpec.scala
```diff
@@ -34,6 +36,16 @@ class MinimalFunctionalitySpec
     removeProjectInGitLab
     verifyProjectWasRemoved
 
+    launchUnprivilegedEnvironment(anonEnvConfig)
+    stopEnvironment(anonEnvConfig.projectId)
+
     logOutOfRenku
+
+    launchAnonymousEnvironment(anonEnvConfig) match {
```
jachro, line 44: How do you feel doing that this way:
`launchAnonymousEnvironment(anonEnvConfig) map (_ => stopEnvironment(anonEnvConfig.projectId))`

ciyer:
> Do I interpret correctly that this will also test anonymous sessions iff they are enabled on the deployment which is being tested?

I thought about looking at the environments page and, if the "Anonymous environments are not supported" message is shown, skipping the test, but I think it is easy to introduce an error that will be missed then.

Instead, it is based on configuration (which I just realized I need to document...) . If the configuration says there are anonymous environments, it will try to test them. Otherwise, it will skip. The default configuration is there are anonymous environments on dev.renku.ch, but nowhere else.

ciyer:
> does this close #975?

Yes, thanks for pointing that out. It closes #975 and #976

ableuler:
> I thought about looking at the environments page and, if the "Anonymous environments are not supported" message is shown, skipping the test, but I think it is easy to introduce an error that will be missed then.

If you want to automate the detection looking at the `config.json` (see https://dev.renku.ch/config.json for example) would be the easiest.

tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
ciyer: I initially tried that as well, but that results in a complication error: "value getOrElse is not a member of Object."

tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
ciyer: I think that makes sense, but the initial implementation was also not declared private, and none of the other {foo}Button methods are private either, so I left it with default visibility. What do you think?

jachro: It's not the end of the world to leave it public but if I think it was left public by accident which I suppose it shouldn't.

ciyer: I suggest leaving it public, since all the other similar methods are public, but I think it might make sense to go through and convert all to private.

tests/src/test/scala/ch/renku/acceptancetests/MinimalFunctionalitySpec.scala
ciyer: You are right, that looks like more idiomatic scala.

tests/src/test/scala/ch/renku/acceptancetests/workflows/Environments.scala
ciyer: I have made this change.

I didn't initially implement it this way because I thought I would use different projects for the unprivileged test and anonymous test, but they now use the same project, so there is no reason against making the Config implicit anymore.

ableuler:
Thanks a lot @ciyer for resolving this. It's great to have this feature tested now! I'm not sure if you've seen my comment about reading the configuration (anonymous sessions on/off) directly from the deployment itself. Anyway, I've created a small issue for this as a later improvement (#1087).

PR #1039, opened 2020-04-28
added TestsDefaults; reorganized readme

jachro:


jachro:
Yes, that could work. Let me try to do that change.

jachro:
@ciyer I've just pushed the change

PR #1066, opened 2020-05-05
tests: pages to report loaded status

### LINKED ISSUE #1064: Verify browserAt page should check for element presence
As an acceptance-tests maintainer, I'd like the `verify browserAt page` should not only check page's URL and title but also presence of an element of a choice.

**Acceptance criteria:**
* the `Page` abstract class to force defining `def pageReadyElement: WebElement` (the def name might of course have a better name);
* the `Grammar.verify.browserAt` to check the presence of the element;
* add the elements to all the pages defined in the tests.

jachro:
closes #1064

### DIFF
--- a/charts/renku/requirements.yaml
+++ b/charts/renku/requirements.yaml
@@ -22,7 +22,7 @@ dependencies:
 - name: renku-graph
   alias: graph
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: 1.0.0
+  version: 1.0.1
   condition: graph.enabled
 - name: postgresql
   version: 0.14.4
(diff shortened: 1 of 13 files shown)


jachro:
Thanks Sekhar for looking at the PR.

So the reason why I made the
```def pageReadyElement(implicit webDriver: WebDriver): Option[WebElement]```
to return an option is that I was thinking about cases when we have a page we don't care that much (maybe a non-Renku page) and we simply return None so we can skip waiting for the elements to load. Do you think it's a possible scenario? If not, I can take the option out.

ciyer:
> Thanks Sekhar for looking at the PR.
> 
> So the reason why I made the
> 
> `def pageReadyElement(implicit webDriver: WebDriver): Option[WebElement]`
> 
> to return an option is that I was thinking about cases when we have a page we don't care that much (maybe a non-Renku page) and we simply return None so we can skip waiting for the elements to load. Do you think it's a possible scenario? If not, I can take the option out.

I suspected it was something like that. Then let us leave it like it is. It doesn't hurt, and it might save us a bit of work in the future.

PR #1075, opened 2020-05-06
docs: upgrading renku image version

emmjab:
WIP - waiting for myself to review https://github.com/SwissDataScienceCenter/renkulab-docker/pull/104 to update the naming conventions & readme.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -15,6 +15,7 @@ The Renku Project Docs
 
    About Renku <introduction/index>
    Getting started <getting_started>
+   Upgrading a Renku Project <upgrading_renku>
    Get in touch <get_in_touch>
    Roadmap <roadmap>
    Release notes <release_notes>
--- a/docs/upgrading_renku.rst
+++ b/docs/upgrading_renku.rst
@@ -1,0 +1,67 @@
+.. upgrading_renku:
+
+Upgrading a Renku Project
+=========================
+
+The Renku Project is in constant flux! This means that often there are changes
+to RenkuLab (e.g. the web platform (UI) and the knowledge graph (KG)), and ``renku``
+(the command-line interface (CLI), which is installed via ``pipx`` in the
+interactive environments).
+
+Using new features on RenkuLab
+------------------------------
+
+Often, changes to the RenkuLab UI and KG don't require you to do anything special
+to benefit from the new features. For instance, when we added the ability for
+anonymous (i.e. not logged in) users to launch interactive environments, it
+automatically applied for all projects.
+
+Other changes to the RenkuLab UI do
+require you to update project settings to make use of the features.
+
+You can read the CHANGELOG for more details.
+
+Using new features in the ``renku`` CLI
+---------------------------------------
+
+When we release a new version of ``renku``, the CLI, you do have to take actions
+if you want to update your project to the latest version.
+
+On RenkuLab, e.g. for Interactive Environments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In an Interactive Environment, the version of ``renku`` is controlled by the image
+built from the ``Dockerfile`` in the project. By default, the first line defines
+a pre-built RenkuLab compatible image (e.g. an image that can launch JupyterLab
+and/or RStudio, with a bunch of other convenient software installs). This image
+defines a fixed version of the ``renku`` CLI, indicated in the name.
+
+For example, ``renku/renkulab-py3.7:renku0.10.3-0.6.2`` has ``renku`` version 0.10.3.
+If a new release is available, you can check https://github.com/SwissDataScienceCenter/renkulab-docker
+for the latest version of the flavor you're using (e.g. python, cuda, R, Biocondutor, etc.).
+Then, replace the ``FROM`` line with the new image name, and commit & push the changes.
+Pushing the changes causes gitlab to build the new image. Then, you can launch a new
+interactive environment from the latest commit, and you'll have the latest version
+of renku CLI.
+
+You can read the CHANGELOG for the ``renku`` CLI here.
+
+
+Locally
+~~~~~~~
+
+If you are working locally (e.g. running ``renku`` commands in a terminal on your
+laptop), and you have installed renku with ``pipx`` (recommended), follow these
+instructions.
+
+To upgrade renku to the latest stable version:
+
+::
+
+    $ pipx upgrade renku
+
+To upgrade to the latest development version:
+
+::
+
+    $ pipx upgrade --pip-args=--pre renku


docs/upgrading_renku.rst
```diff
@@ -11,57 +11,52 @@ interactive environments).
...
+for anonymous (i.e. not logged in) users to launch interactive environments, all
+projects were immediately accessible to anonymous users without the project owner
+needing to change any settings.
 
-You can read the CHANGELOG for more details.
+.. _renku-cli-upgrade:
 
-Using new features in the ``renku`` CLI
----------------------------------------
+Using new features ``renku`` CLI features
```
rokroskar, line 24:
```suggestion
Using new ``renku`` CLI features
```

docs/upgrading_renku.rst
```diff
@@ -11,57 +11,52 @@ interactive environments).
...
 
-When we release a new version of ``renku``, the CLI, you do have to take actions
-if you want to update your project to the latest version.
+Note: This section is for upgrading the version of ``renku`` CLI installed into
+the Interactive Environments on RenkuLab for your project. See upgrading-local_
+for upgrading your local machine's version of ``renku``.
 
-On RenkuLab, e.g. for Interactive Environments
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+When we release a new version of ``renku``, the CLI, you do have to make some
```
rokroskar, line 31:
```suggestion
When we release a new version of the ``renku`` CLI, you do have to make some
```

docs/upgrading_renku.rst
```diff
@@ -11,57 +11,52 @@ interactive environments).
...
-Pushing the changes causes gitlab to build the new image. Then, you can launch a new
-interactive environment from the latest commit, and you'll have the latest version
-of renku CLI.
+1. Replace this line with the latest available image for the "flavor" you're using,
+   which you can find by:
 
-You can read the CHANGELOG for the ``renku`` CLI here.
+* checking the `renkulab-docker repo README.md <https://github.com/SwissDataScienceCenter/renkulab-docker/blob/master/README.md>`_
+  for the naming conventions of the "flavor" of image you're updating (e.g. if you're using
+  the "minimal python project", you will look in the py3.7 section)
```
rokroskar, line 43:
```suggestion
  the "minimal python project", you will look in the ``py`` section)
```

docs/upgrading_renku.rst
```diff
@@ -0,0 +1,62 @@
...
+for anonymous (i.e. not logged in) users to launch interactive environments, all
+projects were immediately accessible to anonymous users without the project owner
+needing to change any settings.
+
+.. _renku_cli_upgrade:
+
+Using new ``renku`` CLI features
+--------------------------------
+
+Note: This section is for upgrading the version of ``renku`` CLI installed into
```
rokroskar, line 27: If you want this to appear as a special "note" box, you can use the `note` directive - see https://docutils.sourceforge.io/docs/ref/rst/directives.html#specific-admonitions

docs/upgrading_renku.rst
```diff
@@ -0,0 +1,62 @@
...
+unless you absolutely need the latest development version of ``renku``, since it makes
+debugging and reproducibility much simpler.
+
+2. Push your changes (either from the GitLab editor, or in a running Interactive
+   Environment), and start up a new Interactive Environment to use this new image.
+
+3. When you start to run ``renku`` commands, you might be asked to call ``renku migrate``.
+   This command ensures that the metadata in your project is updated.
+
+Note: ``renku`` is installed in the image with ``pipx``. You can also upgrade ``renku``
```
rokroskar, line 59: same as above - use the `note` directive

docs/upgrading_renku.rst
```diff
@@ -0,0 +1,62 @@
...
+in a running Interactive Environment by following the docs for upgrading your local
+installation of ``renku``, however these changes will not hold through the next
+time you launch an Interactive Environment. Thus, we recommend updating the image.
```
rokroskar, line 62: do you want to add information here on how to add a `pipx` line to the `Dockerfile` to install a specific version of renku? e.g.

```
pipx install --force renku==0.10.3
```

rokroskar:
sorry, I approved on accident, meant to request changes...

docs/upgrading_renku.rst
emmjab: but we don't want people to do this, right?

rokroskar: why not?

rokroskar: in some cases it might be painfully long to rebuild the whole image, but just reinstalling renku is pretty easy

emmjab: same debugging/reproducibility reasons as usual

emmjab: also i could be wrong, but i think sometimes the layers after a new line are rebuilt anyway

rokroskar: not if they exist on the builder already

docs/index.rst
```diff
@@ -15,6 +15,7 @@ The Renku Project Docs
 
    About Renku <introduction/index>
    Getting started <getting_started>
+   Upgrading a Renku Project <upgrading_renku>
```
rokroskar, line 18: I just noticed that this is at the top level - do we really want it here?

emmjab: yeah :D 

my reasoning was that it's something that people don't think to do; if it's in the top level menu it's more visible (and we can keep iterating through what should go in which menu as we re-organize the topics in User (which I don't think should be nested into user at all))

rokroskar: but maybe we should think about putting it into a link on one of the first few pages, not in the ToC?

emmjab: why not put it in the ToC?

docs/upgrading_renku.rst
```diff
@@ -0,0 +1,69 @@
...
+
+2. Push your changes (either from the GitLab editor, or in a running Interactive
+   Environment), and start up a new Interactive Environment to use this new image.
+
+3. When you start to run ``renku`` commands, you might be asked to call ``renku migrate``.
+   This command ensures that the metadata in your project is updated.
+
+.. note::
+
+  ``renku`` is installed in the image with ``pipx``. You can also upgrade ``renku``
```
ciyer, line 64: This section on upgrading the renku version inside a running environment using pipx is true, but is it necessary to inform our users about it? I think we should either drop it, or explain why you might want to do this (for example, testing out a new version of renku to see if it works with your project).

docs/index.rst
rokroskar: because it's not a top-level topic?

docs/upgrading_renku.rst
rokroskar: it's also useful to do if you don't want to rebuild your whole image (or can't because of some dependency problem)

docs/upgrading_renku.rst
emmjab: 👍 i turned the note into a section that explains this

docs/index.rst
emmjab: i moved it into the "User" section and added the user section to the top level.

docs/upgrading_renku.rst
emmjab: 👍 i turned the note into a section that explains this

docs/user/upgrading_renku.rst
```diff
@@ -0,0 +1,89 @@
+.. upgrading_renku:
+
+Upgrading a Renku Project
```
rokroskar, line 3: I think here you're referring to a user's renku project, but in the line below you mean the Renku Project as a whole - maybe call it something different here? or say "Keeping your project up-to-date"

emmjab: yeah good point! fixing.

docs/user/upgrading_renku.rst
```diff
@@ -0,0 +1,89 @@
...
+like any other terminal install, will not hold through the next time you launch.
+
+Keep in mind that your project's metadata will be upgraded when you start running
+the new version ``renku`` commands - don't push these changes if you want to revert
+back to an older version.
+
+You should follow the steps in the previous section to swap out the base image
+once you're satisfied with latest ``renku``.
+
+Lazy upgrade
```
ciyer, line 76: Why do you call this a "lazy upgrade"?

emmjab: 😅 inspired by the term "lazy loading" ("to defer initialization of an object until the point at which it is needed"); since we prefer people to swap out the image if possible, but that they can "lazily" append the upgrade line to the dockerfile otherwise.

docs/user/cli-installation.rst
```diff
@@ -7,15 +7,15 @@ Interaction with the platform takes place via the Python-based command-line
 interface (CLI) and the Python API. Most available CLI commands are
 documented in our `cheatsheet <https://raw.githubusercontent.com/SwissDataScienceCenter/renku/master/docs/_static/cheatsheet/cheatsheet.pdf>`_
 
-Mac OS X
---------
+Using the CLI in an Interactive Environment
```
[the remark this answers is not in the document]
emmjab: agree! fixing.

docs/user/upgrading_renku.rst
ciyer: That's why I was confused:  it isn't lazily upgraded here in the sense of "lazy loading." I would take that to mean that the version gets upgraded the moment you try to do something that requires the new version, but that's not what is happening.

I think "Lightweight upgrade" would be a better term.

emmjab: well i mean "the point at which it's needed" as a layer in the Dockerfile; but i see that there are multiple interpretations. Lazy has a negative connotation to hopefully discourage people from doing this unless they have to. Lightweight has a positive connotation that might lead people to following it even if they don't have to.

docs/user/upgrading_renku.rst
rokroskar: I don't think this is fixed?

emmjab: I fixed the lines below it to just refer to user's renku project.

docs/user/upgrading_renku.rst
rokroskar: In light of the conversation/decisions today, some of this will need to be reworked anyway - should we postpone this PR until then? Or merge it as is and update?

emmjab: i'll update it to reflect the changes!

emmjab: but we might be in an uncanny valley because people with existing projects will have to add content to their dockerfiles & gitlab-ci, which won't be fun~

emmjab: so maybe let's merge and update, so that this version of the docs exists at some point in the history of the docs.

docs/user/upgrading_renku.rst
rokroskar: hmm this is still confusing - this is what it looks like:

image

"Renku Project" is used to mean two different things. Maybe we can call this section "Upgrading the image in your project"? Because it's not just upgrading the CLI it's also useful to know how to upgrade the image for a runtime update.

emmjab:
Whoops. That's true. I didn't notice.

emmjab:
What about "Upgrading your Renku Project"? That might help distinguish "your" and "a"... 

there's also template changes that i don't think i wrote about...

emmjab:
ok i fixed both headers @ciyer and @rokroskar; hope it's good enough for now 🤞

PR #1118, opened 2020-05-26
Bump gitlab and gateway versions

ableuler:
GitLab versions >= 12.9.0 require a different logout behaviour from the side of client applications, therefore we also have to update the gateway.

**Merge this PR instead of the auto-generated one after merging https://github.com/SwissDataScienceCenter/renku-gateway/pull/202**

Note that I've built and pushed the gitlab chart `0.5.0` to break the circular versions dependency created by the fact that the gitlab chart is a dependency of the renku chart and also part of the same repo...

### DIFF
--- a/charts/gitlab/Chart.yaml
+++ b/charts/gitlab/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for the Renku Gitlab server
 name: gitlab
-version: 0.4.1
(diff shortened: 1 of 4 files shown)


pameladelgado:
Looks good, thanks @ableuler !
Waiting on the new gateway release to test this :slightly_smiling_face:

ableuler:
> Waiting on the new gateway release to test this 🙂

That would be now :)

PR #1121, opened 2020-05-26
docs: add templates endpoints diagrams

lorenzo-cavazzi:
Add core service diagrams for the new templates endpoints implemented in SwissDataScienceCenter/renku-python#1254 and SwissDataScienceCenter/renku-python#1287


docs/_static/uml/core-templates-read.uml
```diff
@@ -0,0 +1,24 @@
...
+    boundary Worker
+    entity "Remote Git" as Git
+    entity "External Templates Repository" as External
+
+    == Read manifest from a templates repository ==
+    Client->Service: GET /<prefix>/templates.read_manifest
+    activate Service
+
+    Service->External: Clone templates repository
+    External->Filesystem: Persist project
```
ciyer, line 20: What does `persist project` mean in this context?

docs/developer/services/core_service.rst
```diff
@@ -100,3 +100,19 @@ An endpoint for editing dataset metadata. This command is equivalent to the ``re
 
 .. uml:: ../../_static/uml/core-datasets-edit.uml
 
+
+**GET /<prefix>/templates.read_manifest**
```
ciyer, line 104: Does it make sense to sort these operations alphabetically?

docs/_static/uml/core-templates-read.uml
lorenzo-cavazzi: I took inspiration from the description in the other charts.
It means that the project is saved to the local Filesystem cache so that it can be potentially used for other operations -- e.g. creating another project for the same user almost at the same time

PR #1122, opened 2020-05-26
Periodically remove old github actions artifacts

ableuler:
Looks like it's doing the job. The first attempt with a version that also triggered "on PR" deleted many artifacts.

### DIFF
--- a/.github/workflows/remove-old-artifacts.yaml
+++ b/.github/workflows/remove-old-artifacts.yaml
@@ -1,0 +1,20 @@
+name: Remove old artifacts
+
+on:
+  schedule:
+    # Every day at 1am
+    - cron: '0 1 * * *'
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/remove-old-artifacts.yaml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  remove-old-artifacts:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+    - name: Remove old artifacts
+      uses: c-hive/gha-remove-artifacts@v1
+      with:
+        age: '1 week'
```
rokroskar, line 17: how long have we been building the docs for?

ableuler: 3 weeks maybe?

PR #1145, opened 2020-06-02
set prometheus_multiproc_dir=/dev/shm/gitlab

fgeorgatos:
make the path explicit, especially to help with version of gitlab<12.7


rokroskar:
This will still use `/dev/shm` which will be limited to 64MB so I don't think it fixes the problem.

fgeorgatos:
> This will still use `/dev/shm` which will be limited to 64MB so I don't think it fixes the problem.

the point is to make the size monitoring of the subdirectory simpler to implement

rokroskar:
Why is watching `/dev/shm/gitlab` different from watching `/dev/shm`? The bug materializes if the device fills up regardless of which subdirectory it is.

fgeorgatos:
as you please

fgeorgatos:
resolving.

This issue was meant to be of help in case other Renku sysadmins would hit the same bug, so that tracing cause would be a bit faster. In the meantime, latest gitlab versions solve both the originating issue as well as changing the filenames, which makes it more clear what is going on.

Feel free to reopen the issue if you believe otherwise.

PR #1150, opened 2020-06-03
docs: improve minikube docs and comments

lorenzo-cavazzi:


charts/example-configurations/external-gitlab-values.yaml
```diff
@@ -13,7 +13,7 @@ variables:
   gitlabClientSecret: &gitlabClientSecret  # Put GitLab Secret here!
   gitlabRegistryHost: &gitlabRegistryHost  # registry.gitlab.com or registry.renkulab.io
 
-# Comment out or remove this block when using renkulab.io/gitlab
+# Comment out this block when using renkulab.io/gitlab or any other GitLab on-premises
```
ableuler, line 16:
```suggestion
# Comment out this block when using `renkulab.io/gitlab` or any other GitLab running under `<some-host-name>/gitlab. Commenting this block will set the gitlab url prefix back to its default chart value (`/gitlab`).
```

docs/developer/setup.rst
```diff
@@ -179,9 +197,15 @@ Start the renku platform using helm:
     --set notebooks.jupyterhub.hub.services.gateway.oauth_redirect_uri=http://$RENKU_DOMAIN/api/auth/jupyterhub/token \
```
ableuler, line 197: I can't really make that suggestion here, but to be consistent with the docs maybe we should set the timeout to 3600s (two lines above).

lorenzo-cavazzi: good point!

PR #1156, opened 2020-06-04
test: create a new issue

ciyer:
Acceptance test that creates a new issue.

### DIFF
--- a/tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
+++ b/tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
@@ -5,11 +5,12 @@ import ch.renku.acceptancetests.pages._
 import ch.renku.acceptancetests.tooling.AcceptanceSpec
 
 import scala.language.postfixOps
+import scala.concurrent.duration._
 
 trait Collaboration {
   self: AcceptanceSpec =>
 
-  def verifyCollaborationIsEmpty(implicit projectDetails: ProjectDetails): Unit = {
+  def verifyMergeRequestsIsEmpty(implicit projectDetails: ProjectDetails): Unit = {
     val projectPage = ProjectPage()
     When("the user navigates to the Collaboration tab")
     click on projectPage.Collaboration.tab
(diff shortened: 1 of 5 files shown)


tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
```diff
@@ -18,4 +19,35 @@ trait Collaboration {
     Then("they should see a 'No merge requests to display' info")
     verify userCanSee projectPage.Collaboration.MergeRequests.noMergeRequests
```
jachro, line 20: Not entirely sure if that works properly. If I click the Collaboration tab (line 16) then (at least on the DEV as of 8/6/2020) the Issues tab gets displayed not the Merge Reguests. So shouldn't you first do:
```click on projectPage.Collaboration.MergeRequests.tab```
before you try to verify if the `No merge requests to display.` text is visible?

File: tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala [hunk 2]
```diff
@@ -18,4 +19,35 @@ trait Collaboration {
...
   }
+
+  def verifyIssuesIsEmpty(implicit projectDetails: ProjectDetails): Unit = {
+    val projectPage = ProjectPage()
+    When("the user navigates to the Collaboration tab")
+    click on projectPage.Collaboration.tab
+    And("they navigate to the Merge Requests sub tab")
```
jachro, line 27: Probably that should go like:
And("they navigate to the Issues sub tab")

tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
```diff
@@ -18,4 +19,35 @@ trait Collaboration {
...
+    click on projectPage.Collaboration.Issues.tab
+    Then("they should see a 'No issues to display' info")
+    verify userCanSee projectPage.Collaboration.Issues.noIssues
```
jachro, line 30: What would you say if we'd do the assertion in the spec/workflow instead of the Page class? What I mean is smth like that:
```projectPage.Collaboration.Issues.noIssues.text shouldBe 'No issues to display.'```
We do have a workflow for the tab so I think it's maybe a better place?

Potentially the `No merge requests to display.` could be done in a similar way...

File: tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala [hunk 4]
```diff
@@ -18,4 +19,35 @@ trait Collaboration {
...
+    And("they navigate to the Merge Requests sub tab")
+    click on projectPage.Collaboration.Issues.tab
+    // Give some time for the tab change to take effect
+    sleep (1 second)
+    And("the user clicks on the 'New Issue' button")
+    click on projectPage.Collaboration.Issues.newLink
+    And("they fill out the form")
+    val issueTitle = "test issue"
+    val issueDesc = "test description"
+    projectPage.Collaboration.Issues.NewIssue.createIssueWithTitleAndDescription(issueTitle, issueDesc)
```
jachro, line 46: Just a thought that maybe having the body of the `createIssueWithTitleAndDescription` method directly in the workflow here could be a bit better. Not so sure if with the current approach we are mixing up the responsibilities of the workflows (which I believe should be actions) with the page classes (which are page definitions). I know we took a similar approach in other places but maybe we can try to refactor these (not necessarily as part of this issue).

tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
ciyer: I do not follow this comment. It looks like it does first click on the MergeRequests.tab before checking the text:

```
  def verifyMergeRequestsIsEmpty(implicit projectDetails: ProjectDetails): Unit = {
    val projectPage = ProjectPage()
    When("the user navigates to the Collaboration tab")
    click on projectPage.Collaboration.tab
    And("they navigate to the Merge Requests sub tab")
    click on projectPage.Collaboration.MergeRequests.tab
    Then("they should see a 'No merge requests to display' info")
    verify userCanSee projectPage.Collaboration.MergeRequests.noMergeRequests
  }
```

Am I missing something?

tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
ciyer: I see what you mean, but it is not clear cut because the structure of the page itself depends on whether or not there are issues.

The test IMHO should check if there is an indication that there are no issues to see, whether the text of that indication matches some expectation is of less importance.

tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
ciyer: I am not convinced this will improve the tests. I agree that it is debatable whether filling out a form is part of the page definition, but keeping it in the page makes the workflow easier to read, since the code that fills out the form is neither particularly interesting nor relevant to the workflow; the workflow just cares that the form gets filled out and submitted.

tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
jachro: You are right. We got tricked by GitHub which skipped the `click on projectPage.Collaboration.MergeRequests.tab` line :)
Sorry

tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
jachro: 👍

tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
jachro: We agree that we should keep workflows as simple as possible but we'd still like workflows to be actions grouping entities, not pages. So what would you think if we have simple workflows responsible for things like `createIssueWithTitleAndDescription`? And maybe we could use backticks for naming workflows `def` so they look more like sentences and help to boost readability? (I mean to have ``  `create an issue with title and description` ``  instead of `createIssueWithTitleAndDescription`)

PR #1180, opened 2020-06-10
duplicated PR

ableuler:
Intential duplication of #1013 to avoid rate limiting.


.github/workflows/pull-request-teardown.yml
```diff
@@ -32,6 +32,8 @@ jobs:
           helm init --client-only
           echo "$RENKUBOT_KUBECONFIG" > renkubot-kube.config
           helm delete --purge $RENKU_RELEASE
+          helm delete --purge $RENKU_RELEASE-tmp-notebooks || true
+          kubectl delete namespace $RENKU_RELEASE-tmp || true
```
rokroskar, line 36: why not delete the other namespace as well?

ableuler: good point - if I create the primary namespace manually and not through helm, I should probably also remove it manually...

rokroskar: it doesn't matter - helm won't delete the namespace so if you want it gone it needs to be done manually. It wasn't done initially because of RBAC restrictions but now that seems to have been solved.

ableuler: > helm won't delete the namespace

Ah, not even the namespace it created? good to know...

rokroskar:
Should #1013 be closed?

rokroskar:
why is this labeled "do not merge"?

ableuler:
I thought I'd merge the original one as it has more of the context. By now the rate-limiting issue of the first one should be gone. Therefore also the `do-not-merge` label. I will add your suggestion to this PR to #1013 and close this PR.

PR #1196, opened 2020-06-12
fix: acceptance-tests to create artifact on failure

jachro:


rokroskar:
I'm not sure I understand what this is doing - weren't we uploading all these things already anyway?

jachro:
The thing is that after the change we did to the acceptance-tests so they run the flights tutorial on the machine where the tests are executed, we thought it would be good to package the whole project the tests creates and package it to the `test-artifacts` too. You can than unzip it locally and check what was the problem.

rokroskar:
Sure, but what I see is that the script copies things from the `target` directory to a different directory and then ships that as an artifact. The existing action was just bundling everything under `target`, so is the point here that you're just more selective about what gets packaged up?

jachro:
Yes, sorry, I forgot to mention before that the packaging step was failing while zipping the `target/<project-folder>/.renku/cache` with some file access errors. So we're removing things which we don't care and package things selectively.

PR #1202, opened 2020-06-16
fix: so the tests waits enough for a project to get created

jachro:


jachro:
Yes, I think so. There's only tiny thing which I believe would be good to tweak. At the moment the `pause asLongAsBrowserAt` waits patience interval time (30 sec) until it checks again if the current page is still the given one. And it tries 10 times. Maybe the checks should be done more frequently so it speeds up the test a bit.

ciyer:
> Yes, I think so. There's only tiny thing which I believe would be good to tweak. At the moment the `pause asLongAsBrowserAt` waits patience interval time (30 sec) until it checks again if the current page is still the given one. And it tries 10 times. Maybe the checks should be done more frequently so it speeds up the test a bit.

Maybe the wait could be changed to patience interval time * 1/3. I think is is good to have one central point of configuration for wait times, etc., but I see your point that, especially if we use this in more places, it can cause a notable slowdown in the test run duration.

jachro:
I added the `1/3` factor to the function.

PR #1219, opened 2020-06-26
docs: improvements to minikube setup docs

ciyer:


ableuler:
Thanks @ciyer for this suggestion! It's a much better description of what's happening, but would prefer to avoid the duplication into separate values files. We used to have separate files until recently for the two cases and they got out of sync quickly as people forgot to change values in one or the other place. 

How about taking the block
```
# global:
#   gitlab:
#     urlPrefix: ""
```
out completely, using your improved description to explain the situation and tell the user to `export GITLAB_URL_PREFIX=/` or `export GITLAB_URL_PREFIX=/gitlab` respectively. Then we add `--set global.gitlab.urlPrefix=$GITLAB_URL_PREFIX` to the deployment command and we're good.

ableuler:
We can do the suggested de-duplication [later](https://github.com/SwissDataScienceCenter/renku/issues/1258), at the moment this PR represents a welcome improvement over the status quo as it removes a pitfall that several people have gotten trapped by. Thanks @ciyer for adding this.

PR #1228, opened 2020-06-30
chore: add changelog for 0.6.4 release

ableuler:


pameladelgado:
I think the requirements are not correct, the bugfix release should be done on top of `0.6.3`'s commit [see](https://github.com/SwissDataScienceCenter/renku/blob/0.6.3/charts/renku/requirements.yaml).
Though I see a `0.6.3-bugfix` branch that seems to have the correct values, is there another PR for that?

pameladelgado:
ah sorry, I saw now that the tag and release have been done already.

PR #1245, opened 2020-07-06
ci: speed up tests by avoiding build

ableuler:
I wanted to try out the more advanced caching features that the new docker build kit offers and quickly applied this to our test setup. This PR uses the `renku/tests:latest` during the `sbt` build, thus shortening the tests by a few minutes. It relies on `renku/tests:latest` being built by the new docker build kit.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -61,6 +61,7 @@ jobs:
         RENKU_BOT_DEV_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
       run: |
         cd tests
+        COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose build sbt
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -61,6 +61,7 @@ jobs:
         RENKU_BOT_DEV_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
       run: |
         cd tests
+        COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose build sbt
```
ableuler, line 64: I don't know why, but setting the same environment variable on the `docker-compose run` command directly triggers a build using the standard build kit, therefore the two line.

rokroskar: you don't have to specify which image to use for the cache?

rokroskar: oh nevermind, it just uses `latest` by default

PR #1248, opened 2020-07-08
Notify slack when tests fail on CD deployment

ableuler:
It seems to me that failures on our master branch often go unnoticed and are not treated with the appropriate amount of urgency.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -67,6 +67,15 @@ jobs:
                            -e RENKU_TEST_EMAIL="renku@datascience.ch" \
                            -e RENKU_TEST_USERNAME="renku-test" \
                            -e RENKU_TEST_PASSWORD="$RENKU_BOT_DEV_PASSWORD" sbt
+    - name: Notify slack
+      if: failure()
+      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -67,6 +67,15 @@ jobs:
...
+      if: failure()
+      env:
+        RENKU_SLACK_BOT_TOKEN: ${{ secrets.RENKU_SLACK_BOT_TOKEN }}
+      run: |
+        curl -X POST https://slack.com/api/chat.postMessage \
+             -H "Authorization: Bearer $RENKU_SLACK_BOT_TOKEN" \
+             --data "channel=C9U45DL1H" \
+             --data "text=Acceptance tests on https://dev.renku.ch failed! :scream_cat:"
     - name: Prepare artifact for packaging on failure
       if: failure()
```
rokroskar, line 80: this refers to the previous step - in this case the previous step will be the "notify slack" step not the test step. Maybe these two steps should be combined?

ableuler: Failure should trigger when any of the previous steps fail: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#failure

rokroskar: Aha! All good then

PR #1259, opened 2020-07-13
build: change the sample values for ui templates

lorenzo-cavazzi:
These are the changes to the values file required by SwissDataScienceCenter/renku-ui#963


charts/values.yaml.changelog.md
```diff
@@ -5,6 +5,11 @@ Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
 ----
+## Changes on top of Renku 0.6.5
+* DELETE - *ui.templatesRepository*
+* NEW - *ui.templates*: the UI project creation has been overhauled and the settings are stored here.
```
ableuler, line 10:
```suggestion
* NEW - *ui.templates*: the UI project creation has been overhauled and can be configured here.
```

charts/values.yaml.changelog.md
```diff
@@ -5,6 +5,12 @@ Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
 ----
+
+## Changes on top of Renku 0.7.0
```
ableuler, line 9:
```suggestion
## Upgrading to Renku 0.8.0 (INCLUDES BREAKING CHANGES!)
```
I guess this should go into a `0.8.0` release as it requires changes to the values file, right?

charts/renku/values.yaml
```diff
@@ -432,6 +432,13 @@ gitlab:
   ## Set the JupyterHub URL
   # jupyterhubUrl: "http://jupyterhub.renku.build"
 
+  ## Configuration for Renku templates
+  templates:
+    # custom: true
```
ableuler, line 437: I guess these defaults are already set in the UI `values.yaml`?

ableuler (changes requested):

charts/renku/values.yaml
lorenzo-cavazzi: Yes they are. Shouldn't I add `templates` anyway here?

charts/values.yaml.changelog.md
lorenzo-cavazzi: I thought the version 0.7.0 is yet to be released, right? Should I move these changes in the already existing 0.7.0 section?

ableuler: ah yeah, ofc, sorry. Yes, just add it to the 0.7.0 section 👍

charts/renku/values.yaml
ableuler: It's fine the way it is. I'd personally be in favour of not repeating all configuration options from the chart dependencies, but it's consistent with what we've been doing so far.

rokroskar: I was going to ask the same thing - do you need `templates` to be uncommented?

rokroskar: oops nvm I missed that you changed it already 😄

ableuler (approved):

PR #1260, opened 2020-07-13
Update gitlab chart

ableuler:
Note that the gitlab 0.5.1 chart has already been built and published manually.


rokroskar:
Ideally we would move the gitlab chart to its own repo or at least directory

ableuler:
> Ideally we would move the gitlab chart to its own repo...

Or just make it a folder in our renku template directory.

rokroskar:
That's where it started and we factored it out

PR #1273, opened 2020-07-16
Stop using deprecated k8s beta APIs

ableuler:


ableuler:
> Looks good!
> We are just missing logging charts but those are deprecated, we should probably clean them up in another PR.

Yeah - I really just changed the APIs for the well-established concept doing a search-replace over the entire repo, cleaning up the logging chart is another story.

Issue #1271, opened 2020-07-16
Stop using deprecated k8s beta APIs


ableuler:
Several APIs which we are still using are approaching the end of their lifecycle:

- [x] apps/v1beta2 (deprecated in 1.16) --> apps/v1 (default for core workload since 1.9)

- [x] rbac.authorization.k8s.io/v1beta1 (deprecated in 1.20) --> rbac.authorization.k8s.io/v1 (introduced in 1.8)

- [x] ingress: extensions/v1beta1 (deprecated in 1.17) --> networking.k8s.io/v1beta1 (introduced in 1.14)

ableuler:
Closed by #1273.

PR #1285, opened 2020-07-19
Bump postgres chart version

ableuler:
See https://github.com/SwissDataScienceCenter/renku/issues/1271 for context. 

This PR should either be tested on a deployment which includes gitlab, or we should monitor closely that dev functions properly _after_ merging this PR and potentially roll back.

Update: The upgrade process of an existing deployment has been tested and is documented as part of this PR. Merging this PR needs some manual intervention.


rokroskar:
See also the changes in https://github.com/SwissDataScienceCenter/renku/pull/579 - iirc it was possible to remove lots of values from our values file when switching to the new charts. Also, we would ideally test the update process on a running deployment to see if there are any issues.

rokroskar:
you may want to delete the existing deployment + namespace to deploy everything fresh for the PR

ableuler:
> you may want to delete the existing deployment + namespace to deploy everything fresh for the PR

good point, this ci deployment is definitely very broken :)

charts/values.yaml.changelog.md
```diff
@@ -5,6 +5,9 @@ Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
 ----
+## Upgrading to Renku 0.7.0
+* EDIT - *postgresql.persistence.existingClaim*: Renku `0.7.0` upgrades the postgres chart dependency, which requires modification of the postgres data volume of existing deployments. See [this helper job manifest](https://github.com/SwissDataScienceCenter/renku/blob/master/charts/utils/migrate-pg-to-bitnami.yaml).
```
rokroskar, line 9: I think we need a stronger message here - this is a breaking change

charts/renku/values.yaml
```diff
@@ -203,155 +203,47 @@ keycloak:
...
-    ## Requires service.type: NodePort
-    # nodePort:
-
-  networkPolicy:
-    ## Enable creation of NetworkPolicy resources.
-    ##
-    enabled: false
+  image:
+    repository: bitnami/postgresql
+    tag: 9.6
```
rokroskar, line 232: not newer?

charts/values.yaml.changelog.md
ableuler: We've had those before without explicitly mentioning, but I agree it would be good practice, especially while we're pre 1.0 where semantic versioning is not as unambiguous as after 1.0. Adapted.

charts/renku/values.yaml
ableuler: I was considering, but 9.6 is supported until November 2021, and migrating to 10 (or even 11) would yet be another stunt. At the same time I think it's important to get the chart update out ASAP as our current chart with the current postgres dependency does not run on recent k8s versions. So I'd prefer to defer this upgrade.

PR #1307, opened 2020-07-28
build: remove jh optimizations

rokroskar:
This removes the continuous image puller, the user scheduler and the user placeholder deployments. They are not needed in our default configuration.

### DIFF
--- a/charts/renku/values.yaml
+++ b/charts/renku/values.yaml
@@ -632,10 +632,23 @@ notebooks:
     #   every: 60
     singleuser:
       image:
-        name: renku/singleuser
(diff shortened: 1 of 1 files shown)


charts/renku/values.yaml
```diff
@@ -632,11 +632,24 @@ notebooks:
     #   every: 60
     singleuser:
       image:
-        name: renku/singleuser
```
ableuler, line 635: Is this on purpose? just asking because it's a change that has nothing to do with the rest of this PR.

rokroskar: yes I got tired of looking at an image ref that is >1 year old

PR #1314, opened 2020-07-29
docs: add cookies documentation

lorenzo-cavazzi:
This is a follow up of SwissDataScienceCenter/renku-ui#949 , adding the relevant documentation about the cookie banner and the privacy page

### DIFF
--- a/docs/admin/index.rst
+++ b/docs/admin/index.rst
@@ -146,6 +146,59 @@ Check list:
     4. launch a notebook
   - You should be now able to follow "First steps"
 
+Additional configurations
+-------------------------
+
+Privacy page
+~~~~~~~~~~~~
+
+The UI has a privacy page with a completely configurable content, suited for showing
+any policy/terms related information, like the `Privacy Policy Statement` or the
+`Terms of Use`.
+
+The content is read from the privacy ``ConfigMap`` installed with RenkuLab, precisely
+the ``privacy_statement`` key.
+
+.. note::
+
+  The exact name of the privacy ConfigMap may slightly vary depending on your specific
+  RenkuLab configuration. The default is ``renku-ui-privacy``. All the variations
+  will end with `-privacy` in the name.
+
+We support the `Markdown syntax <https://en.wikipedia.org/wiki/Markdown>`_ for the
+privacy page content.
+
+You can refer to the current
+`ConfigMap in the renku-ui repository <https://github.com/SwissDataScienceCenter/renku-ui/blob/master/helm-chart/renku-ui/templates/configmap.yaml>`_
+to find an example of the default Markdown.
+
+
+Cookie banner
+~~~~~~~~~~~~~
+
+Since RenkuLab requires user identification and implements a login system, cookies are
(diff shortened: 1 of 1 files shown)


rokroskar:
Thanks for the quick turnaround on this! Maybe an abbreviated version of the instructions could also go here: https://github.com/SwissDataScienceCenter/renku/blob/master/charts/renku/templates/NOTES.txt#L5

lorenzo-cavazzi:
Sure I'll add a line there. I'm wondering if the configmap would be overwritten at each `helm upgrade` (I believe the answer is yes) and if there is a way to prevent it :thinking:

rokroskar:
That's a good point - I'm not sure. We have some conditionals for secrets that we don't want overwritten so maybe we need a similar mechanism

rokroskar:
Maybe the solution is to have a value like

```
ui:

  privacyPage:
    enabled: True
    configMap: <potential config map override>
```

but this has `enabled: False` by default and we don't actually put in a place holder config map at all. This way the admin has to set it up and then turn on the feature.

lorenzo-cavazzi:
So we leave the configMap file in the UI and we pilot it with `enabled: True`?
And the `configMap` value should contain the configmap name?

rokroskar:
No i think we remove the configmap from the templates in the UI chart. It makes it a manual process but it also means it won't get overwritten. It will only really need to be done once, presumably (or very very rarely)

lorenzo-cavazzi:
I'm changing the ui so that, by default, the privacy page is empty and the configmap is not mapped to a local volume, but the current configmap would still be there as a sample. One thing I don't like about this solution is that we should also remove the link to the privacy page from the cookie banner, making it just a short statement without a follow-up page.
I think we should use the `enabled` value to show/hide both the cookie popup and the privacy page at the same time.

lorenzo-cavazzi:
I ended up making the whole `privacy` feature easy to disable by providing `ui.privacy.enabled: false`.
I left a sample configMap in the UI repo, but the user needs to create a new one (the idea is that it's much easier and less error-prone to clone the sample) and set the proper values in `ui.privacy.page` to activate the features.
SwissDataScienceCenter/renku-ui#981

Let me know if this may work, I will quickly update the docs here.

docs/admin/index.rst
```diff
@@ -156,33 +156,32 @@ The UI has a privacy page with a completely configurable content, suited for sho
...
+enabling the privacy page.
 
 .. note::
 
-  The exact name of the privacy ConfigMap may slightly vary depending on your specific
-  RenkuLab configuration. The default is ``renku-ui-privacy``. All the variations
-  will end with `-privacy` in the name.
+  If you don't set the ConfigMap name and key,
+  `a sample <https://github.com/SwissDataScienceCenter/renku-ui/blob/master/helm-chart/renku-ui/templates/configmap.yaml>`_
+  will be used instead. You can start from is as a template to create your customized ConfigMap.
```
rokroskar, line 168:
```suggestion
  will be used instead. You can start from it as a template to create your customized ConfigMap.
```

PR #1322, opened 2020-07-31
Update 03_python_environment.rst

setgree:
A space between `commit -m` and the commit message. It's a tiny detail, everything looks great to me!

PR #1323, opened 2020-07-31
Update 06_create_workflow.rst

setgree:
The "under the hood" link to https://renku-python.readthedocs.io/en/latest/api.html Yields a "SORRY This page does not exist yet" error. I replaced it with something else from the docs that details Renku's data procedures in more detail.

### DIFF
--- a/docs/tutorials/first_steps/03_python_environment.rst
+++ b/docs/tutorials/first_steps/03_python_environment.rst
@@ -73,7 +73,7 @@ And we need to capture this change in git.
 .. code-block:: console
 
     git add requirements.txt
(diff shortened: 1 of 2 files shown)


rokroskar:
Hi @setgree thanks for the fixes! I've cherry-picked the commit from the [other PR](https://github.com/SwissDataScienceCenter/renku/pull/1322) so we can merge them at the same time.

rokroskar:
Ah, and actually, since our tests use secrets, they can't run on branches from forks. So I will close this and reopen it with the same commits from the main repo.

PR #1326, opened 2020-08-03
build: parametrize the renku-python version

rokroskar:
Introduces the means to parametrize the renku-python version in renku-graph and the tests.


charts/renku/requirements.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
 - name: renku-ui
   alias: ui
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: 0.11.0-a6fe6fe
+  version: 0.11.0-f9871b1
```
lorenzo-cavazzi, line 9: if you use this chart, you may want to merge the branch `703ui-integration-tests` that contains the relevant changes for the integration test

rokroskar: ah great, thanks, will do

PR #1348, opened 2020-08-07
Use virtual environments for configuring the test environment

travissarles:


tests/src/test/scala/ch/renku/acceptancetests/workflows/FlightsTutorial.scala
```diff
@@ -29,18 +29,34 @@ trait FlightsTutorial extends GivenWhenThen with Matchers with ScalatestMatchers
...
+    console %> (c"echo requests>=2.20.0" >> "requirements.txt")
+    console %> (c"echo jupyterhub==1.1.0" >> "requirements.txt")
+    console %> (c"echo nbresuse==0.3.3" >> "requirements.txt")
+    console %> (c"echo jupyterlab-git==0.20.0" >> "requirements.txt")
+    console %> (c"echo renku==0.10.4" >> "requirements.txt")
+
 
     And("installs the packages")
-    console %> c"pip install -r requirements.txt"
+    console %> c"./venv/bin/pip install -r requirements.txt"
```
rokroskar, line 52: why not activate the environment instead of specifying the full path into the venv directory? You have to then deactivate it at the end or in the case of an error.

tests/src/test/scala/ch/renku/acceptancetests/workflows/FlightsTutorial.scala
```diff
@@ -57,7 +73,7 @@ trait FlightsTutorial extends GivenWhenThen with Matchers with ScalatestMatchers
 
     And("runs step 1")
     console %> c"mkdir $projectFolder/data/output"
-    console %> c"renku run python src/00-FilterFlights.py data/201901_us_flights_1/2019-01-flights.csv.zip data/output/2019-01-flights-filtered.csv"
+    console %> c"./venv/bin/python -m renku.cli run ./venv/bin/python src/00-FilterFlights.py data/201901_us_flights_1/2019-01-flights.csv.zip data/output/2019-01-flights-filtered.csv"
```
rokroskar, line 76: If you activate the virtualenv you won't have to have full paths here which will make this quite a bit easier to read and potentially avoid some pitfalls for others who might be adding steps in the future

tests/install_renku.sh
```diff
@@ -0,0 +1,13 @@
...
+echo "Installed virtualenv version: $installed_virtualenv_version"
+
+if [[ ! "$installed_virtualenv_version" == *"$target_virtualenv_version" ]]; then
+  echo "WARNING: Incompatible version of virtualenv installed: $installed_virtualenv_version . Replacing with version $target_virtualenv_version"
+  python3 -m pip install virtualenv==$target_virtualenv_version
+fi
+
+virtualenv venv
+source ./venv/bin/activate
+pip install 'renku==%s'
```
rokroskar, line 13: you're installing `renku` inside the `FlightsTutorial` code - so what is the purpose of installing it here?

travissarles: I think previously we were installing it globally so it made sense to do it here but now i think this script should just be used to check the virtualenv version. I'll take out the lines which set up a virtualenv and install renku

tests/src/test/scala/ch/renku/acceptancetests/workflows/FlightsTutorial.scala
```diff
@@ -29,18 +29,34 @@ trait FlightsTutorial extends GivenWhenThen with Matchers with ScalatestMatchers
...
     And("adds some Python packages to the requirements.txt")
     console %> (c"echo pandas==0.25.3" >> "requirements.txt")
     console %> (c"echo seaborn==0.9.0" >> "requirements.txt")
+    console %> (c"echo renku==0.10.4" >> "requirements.txt")
+    console %> (c"echo papermill<1.2.0" >> "requirements.txt")
+    console %> (c"echo requests>=2.20.0" >> "requirements.txt")
+    console %> (c"echo jupyterhub==1.1.0" >> "requirements.txt")
+    console %> (c"echo nbresuse==0.3.3" >> "requirements.txt")
+    console %> (c"echo jupyterlab-git==0.20.0" >> "requirements.txt")
+    console %> (c"echo renku==0.10.4" >> "requirements.txt")
```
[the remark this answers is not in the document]
travissarles: Yes, I think we could do that.

tests/src/test/scala/ch/renku/acceptancetests/workflows/FlightsTutorial.scala
travissarles: The tests were hanging when I tried to activate the the environment. However, it might be possible to set the same environment variables that the activate script does to avoid specifying the full path.

tests/src/test/scala/ch/renku/acceptancetests/workflows/FlightsTutorial.scala
travissarles: I agree that specifying the full path is fragile and prone to typing errors

jachro:
I believe not. Closing...

PR #1389, opened 2020-08-19
chore: merge ui changes and update integration test

lorenzo-cavazzi:
TEMP - for testing purposes


.github/workflows/pull-request-test.yml
```diff
@@ -94,18 +106,21 @@ jobs:
...
                     -f values.yaml \
+                    -f tmpvalues.yaml \
                     --timeout 1800 \
                     --cleanup-on-fail \
                     --set gateway.gitlabClientId=${APP_ID} \
                     --set gateway.gitlabClientSecret=${APP_SECRET} \
                     --set notebooks.jupyterhub.auth.gitlab.clientId=${APP_ID} \
                     --set notebooks.jupyterhub.auth.gitlab.clientSecret=${APP_SECRET} \
                     --set global.anonymousSessions.enabled=true \
+                    --set graph.triplesGenerator.renkuPythonDevVersion="git+https://github.com/SwissDataScienceCenter/renku-python.git@${RENKU_PYTHON_VERSION}#egg=renku" \
```
jachro, line 123: I have to say I quite like the idea of using the same version of the CLI everywhere in the stack but I think we're not there yet in terms of TG. I think we'll be when we switch to the new log endpoint on the Core. For the time being, I'd be more confident if we'd leave the version of CLI on the TG as tested and built-in to the image and simply leave the `renku migrate` step in the flights tutorial as a part of the acceptance-tests. The reason for this is:
* our acceptance-tests just smoke tests the whole triples generation process and it's probably a bit too optimistic to say the whole process will work with any version of renku-python;
* as of now, TG has a fixed version of the CLI build into the image which is tested and which is going to be used in the PROD; we could change the PROD deployment process so it sets the CLI version to the TG but
  * as I said I don't feel comfortable with that because in most cases TG is very tightly coupled to a hand-picked version of CLI;
  * not so sure if it's worth changing the kind of established process much when we are about to move to the Core doing the `renku log` soon.

jachro:
I believe we still want to remove the `TEST_CLI_VERSION` property from the acceptance-tests args so the tests use either CLI installed on the docker image (CI run) or local CLI (dev machine run). If yes, @joke1196 @travissarles could you agree on that with @rokroskar and @lorenzo-cavazzi and do the change, please?

.github/workflows/pull-request-test.yml
rokroskar: But this would be fine if the TG always used the renku-python master branch, no? In any case, as long as the versions used have compatible metadata it doesn't matter too much.

lorenzo-cavazzi: I see your point. Currently, if we don't provide any `renkuPythonDevVersion`, the default is `master`, right? Which is not any more stable than providing a specific commit.
The alternative (until the core log endpoint is in place) of pinning the version in the values file in the TG seems fine, even if less flexible (but as you said, also less error-prone). That would require another change in the TG, right? Would it be fine to first merge this PR and then make the change on top of the latest knowledge-graph master and fix the IT again in a new PR here?
P.S. I hope I understood correctly your suggestion, otherwise please correct me :grin:

PR #1404, opened 2020-08-23
docs: update intro text

rokroskar:


README.rst
```diff
@@ -39,32 +39,74 @@ RENKU (連句)
...
-Renku project <https://renku.readthedocs.io/en/latest/introduction/index.html#renku-introduction>`_.
+Please review our `contributing guidelines
+<https://github.com/SwissDataScienceCenter/renku/blob/master/CONTRIBUTING.rst>`_
+before submitting a pull request.
 
 
-If you are looking for detailed **renku** command-line interface (CLI) documentation,
-consult the `renku-python docs <https://renku-python.readthedocs.io/en/latest/>`_.
+Getting in touch
+----------------
```
ciyer, line 96: I think this needs a bit more introduction:

```
There are several channels you can use to communicate with us; we monitor all of them, so your messages will always get to us, but communication will be slightly more streamlined if you pick a channel that most suits your purpose and needs.
```

ciyer (changes requested):
Nice improvement: this makes the documentation read better overall!

I have some comments on a few of the sections. I realize that much of this content was already there before, but in some cases moving the content requires additional changes to the text, and in some cases I just now noticed issues that were already there before.

rokroskar:
Thanks @ciyer I applied the comments.

README.rst
```diff
@@ -33,38 +33,78 @@ RENKU (連句)
...
+a web-based application and `Renku
+<https://renku.readthedocs.io/en/latest/introduction/renku.html#renku>`_, a
+command-line tool for managing code, data, workflows and making practical use of
+the Knowledge Graph.
 
+A public instance of **RenkuLab** is available at https://renkulab.io, and there
+are several other deployments at various institutions. To start exploring Renku,
+feel free to make an account try it out! You can follow the `first steps
+<https://renku.readthedocs.io/en/latest/tutorials/01_firststeps.html>`_ tutorial
+in python or R or `continue reading about the Renku project
```
ciyer, line 74: Though I think we should have R and Julia versions of the first-steps tutorial, we don't actually have anything except Python at the moment. So we can just say 

```
You can follow the `first steps
 <https://renku.readthedocs.io/en/latest/tutorials/01_firststeps.html>`_ tutorial
 or `continue reading about the Renku project...
```

ciyer (changes requested):
Just one more small thing, then we are ready to merge, I think.

README.rst
rokroskar: oops true, I don't know when R was added there...

ciyer (approved):
👍

PR #1411, opened 2020-08-25
Rename tests -> acceptance-tests to be clear

travissarles:
To avoid confusion with unit tests and to be consistent with the term we use in communication, rename "tests" to "acceptance-tests". I used `git mv` in order to preserve the file history. Otherwise git thinks that files were deleted and a huge diff is created.


rokroskar:
Make sure you also fix the references in the github actions workflows.

ciyer (dismissed):
Thanks for taking initiative on this. It definitely makes the repo layout clearer.

ableuler:
Thanks for this suggestion @travissarles. Can you please follow the suggestion of @rokroskar

> Make sure you also fix the references in the github actions workflows, see https://github.com/SwissDataScienceCenter/renku/pull/1411/checks?check_run_id=1025454937#step:5:26

to make the tests pass such that we can merge this PR.

travissarles:
@ciyer could I get another review?

rokroskar (approved):

PR #1442, opened 2020-08-31
chore(charts): remove traces of gitlab sudo token

ableuler:


pameladelgado:
I was just going to suggest to delete it from here
https://github.com/SwissDataScienceCenter/renku/blob/master/charts/renku/templates/tests/test-renku.yaml#L12
I guess this latest commit covers what was missing.

ableuler:
> I was just going to suggest to delete it from here
> https://github.com/SwissDataScienceCenter/renku/blob/master/charts/renku/templates/tests/test-renku.yaml#L12
> I guess this latest commit covers what was missing.

yep, I had initially missed that - good catch!

pameladelgado (approved):
Thanks! Looks good!

PR #1446, opened 2020-09-01
fix(charts): correct proxy-public port in ingress

ableuler:
While port 8000 is exposed on the container, 80 is the correct port of the service.


pameladelgado (approved):
thanks for fixing this! :+1:

PR #1453, opened 2020-09-02
chore(charts): add missing default to renku values

ableuler:


charts/renku/values.yaml
```diff
@@ -512,10 +512,11 @@ notebooks:
   jupyterhub:
     ## Configure rbac
     # rbac:
-    #   enabled: true
-    ## Set the notebooks service API token
+    #   enabled: true    
     hub:
+      baseUrl: /jupyterhub/
```
rokroskar, line 517: trailing slash?

ableuler: The notebooks service has template strings like
```
value: {{ .Values.jupyterhub.hub.baseUrl }}services/notebooks/
```
in its chart. Should be refactored there at some point.

rokroskar (approved):

PR #1484, opened 2020-09-11
chore: pin nbconvert dependency

rokroskar:
Pin `nbconvert` to solve incompatibility with `papermill`


ciyer (approved):
👍

rokroskar:
Bumping `papermill` to the current version seems better than insisting on an old version... please re-approve :)

ciyer (approved):
Nice!

PR #1487, opened 2020-09-12
chore(charts): add some documentation to values.yaml

ableuler:


charts/renku/values.yaml
```diff
@@ -540,18 +547,26 @@ notebooks:
...
-    
+
     proxy:
       ## Must be set explicitly, generate using `openssl rand -hex 32`
-      secretToken: 
+      secretToken:
 
-    ## Simple default setup for gitlab oauth
-    # auth:
+    ## Configuration for gitlab oauth
```
rokroskar, line 555: maybe worth explaining here which gitlab oauth this is referring to?

ableuler: Adapted the wording.

charts/renku/values.yaml
```diff
@@ -540,18 +547,26 @@ notebooks:
...
+
     proxy:
       ## Must be set explicitly, generate using `openssl rand -hex 32`
-      secretToken: 
+      secretToken:
 
-    ## Simple default setup for gitlab oauth
-    # auth:
+    ## Configuration for GitLab acting as OAuth2 provider for Jupyterhub.
+    auth:
```
rokroskar, line 556:
```suggestion
    ## Configuration for GitLab acting as OAuth2 provider for Jupyterhub.
    ## If deploying against an existing GitLab, the OAuth2 application has to be created 
    ## manually and the corresponding values added to the fields below.
    auth:
```

ableuler: These values always have to be set explicitly. We can currently not auto-create them. For this we will first have to tackle https://github.com/SwissDataScienceCenter/renku/issues/1490.

I'm not claiming that this PR improves the documentation aspect in this values file by much, I just wanted to quickly add references to the `cookieSecret` and the `cryptoKey` which both need to be set by the user but were not mentioned here at all.

rokroskar: it's been a long time since I've deployed renku from scratch with gitlab, but my impression was that we create those apps in gitlab automatically? Is that not the case anymore?

ableuler: Yes, that is the case. But the above block configures Jupyterhub to *use* GitLab as OAuth2 provider. I just adapted the wording a bit to make this clear.

rokroskar: I just figured the user will wonder where to find these values but ok we can leave it unspecified.

rokroskar (approved):

rokroskar (approved):

PR #1493, opened 2020-09-14
fix: do not create duplicate auto PRs

### LINKED ISSUE #1213: change how automatic PRs are handled
Right now the automatic PRs are created when there is a push to a branch matching `auto-update/**` - this means that we get duplicate PRs when the branches are updated because it counts as a push. 

Two options:

1. the PR could be made by the component pushing the branch via a github API call
2. the auto-PR action can stay in this repo, but does not run if a PR for that branch is already running

RenkuBot:
closes #1213

### DIFF
--- a/.github/workflows/auto-reqs-update-pr.yml
+++ b/.github/workflows/auto-reqs-update-pr.yml
@@ -15,7 +15,7 @@ jobs:
           [[ ${GITHUB_REF} =~ $regex ]]
           pr_repo=$BASH_REMATCH
           # find other PRs from that repo
-          other_prs=$(curl -s ${renku_repo}/pulls | jq ".[] | select(.head.ref | test(\"auto-update/${pr_repo}\")) | .number")
+          other_prs=$(curl -s ${renku_repo}/pulls | jq ".[] | select((.head.ref | test(\"auto-update/${pr_repo}\")) and (.head.ref != \"${GITHUB_REF:11}\")) | .number")
           for pr in $other_prs ; do curl -s -H "Authorization: token $GITHUB_TOKEN" -X PATCH -d '{"state": "closed"}' ${renku_repo}/pulls/${pr} ; done
         env:
           GITHUB_TOKEN: ${{ secrets.RENKU_CI_TOKEN }}


.github/workflows/auto-reqs-update-pr.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           [[ ${GITHUB_REF} =~ $regex ]]
           pr_repo=$BASH_REMATCH
           # find other PRs from that repo
-          other_prs=$(curl -s ${renku_repo}/pulls | jq ".[] | select(.head.ref | test(\"auto-update/${pr_repo}\")) | .number")
```
ableuler, line 18: hmm sorry, don't get it yet: `.head.ref` will be `auto-update/renku-testing-1234`, but won't GITHUB_REF be something like `refs/pull/1493` ?

rokroskar: don't think so - the event is a branch push not a PR update.

.github/workflows/auto-reqs-update-pr.yml
```diff
@@ -15,7 +15,7 @@ jobs:
...
+          other_prs=$(curl -s ${renku_repo}/pulls | jq ".[] | select((.head.ref | test(\"auto-update/${pr_repo}\")) and (.head.ref != \"${GITHUB_REF:11}\")) | .number")
```
ableuler, line 18:
```suggestion
          other_prs=$(curl -s ${renku_repo}/pulls | jq ".[] | select((.head.ref | test(\"auto-update/${pr_repo}\")) and (.head.ref != \"${GITHUB_REF:12}\")) | .number")
```
I think you also have to remove the slash.

.github/workflows/auto-reqs-update-pr.yml
ableuler: ok, get it!

ableuler (changes requested):

.github/workflows/auto-reqs-update-pr.yml
rokroskar: ah thanks, in the test string I had I accidentally removed the first slash...

rokroskar:
obviously it doesn't work

image

😢

.github/workflows/auto-reqs-update-pr.yml
rokroskar: turns out I had it right :) the counting starts at zero

```
> echo $REF
refs/heads/auto-update/renku-testing-1234
>  echo ${REF:11}
```

ableuler: aaa, sorry - I thought you wanted to cut the `auto-update/`, from `auto-update/renku-testing-1234` which wouldn't make much sense. Sorry, my bad.

ableuler (approved):

.github/workflows/auto-reqs-update-pr.yml
rokroskar: no worries, I'm usually off by one on these sorts of things :)

PR #1497, opened 2020-09-15
Move CI to helm3

pameladelgado:


.github/workflows/pull-request-test.yml
```diff
@@ -86,20 +86,26 @@ jobs:
...
         helm repo add jupyterhub https://jupyterhub.github.io/helm-chart
         rm -rf charts/renku/charts
         helm dependency update charts/renku
         printf "%s" "$CI_RENKU_VALUES" | sed "s/<replace>/${RENKU_RELEASE}/" > values.yaml
-        kubectl create namespace $RENKU_NAMESPACE || true
+        curl -u "token-fxgsb:$RENKUBOT_RANCHER_APISECRET" \
+        -X POST \
+        -H 'Accept: application/json' \
+        -H 'Content-Type: application/json' \
+        -d '{"containerDefaultResourceLimit":null, "name":${RENKU_RELEASE}, "projectId":"c-l6jt4:p-rz88m", "resourceQuota":null}' \
```
ableuler, line 103: Guess it should be `RENKU_NAMESPACE` here, otherwise looks good! 👍

ableuler (changes requested):

.github/workflows/pull-request-test.yml
pameladelgado: this is not yet ready for review

ableuler: oops sorry, misinterpreted the state of the PR

pameladelgado: yep, I did not put WIP or similar because I wanted the github actions to run. I also need to test the teardown.

pameladelgado:
Closing to test teardown

pameladelgado:
That seems to have worked :tada:

.github/workflows/pull-request-teardown.yml
```diff
@@ -24,15 +24,21 @@ jobs:
...
           export PR_NUMBER=${{ github.event.number }}
           echo $PR_NUMBER
           export RENKU_RELEASE="ci-${PR_NUMBER}-renku"
-          helm init --client-only
+          export RENKU_NAMESPACE="ci-${PR_NUMBER}-renku"
           echo "$RENKUBOT_KUBECONFIG" > renkubot-kube.config
-          helm delete --purge $RENKU_RELEASE
-          helm delete --purge $RENKU_RELEASE-tmp-notebooks || true
+          helm uninstall $RENKU_RELEASE --namespace $RENKU_NAMESPACE
+          kubectl delete all --all --namespace $RENKU_NAMESPACE
```
ableuler, line 35: I think you could simplify this and just delete the namespace without previously deleting the resources in it.

.github/workflows/pull-request-teardown.yml
```diff
@@ -24,15 +24,21 @@ jobs:
...
+          kubectl delete --all secrets --namespace $RENKU_NAMESPACE
+          kubectl delete ns $RENKU_NAMESPACE
+          helm uninstall $RENKU_RELEASE-tmp-notebooks --namespace $RENKU_NAMESPACE-tmp || true
```
ableuler, line 38: I think the `|| true` is a legacy from the time when not all ci-deployments had anonymous notebooks enabled. I'd remove it to let the pipeline fail rather than report success while something has gone wrong during the removal of the tmp deployment/namespace.

ableuler (changes requested):
Fixing these pipelines is always a very arduous task - thanks a lot for persevering! 💪 
Just two minor suggestions, rest looks good!

.github/workflows/pull-request-test.yml
```diff
@@ -86,20 +86,25 @@ jobs:
...
         helm repo add gitlab https://charts.gitlab.io/
         helm repo add jupyterhub https://jupyterhub.github.io/helm-chart
         rm -rf charts/renku/charts
         helm dependency update charts/renku
         printf "%s" "$CI_RENKU_VALUES" | sed "s/<replace>/${RENKU_RELEASE}/" > values.yaml
-        kubectl create namespace $RENKU_NAMESPACE || true
+        curl -u "token-fxgsb:$RENKUBOT_RANCHER_APISECRET" \
+        -X POST \
+        -d "name=${RENKU_NAMESPACE}" \
+        -d "projectId=c-l6jt4:p-rz88m" \
```
ableuler, line 102: I was wondering if we want to hide user and project ID, but I actually think it's fine.

.github/workflows/pull-request-teardown.yml
pameladelgado: yes, I thought that we might still want to disable anonymous notebooks for some CIs, but if that's not the case I can just clean up those leftovers.

.github/workflows/pull-request-teardown.yml
pameladelgado: I would do what you propose but this is a bit tricky actually. Yesterday I've been hitting some issues when a namespace was stuck in `Terminating` state and had to manually delete them. Also, the first time I tried the teardown (that had the deletion of a namespace) not all of the resources were deleted not really sure why. I prefer better to be on the safe side for now.

.github/workflows/pull-request-test.yml
pameladelgado: yeah I thought about that too, I can add them to the secrets. Just FYI, here we have the cluster and project ID, not user ID.

charts/deploy-tmp-notebooks.py
```diff
@@ -110,6 +120,10 @@ def make_tmp_values(values, renku_namespace, out_path=None):
     # Add some reasonably short defaults for server culling
     new_values["jupyterhub"]["cull"] = {"enabled": True, "timeout": 3600, "every": 60}
 
+    # Disable user userScheduler
+    new_values["jupyterhub"]["scheduling"]["userScheduler"] = {"enabled": False }
```
rokroskar, line 124: I guess this will not be needed after https://github.com/SwissDataScienceCenter/renku-notebooks/pull/429?

pameladelgado: yes, I'll do a small PR to cleanup this and the main Renku values after merging that PR into this repo.

rokroskar: I'm not sure you need to bother. Having those deployed for another day won't hurt anything and it's just extra work for you. It's a problem that will correct itself soon.

.github/workflows/pull-request-test.yml
```diff
@@ -86,20 +86,26 @@ jobs:
         KUBECONFIG: ${{ github.workspace }}/renkubot-kube.config
         RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_KUBECONFIG }}
         CI_RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
+        RENKUBOT_RANCHER_APISECRET: ${{ secrets.RENKUBOT_RANCHER_APISECRET }}
+        CI_PROJET_ID: ${{ secrets.CI_PROJET_ID }}
       run: |
         echo "$RENKUBOT_KUBECONFIG" > renkubot-kube.config
-        helm init --client-only
+        helm repo add stable 	https://kubernetes-charts.storage.googleapis.com
```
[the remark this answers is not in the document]
pameladelgado: oops, some test leftover. Let me clean that up.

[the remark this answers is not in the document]
ableuler: I thought it was keycloak which in turn still has and old postgres as dependency, but I might be wrong.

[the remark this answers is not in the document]
rokroskar: ? keycloak doesn't deploy it's own postgres

charts/deploy-tmp-notebooks.py
rokroskar: what I'm saying is: just remove these two lines :)

pameladelgado: if I remove the two lines the checks would not pass

pameladelgado: I mean the PRs would not be able to deploy because they are not using Tiller which had more permissions than renkubot

rokroskar: Oh got it because they need service accounts made 👍

.github/workflows/pull-request-test.yml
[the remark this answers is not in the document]
ableuler: We don't use it like that, but postgres is still a chart dependency of keycloak. Either way I was wrong, it is our keylcoak version itself which is still from `stable`.

.github/workflows/pull-request-teardown.yml
ableuler: ok, if you've experienced this to work better like that, let's leave it.

ableuler (approved):

PR #1507, opened 2020-09-18
Release candidate 0.7.0

pameladelgado:


CHANGELOG.rst
```diff
@@ -1,5 +1,89 @@
...
+- **Template simplification**: We have decoupled the CLI version from the base image and
+  made it easier to override in your own environments. See the `project template
+  README <https://github.com/SwissDataScienceCenter/renku-project-template/blob/master/README.md>`_
+  for details.
+
+- **Improved editor**: All text editing components are using an enhanced editor
+  that allows seamless switching between WYSIWYG and markdown.
+
+
+- **``renku save``**: the Renku CLI now features a ``save`` command that
```
lorenzo-cavazzi, line 37: Not sure if this gets rendered correctly with `**''`

Screenshot_20201029_164146

lorenzo-cavazzi (dismissed):
🚀

ableuler (dismissed):
I would say let's 🚢 - it

rokroskar:
There is a small detail that is not mentioned - importing datasets from renku projects doesn't currently work for datasets in private repos - should this be mentioned somewhere?

pameladelgado:
> should this be mentioned somewhere?

yes, we should probably mention it somewhere. If this behavior was introduced in this release then it should go in the release notes, otherwise I would put it in the readthedocs part. Is there a plan to support this in the future?

ableuler (approved):

PR #1516, opened 2020-09-24
chore: minimal values script

pameladelgado:
* Add a script to automatically generate a valid renku values file.
* Add script to Admin documentation.
* Reorganize Admin documentation.


ableuler (commented):
Cool, I think this will be really helpful! 
Unfortunately, mac os comes with a very outdated bash version (3.2) which can not run the script. I think you can still leave it like as it is, but describing how to run the script inside a docker container would be nice. I quickly made an image (just ubuntu + openssl). Now the command 
```
docker run -v $(pwd):/script -w /script  ableuler/ubuntu-openssl:focal ./make-values.sh -d example.com -o ./myvalues.yaml -g gitlab.com -i someid -s somesecret
```
executes the script inside a container with the current working directory mounted.

rokroskar:
@ableuler which mac OS? On Catalina:

```
bash --version
GNU bash, version 5.0.7(1)-release (x86_64-apple-darwin18.5.0)
```

ableuler:
@rokroskar I'm on Mojave. But are you sure that you're using the default bash that came with Catalina? According to a quick google search Apple stays away from newer bash versions because they don't want to use any GPLv3 projects, and are even planning to switch to `zsh` in the future.

I'd just mention that the script needs bash >=4.0 and openssl installed, and point to the provided image for those who don't want to install these dependencies locally.

rokroskar:
@ableuler ah no you're right - I installed it with brew 🤷

docs/admin/prerequisites/dns.rst
```diff
@@ -3,7 +3,7 @@
 DNS
 ====
 
-Renku needs a DNS.
+Renku requires a DNS domain.
```
ableuler, line 6:
```suggestion
Renku requires a domain.
```

docs/admin/prerequisites/certificates.rst
```diff
@@ -27,9 +27,11 @@ For more information, please check `LetsEncrypt cert-manager Helm documentation
 Generate manually
 --------------------
 
-You can also use a self-signed certificate created manually.
-Add the previously created certificate as a secret to the ``renku`` namespace.
+You can also use an SSL certificate issued by a certificate authority.
```
ableuler, line 30: Maybe this could be a bit confusing as Letsencrypt is also a CA.

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
+#. Add an ``Identity Provider`` of type ```OpenID Connect v1.0``.
+#. Set ``Alias`` to <renku-domain>, ``Authorization URL`` to https://<gitlab-domain>/oauth/authorize , ``Token URL`` to https://<gitlab-domain>/oauth/token and ``Client ID`` and ``Client Secret`` to the values used in step 1.
+#. Click ``Save``. The new Identity Provider should appear and any user from the stand-alone GitLab instance should be able to login.
 
-Troubleshooting
+Updating RenkuLab
 ------------------
 
-If some Renku pods are not starting or present some errors please check the troubleshooting page.
+To update Renku to a newer version please check the **Upgrading from 0.x.x** section of the `release notes <https://github.com/SwissDataScienceCenter/renku/releases>`_ in case the renku-values.yaml file needs to be modified.
```
ableuler, line 205:
```suggestion
To update RenkuLab to a newer version please check the **Upgrading from 0.x.x** section of the `release notes <https://github.com/SwissDataScienceCenter/renku/releases>`_ in case the renku-values.yaml file needs to be modified.
```

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
-   to local content. Use the privacy page to add references to third party sites.
-2. ``layout``: The list of properties for the cookies consent plugin. We use
-   `Bootstrap <https://getbootstrap.com/docs>`_ for the UI, therefore all the Bootstrap
-   classes can be used to customize the appearance.
+If the RenkuLab deployment is using an existing GitLab, this application could be configured as an identity provider for RenkuLab's keycloak. This way all of the users from the existing GitLab instance could authenticate.
 
-You can refer to the current
-`values.yaml in the renku-ui repository <https://github.com/SwissDataScienceCenter/renku-ui/blob/master/helm-chart/renku-ui/values.yaml>`_
-to check the default values and to find an example of how to apply different settings.
+#. Go to https://<renku-domain>/auth (or to your stand-alone Keycloak dashboard), login to ``Admin console`` using admin username and the decoded password stored in ``keycloak-password-secret``.
```
ableuler, line 197:
```suggestion
#. Go to https://<renku-domain>/auth (or to your stand-alone Keycloak dashboard), login to ``Admin console`` using "admin" as username and the decoded password stored in ``keycloak-password-secret``.
```

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
-All the adjustments should be provided as values in ``ui.privacy.banner`` as follows:
+Independent GitLab as identity provider
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-1. ``content``: The HTML content of the cookie banner. Please mind that links should point
-   to local content. Use the privacy page to add references to third party sites.
-2. ``layout``: The list of properties for the cookies consent plugin. We use
-   `Bootstrap <https://getbootstrap.com/docs>`_ for the UI, therefore all the Bootstrap
-   classes can be used to customize the appearance.
+If the RenkuLab deployment is using an existing GitLab, this application could be configured as an identity provider for RenkuLab's keycloak. This way all of the users from the existing GitLab instance could authenticate.
```
ableuler, line 195:
```suggestion
If the RenkuLab deployment is relying on an existing GitLab instance, this instance could be configured as an identity provider for RenkuLab's keycloak. This way all of the users from the existing GitLab instance can log into RenkuLab with their existing login information.
```

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
-~~~~~~~~~~~~~
+GitLab deployed as part of Renku
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Since RenkuLab requires user identification and implements a login system, cookies are
-required at all levels to provide basic functionality. Please mind that cookies are also used
-for anonymous users (i.e. without an account or not currently logged in). To comply with
-international laws, it's strongly advised to explicitly require consent to the user for storing
-these data and using cookies.
+If your RenkuLab deployment includes GitLab you need to follow some additional steps to configure a user admin on GitLab.
```
ableuler, line 185:
```suggestion
If your RenkuLab deployment includes GitLab you need to follow some additional steps to configure an admin user on GitLab.
```

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
 
-To activate this feature, please set ``ui.privacy.enabled: true``. We have already configured a
-default cookie banner to inform the users about the aforementioned requirements and points to
-point them to the privacy page for further details.
+#. turn off automatic redirect to GitLab by setting redeploying with the value ``gitlab.oauth.autoSignIn: false``
```
ableuler, line 187:
```suggestion
#. turn off automatic redirection to GitLab by redeploying with the value ``gitlab.oauth.autoSignIn: false``
```

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
-`Terms of Use`.
-
-The content is read from a ``ConfigMap``. You need to configure the values in
-``ui.privacy.page`` to enable the feature and set the reference ConfigMap name and key.
-Both ``ui.privacy.enabled`` and ``ui.privacy.page.enabled`` need to be ``true`` for
-enabling the privacy page.
+   Troublesooting a Renku deployment <troubleshooting>
 
-.. note::
+**Note**. For deployments with existing GitLab and/or Keycloak, the post-install jobs need to be deleted manually in order for the Helm deployment to succeed.
```
ableuler, line 175: Are you sure? How does this work for our CI deployments?

docs/admin/index.rst
```diff
@@ -60,31 +73,65 @@ Ensure that PVs are added correctly and have the appropriate storage class.
...
+
+Further configuration
+^^^^^^^^^^^^^^^^^^^^^^
+
+You might want to modify some default values, here are some examples:
+
+   * Logged-out users can be given the permission to launch interactive sessions from public notebooks. This feature is turned off by default. For details on how to enable this feature, see the dedicated section below.
+   * A custom privacy notice and cookie banner can be configured (as of Renku 0.6.8). See the dedicated section below.
+   * Resources requests and limits can be adjusted for the diverse RenkuLab components
+   * For not independent GitLab you can configure an S3 backend for `lfs objects <https://docs.gitlab.com/ce/administration/lfs/index.html>`_ and `registry <https://docs.gitlab.com/ee/administration/packages/container_registry.html>`_, this configuration can be added in the ``gitlab`` section
```
ableuler, line 110:
```suggestion
   * For deployments that include GitLab you can configure an S3 backend for `lfs objects <https://docs.gitlab.com/ce/administration/lfs/index.html>`_ and `registry <https://docs.gitlab.com/ee/administration/packages/container_registry.html>`_, this configuration can be added in the ``gitlab`` section
```

docs/admin/index.rst
```diff
@@ -60,31 +73,65 @@ Ensure that PVs are added correctly and have the appropriate storage class.
...
+   * It is possible to dedicate some nodes for running user environments.
```
ableuler, line 111:
```suggestion
   * It is possible to dedicate some nodes for running interactive user sessions.
```

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
-
-The UI has a privacy page with a completely configurable content, suited for showing
-any policy/terms related information, like the `Privacy Policy Statement` or the
-`Terms of Use`.
-
-The content is read from a ``ConfigMap``. You need to configure the values in
-``ui.privacy.page`` to enable the feature and set the reference ConfigMap name and key.
-Both ``ui.privacy.enabled`` and ``ui.privacy.page.enabled`` need to be ``true`` for
-enabling the privacy page.
+   Troublesooting a Renku deployment <troubleshooting>
```
ableuler, line 173:
```suggestion
   Troubleshooting a RenkuLab deployment <troubleshooting>
```

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
 
-  If you don't set the ConfigMap name and key,
-  `a sample <https://github.com/SwissDataScienceCenter/renku-ui/blob/master/helm-chart/renku-ui/templates/configmap.yaml>`_
-  will be used instead. You can start from it as a template to create your customized ConfigMap.
+4. Post deployment configurations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The `Markdown syntax <https://en.wikipedia.org/wiki/Markdown>`_ is fully supported for the
-privacy page content.
+After Renku has been deployed you can make some post deployment configurations.
```
ableuler, line 180:
```suggestion
After RenkuLab has been deployed you can make some post deployment configurations.
```

ableuler (changes requested):
Thanks for putting in this work and sorry for not reviewing this earlier. I've added suggestions that should avoid the login problems we've seen on external deployments recently.

I've also a bunch of small comments for the documentation part.

charts/example-configurations/minimal-values.tmpl
```diff
@@ -0,0 +1,277 @@
...
+  baseURL:                   &baseURL                       https://[domain]
+  jupyterhubURL:             &jupyterhubURL                 https://[domain]/jupyterhub
+  gatewayURL:                &gatewayURL                    https://[domain]/api
+  jupyterhubAuthCallbackUrl: &jupyterhub_gitlab_callbackUrl https://[domain]/jupyterhub/hub/oauth_callback
+  oauth_redirect_uri:        &oauth_redirect_uri            https://[domain]/api/auth/jupyterhub/token
+  ingressTLS:                &ingressTLS                    [ingress-tls]
+
+  gitlabURL:                 &gitlabURL                     https://[gitlab-domain]
+  gitlabPrefix:              &gitlabPrefix                  [gitlab-prefix]           ## /gitlab if part of Renku, / otherwise
+  gitlabClientId:            &gitlabClientId                [gitlab-application-clientID]
```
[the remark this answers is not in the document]
pameladelgado: This is fixed in the newest version. I did not hard-code 'jupyterhub' because the script and template should also work for external GitLab deployments.

charts/example-configurations/minimal-values.tmpl
```diff
@@ -0,0 +1,277 @@
...
+    ## Key for encryption must be set explicitly!
+    ## See https://zero-to-jupyterhub.readthedocs.io/en/latest/reference/reference.html?highlight=cryptokey#auth-state
+      state:
+        enabled: true
+        cryptoKey:    [random]        ## openssl rand -hex 32
+
+## Configuration for the Gateway service
+gateway:
+  gitlabUrl: *gitlabURL
+  gitlabClientId: *gitlabClientId
```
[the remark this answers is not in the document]
pameladelgado: afaik this is needed for external GitLab deployments, so I left it there. We could revise these and the renku-gateway values when we do a chart cleanup.

charts/example-configurations/minimal-values.tmpl
```diff
@@ -0,0 +1,277 @@
...
+  gitlabClientSecret: *gitlabClientSecret
```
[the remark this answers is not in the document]
pameladelgado: latest version includes this and it works for external GitLab deployments

docs/admin/index.rst
pameladelgado: > Are you sure?

yes, we've been doing this for a couple of deployments with external GitLab and Keycloak in a regular basis. I _think_ the problem is for the keycloak deployment, which the CIs and personal deployments do have it enabled. Another issue is having the postgresql database externally, but this is a corner case and not covered in the minimal values script and template.
In any case, it seems clear that I should take out the GitLab part here.

charts/example-configurations/make-values.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
pameladelgado: mm using `env` might give unexpected results, it does not work for me.
I'd prefer to leave it as is for the moment.
There's a small section for MacOS users in the admin documentation which, together with Linux, should cover most users :slightly_smiling_face:

pameladelgado:
> Thanks for putting in this work and sorry for not reviewing this in detail earlier. I've added suggestions that should avoid the login problems we've seen on external deployments recently.
> 
> I've also a bunch of small comments for the documentation part.

Thanks a lot for the review. Indeed, this PR has gone stale, I'm reviving it and I incorporated some fixes I had locally. 
I also included some of your suggestions, in the documentation. :+1:

pameladelgado:
This is ready for review. The values generated with the script were validated for two cases: 
1. a deployment against an external GitLab, and 
2. a deployment that includes GitLab as part of the RenkuLab deployment.

docs/admin/index.rst
pameladelgado: > I should take out the GitLab part here

Done

rokroskar: so the post-install jobs run but never finish? If that's the case it sounds like we need to fix those.

pameladelgado: > so the post-install jobs run but never finish? If that's the case it sounds like we need to fix those.

They fail (iirc post-install keycloak job would never be able to connect to /auth since it has been deployed elsewhere) and as a consequence Helm thinks the deployment failed
```
Release "renku" failed: failed post-install: timed out waiting for the condition
```

So indeed we need to fix this. I created a PR that should fix this: https://github.com/SwissDataScienceCenter/renku/pull/1923 (indeed GitLab post-install job did not need this fix).

rokroskar: Right, I was going to suggest a fix like that - but in that case I would remove this word of warning here since once the other PR is merged this will no longer happen. 👍

pameladelgado: > Right, I was going to suggest a fix like that - but in that case I would remove this word of warning here since once the other PR is merged this will no longer happen. +1

yes, this was the intention. I'm going to delete this warning. :+1:

docs/admin/prerequisites/dns.rst
pameladelgado: I'd leave it there, I prefer to have a bit of redundancy to avoid ambiguity.

docs/admin/index.rst
```diff
@@ -94,105 +141,75 @@ Once all the pieces are in place, you can deploy Renku with the following comman
...
 
-After Renku has been deployed you can make some post deployment configurations.
-For instance, make a user admin on GitLab.
+Check list:
 
-1. turn off automatic redirect to GitLab by setting redeploying with the value ``gitlab.oauth.autoSignIn: false``
-2. log in as the root user using the password from ``gitlab.password``
-3. modify any users you want to modify (e.g. to make them admin)
-4. turn the automatic redirect back on
+  * Helm command completed successfully.
```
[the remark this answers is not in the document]
pameladelgado: I added instructions for the Helm test in a separate section. 
I find these checks to be complementary to the helm tests. Especially the quick checks, which help the admin user to understand better what could be wrong when certain functionality does not work and they are much faster/direct than running the acceptance tests.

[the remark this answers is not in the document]
ableuler: Yep - mentioning them further down is totally sufficient for me.

docs/admin/prerequisites/dns.rst
ableuler: ok

ableuler (approved):
LGTM 👍

PR #1541, opened 2020-10-04
Add egress network policy for Jupyterhub user pods

olevski:
This PR restricts the access that any user's Jupyterhub pods will have within the cluster. I tested things out with egress allowed only to the Jupyterhub pods everything works. Removing this causes the Jupyterhub session to not load at all when the user connects to the new environment in the ui. In addition to this egress is allowed for DNS resolution inside and outside the cluster and also for any IP outside the cluster.

The network policy works by allowing egress to all IPs except the IP/CIDR ranges reserved for internal network use. I did not know this but the IP ranges used internally in a network (i.e. like 192.168.0.1) are reserved and cannot be used/reserved for public use. I have wondered about this but wasn't really sure, here is the wikipedia page that mentions the ranges of IP addresses reserved for private network use https://en.wikipedia.org/wiki/Private_network.

Closes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/413. Not sure if this will happen automatically since the issue is in another repo.

Here are some get requests from within the jupyterhub session __without__ egress policy
```
 work ❯ test2 ▶ master ▶ $ ▶ curl -k https://10.43.0.1:443/livez?verbose --connect-timeout 60
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
 work ❯ test2 ▶ master ▶ $ ▶ curl -k https://www.google.com  --connect-timeout 60
<!doctype html><html itemscope="" itemtype="http://schema.org/WebPage" lang="de-CH"><head><meta content="text/html; charset=UTF-8" http-equiv="Content-Type"><meta content="/images/branding/googleg/1x/googleg_standard_color_128dp.png" itemprop="image"><title>Google</title><script nonce="GhnXSlKxxx7qcUtfBLvkvw==">(function(){window.google=.....
```

Here are some get requests from within the jupyterhub session __with__ egress policy
```
work ❯ test2 ▶ master ▶ $ ▶ curl -k https://10.43.0.1:443/livez?verbose --connect-timeout 60
curl: (28) Connection timed out after 60001 milliseconds
  work ❯ test2 ▶ master ▶ $ ▶ curl -k https://www.google.com --connect-timeout 60
<!doctype html><html itemscope="" itemtype="http://schema.org/WebPage" lang="de-CH"><head><meta content="text/html; charset=UTF-8" http-equiv="Content-Type"><meta content="/images/branding/googleg/1x/googleg_standard_color_128dp.png" itemprop="image"><title>Google</title><script nonce="wdy0Hp3Ui0DIi5RLkwW9Wg==">(function(){window.google.....
```

The policy is active on: https://tasko.dev.renku.ch/.

Questions about finalizing/polishing this PR:
- do we want to add options at the top level values.yaml file so that we can turn this on and off
- if the answer to the above is 'yes' do we want to provide additional options to the values.yaml chart for this
- the issue for this is in another repo - but I thought the easiest way to address it is here. Let me know if you think differently.

p.s. I referred to this documentation page on the kubernetes docs a lot: https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-allow-all-egress-traffic


charts/renku/templates/user-pod-egress.yaml
```diff
@@ -0,0 +1,42 @@
...
+    ports:
+    - port: 53
+      protocol: UDP
+    - port: 53
+      protocol: TCP
+  - to:
+    # allow connection to internal apps that is necessary
+    - podSelector:
+        matchLabels:
+          app: jupyterhub
```
ableuler, line 25: At the moment this allows connection to the jupyterhub proxy and other users notebooks servers too. By specifically selecting the `hub` component here you can narrow this down a little further.

ableuler (changes requested):
Great job! 👍  Apart from the small suggested modification this network policy seems to do exactly what we need. 
Now for the chart aspect, as yo already suggested yourself it this should definitely go into the renku-notebooks chart. Also, it would be nice to have a chart switch for it. It would suggest to use a global chart value for this which will allow controlling network policies on all sub-charts with only one value, and I think the default should probably be to have the network policies turned off. But these are definitely points to discuss. See for example [this thread](https://github.com/helm/charts/issues/4067) for a discussion on this.

ableuler:
And regarding your second question: 

> if the answer to the above is 'yes' do we want to provide additional options to the values.yaml chart for this

For me one global switch seems to be enough control at the moment. I do not see a need for specifying the details of the network policy used through the chart values. Also, I think it would be nice to not bloat the total interface of the chart too much, especially now that helm allows for some post-rendering customisation for rare use-cases.

olevski:
I am closing this since it should have been opened in the notebooks repo in the first place. The new PR is here: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/430

PR #1543, opened 2020-10-05
feat: acceptance test for test notebook launch from private repo

### LINKED ISSUE #1534: Add notebook launch from private repo to acceptance tests
Launching a notebooks from a private repository - especially if that repo contains [LFS data](https://github.com/SwissDataScienceCenter/renku/issues/1409) - would allow us to test many auth-related aspects of renku. Having these aspects tested would make all future refactoring of login/logout, token handling etc a much safer endeavour.

travissarles:
… - closes #1534


acceptance-tests/src/test/scala/ch/renku/acceptancetests/ReplShell.scala
```diff
@@ -1,8 +1,5 @@
-import org.openqa.selenium.{By, WebDriver}
-import org.openqa.selenium.chrome.ChromeOptions
-import org.openqa.selenium.remote.RemoteWebDriver
```
ciyer, line 3: This is a weird class because it is here to be used in the Scala REPL. It's true that the imports are not used in the code, but they are here so the names are available in an interactive REPL session, where they are useful. 

I think they should stay, but if you have a better way of expression that intention, then that would of course be great.

travissarles: I did some import optimization and it looks like they got removed :)

jachro: @travissarles I think we forgot to bring the imports back.

travissarles: Why do we need them? it seems like it still compiles

jachro (approved):
👍

PR #1545, opened 2020-10-05
chore: format acceptance tests

travissarles:


jachro:
Could we maybe add this few files as a single commit to the PR you're working on?

travissarles:
> Could we maybe add this few files as a single commit to the PR you're working on?

We could but we'll have some irrelevant formatting changes mixed with code changes. Is that okay given that there are only five files affected?

jachro:
Actually, you're right. Why not create a new PR. I believe these are all the formatting changes in the project. Did you maybe run the scalafmt from sbt?

travissarles:
> Actually, you're right. Why not create a new PR. I believe these are all the formatting changes in the project. Did you maybe run the scalafmt from sbt?

Yes, it seems to agree with the one from IntelliJ.

jachro:
I was thinking more about running
```
sbt scalafmt
sbt test:scalafmt
```

travissarles:
> I was thinking more about running
> 
> ```
> sbt scalafmt
> sbt test:scalafmt
> ```

After doing a `test:scalafmt` I noticed more formatting changes

rokroskar:
If you want to commit to a formatting style, there should be a format check in the CI pipeline.

travissarles:
> If you want to commit to a formatting style, there should be a format check in the CI pipeline.

Agreed. We have this for renku-graph.

.github/workflows/check-acceptance-test-fmt.yml
```diff
@@ -0,0 +1,15 @@
...
+  formatting-check:
+    name: Scala formatting check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Switch to acceptance tests directory
+        run: |
+          cd acceptance-tests
+      - name: run scalafmt
+        run: |
+          sbt test:scalafmt
```
jachro, line 15: I guess if the step suppose to be checking the formatting we should probably use `sbt scalafmtCheck` as it fails command. `sbt test:scalafmt` simply reformats the code and never fails.

jachro:
Maybe one more thing. Once we're working with the code formatting maybe we should bump the version of the scalafmt plugin as we've got it on renku-graph:
```
addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
```
And in the `scalafmt.conf`
```
version = "2.7.1"
```

jachro (approved):
👍

PR #1553, opened 2020-10-05
feat(ci): always recreate postgres secret for tmp JH deployment

ableuler:
Our ci pipelines often fail after purging the the primary helm release or even deleting and recreating the primary namespace. This is because the (outdated) secret that the secondary Jupyterhub uses to contact postgres is not recreated. This PR fixes this scenario by always recreating the secret.
Closes #1553.


rokroskar (approved):

ableuler:
This currently fails because of RenkuBot not having the permission to delete a secret - I'll wait for https://github.com/SwissDataScienceCenter/renku/pull/1497 to be merged before fixing that because that changes the way we create ci-namespaces anyway.

PR #1562, opened 2020-10-06
chore: remove unused files

rokroskar:


ableuler (approved):
Thanks for bringing out the garbage :)

PR #1573, opened 2020-10-08
chore(ci): add possibility to auto-update PRs

ableuler:
This PR adds the sibling of `automerge`. Set both the `autoupdate` and `automerge` label on an accepted PR and let the 🤖 do the rest.

### DIFF
--- a/.github/workflows/autoupdate.yaml
+++ b/.github/workflows/autoupdate.yaml
@@ -1,0 +1,16 @@
+name: autoupdate
+on:
+  # This will trigger on all pushes to all branches.
+  push: {}
+jobs:
+  autoupdate:
+    name: autoupdate
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: docker://chinthakagodawita/autoupdate-action:v1
+        env:
+          GITHUB_TOKEN: "${{ secrets.RENKU_CI_TOKEN }}"
+          PR_FILTER: "labelled"
+          PR_LABELS: "autoupdate"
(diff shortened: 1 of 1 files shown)


rokroskar:
Why not add the auto-update label on the auto-PRs?

ableuler:
> Why not add the auto-update label on the auto-PRs?

Yeah, good point. Let's merge and see if it does what we want on a PR where we manually add this label. If it behaves as expected we should definitely add this automatically to the auto-update PRs.

rokroskar:
I don't think it does because it should have auto updated this pr already

.github/workflows/autoupdate.yaml
```diff
@@ -0,0 +1,16 @@
...
+  # This will trigger on all pushes to all branches.
+  push: {}
+jobs:
+  autoupdate:
+    name: autoupdate
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: docker://chinthakagodawita/autoupdate-action:v1
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
rokroskar, line 12: should use the renkubot token here

rokroskar (changes requested):

.github/workflows/autoupdate.yaml
ableuler: thx for pointing this out!

rokroskar (approved):

PR #1575, opened 2020-10-09
Deploy renku-ui-server

ciyer:
Deploy the UI server from https://github.com/SwissDataScienceCenter/renku-ui/pull/1043

The UI project needs to be tagged before the renku-ui-server chart is available with the version specified here.


ciyer:
Closed in favor of #2025.

PR #1587, opened 2020-10-13
Combine docs of renku and renku-python

Panaetius:
in conjunction with https://github.com/SwissDataScienceCenter/renku-python/pull/1609

Makes the sidebar shared between the project with cross-links, for a seamless integration. `gensidebar.py` needs to be identical in both (all if we ever add more) repos.

Also takes care of correctly linking between versions of docs for tagged releases.


docs/conf.py
```diff
@@ -214,3 +215,35 @@
 
 # suppress the warning about graph being overriden
 suppress_warnings = ["app.add_directive"]
+
+
+# sidebar
+on_rtd = os.environ.get("READTHEDOCS", None) == "True"
+rtd_version = os.environ.get("READTHEDOCS_VERSION", "latest")
```
rokroskar, line 222: oh I didn't realize you could do this! so you can set the version dynamically based on the RTD build?

rokroskar: if that is the case, why not use this to set version [here](https://github.com/SwissDataScienceCenter/renku/pull/1587/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R81)

Panaetius: Not sure what you mean, that link doesn't point to an actual line of code

rokroskar: ha! you're right. It did before. I mean that we set the `version` towards the top of `conf.py` and it's set by hand. This always leads to the version being wrong because no one ever remembers to change it. If we set it using this RTD environment variable then it's automatically taken care of.

Panaetius: oh I didn't see that. but the version can also be something like `latest` or `stable`, not just the version number

rokroskar: right, in that case it should also just read `latest` probably. It definitely shouldn't read `x.y.z` if it's well past `x.y.z`

rokroskar (approved):

PR #1616, opened 2020-10-23
chore: remove helm from update-component action

jachro:
This PR adds a missing `bitnami` repo required by the `update-component-version` action.

It fixes the following error:
```
helm repo add renku https://swissdatasciencecenter.github.io/helm-charts/
"renku" has been added to your repositories
helm dep update charts/renku
Error: no repository definition for https://charts.bitnami.com/bitnami. Please add them via 'helm repo add'
```


rokroskar:
Actually I guess that `helm` is not needed at all in this action anymore since we are not updating the `requirements.lock` file. I think all of the `helm` bits can be removed...

rokroskar (approved):
🤞

PR #1624, opened 2020-10-27
chore(cd): add manual trigger for CD pipeline on any branch

ableuler:
Adding this _should_ allow us to manually trigger the CD pipeline on the non-default branch. See [the github actions docs](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#workflow_dispatch).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,6 +8,8 @@ on:
     - acceptance-tests/**
     branches: ['master']
     tags: ['*']
+  workflow_dispatch:
+
 
 jobs:
   github-env:


rokroskar:
why do you want to do that?

ableuler:
> why do you want to do that?

Because it will give you a very nice way of triggering the workflow on a non-master branch. This is useful when making changes to the the workflow itself:
image

rokroskar:
won't it still execute whatever matches the `on` section?

ableuler:
Yes, but it's just an additional trigger that you can use ad-hoc instead of adding other branches to the list and (hopefully remember to) remove it again before merging.

rokroskar:
But if the branch is not included in the `on` section, the workflow does not exist on that branch. So you cannot run it on that branch.

ableuler:
I don't think that's the case. As far as I understand it, the `on` section only defines what triggers a workflow and the branch list in the `on` section is underneath `push` and therefore ignored by the manual trigger. I think when triggering manually, the workflow (as defined on master) is executed on the branch selected thought the interface (ie `{{ github.event.ref }}` is set to that branch).

ableuler:
> the workflow does not exist on that branch
Ah, I think I understand what you mean now. Through the UI you would select the workflow first under the actions tab (without specifying the branch) and then select the branch on which to execute the workflow afterwards.

rokroskar (approved):

Issue #1631, opened 2020-10-28
Refactor CI/CD workflows


ableuler:
Currently there is a lot of duplication of code between the "Test and Deploy" workflow deploying master to dev and the "Test and Deploy PR" workflow handling the ci deployments. This could be improved by using common actions as components. Also, we should decouple the release of a renku chart on `tag` from this tag being deployed to our CD deployment. This will give us more freedom in tagging form non-master branches. 

The building blocks for this could be:

- [x] Run acceptance tests through helm (see https://github.com/SwissDataScienceCenter/renku/issues/1762)
- [x] Define a "Deploy RenkuLab" action (#1839)
- [x] Define "Deploy RenkuLab to CD" and "Deploy RenkuLab to CI" action which use the above (#1839 and #1867)
- [x] ~Define "Test Renku" action~  Having`helm test` availble make the running of tests so easy that it doesn't seem to justify a separate action for this anymore. The more complicated handling of test artefacts is being made available through an action (https://github.com/SwissDataScienceCenter/renku/pull/1874). 
- [x] Define a "Publish Chart action"

Then we could have different workflows for PRs, pushes to master and tags which compose those actions in an appropriate way.

rokroskar:
I'd like to add to the wishlist - @Panaetius and I were discussing an action that would allow me to do a comment like

```
/deploy renku-ui==0.11.0
```

in any of our component repos and it would create a renku deployment for the PR using the current renku master as a base + the PR + whatever other versions are specified in the comment. If we do the refactoring you outlined here, I don't think it should be too far of a stretch!

rokroskar:
Afaik we already have the publish chart action? See https://github.com/SwissDataScienceCenter/renku/tree/master/actions/publish-chart

rokroskar:
IMHO the "test renku action" should just require someone to run `helm test` - so I'd like to first see the tests work via helm first. After that the action to trigger that is trivial.

Panaetius:
We can use https://github.com/marketplace/actions/slash-command-dispatch#slash_command-context

I think we should allow specific git refs for individual components to be specified as arguments, as well as just values for the values.yaml, like `/deploy renku-ui=abdfeg123457ad renku-gateway.image.tag=abcdefg12345 jupyterhub.proxy.token=badc0ffee0ddfood`

ableuler:
[...]

PR #1640, opened 2020-10-30
chore(cd): do not deploy to CD on tag

ableuler:
CD should reflect the HEAD of master and not tags from (potentially incompatible) release branches.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,7 +7,6 @@ on:
     - scripts/**
     - acceptance-tests/**
     branches: ['master']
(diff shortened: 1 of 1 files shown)


rokroskar (approved):

rokroskar (changes requested):
Actually I take that back - I think only the deploy and test steps should be skipped on tag, the chart packaging and publishing should still run

ableuler:
Ha, I wasn't going to merge that PR because of the same thinking - see https://github.com/SwissDataScienceCenter/renku/issues/1631.

PR #1683, opened 2020-11-12
chore: push automatic dependency updates to development

rokroskar:
This PR changes the CI/CD pipeline to deploy from the `development` branch. The updates are in three places:

* the github action used by repositories --> this is changed to create the PR based on the development branch
* the automatic PR creation action --> this uses `development` as the base branch
* the CD pipeline is deployed using the `development` branch

The idea is that updates are pushed to `development` and occasionally a merge is made back to `master`, which remains the default branch. All development that is not an automatic PR for a dependency update should still go to the `master` branch. However, merges to `master` from `development` should happen when we are approaching a release and they should be squashed so that the number of automatic commits is minimized in the git history. In principle, this should get us closer to a point where production is updated whenever a merge to `master` is made.

### DIFF
--- a/.github/workflows/auto-reqs-update-pr.yml
+++ b/.github/workflows/auto-reqs-update-pr.yml
@@ -24,4 +24,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.RENKU_CI_TOKEN }}
           BRANCH_PREFIX: "auto-update/"
-          PULL_REQUEST_BRANCH: "master"
+          PULL_REQUEST_BRANCH: "development"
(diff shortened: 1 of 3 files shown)


ableuler:
This looks good. What I think would round off this new setup is a 🤖 that automatically rebases `development` after a PR is merged to master. This would prevent the scenario where a change merged to master does not get deployed on CD for a longer period of time. This would also allow us to assemble all dependency update PRs on-top of `master` for quite some time and then squash them all into one commit.
There are some "auto rebase" github actions, but on a quick look those seem to be designed to run on PRs. I guess we could constantly have a PR from `development` into `master` open, but it would be slightly nicer without that I think.

rokroskar:
I think making a rebase action that runs after merges to master is not a problem. I would want to do this loop a few times by hand first though before making too much automation around it. So I would be in favor of merging this PR, observing how it works for a bit, and then adding an action to automate the rebase of the `development` branch with `master`.

ableuler (approved):

Issue #1003, opened 2020-04-20
Update Keycloak version


pameladelgado:
**Motivation**:
One external Renku deployment needs to update keycloak to a version higher than `8.0.1`, current latest version is `9.0.2`.

~Furthermore, `stable` charts are deprecated and we should update to `codecentric` charts.~
This has been adapted in the meantime.

**Needed change in requirements**
https://github.com/SwissDataScienceCenter/renku/blob/master/charts/renku/requirements.yaml#L31
Current 
```
- name: keycloak
  version: 4.10.1
  repository: "@stable"
  condition: keycloak.enabled
```
desired:
```
- name: keycloak
  version: 7.5.0
  repository: "@codecentric"
  condition: keycloak.enabled
```
We also need to add codecentric to helm `helm repo add codecentric https://codecentric.github.io/helm-charts`

ref: https://github.com/codecentric/helm-charts/blob/master/charts/keycloak/Chart.yaml

**Priority**
Not urgent, but needed in the next few weeks.

Assigning @ableuler since this might need gateway adaptations.

cc @nbarierre

pameladelgado:
I tested a RenkuLab deployment with newer images `8.0.2` and `9.0.3` and it works fine.
For now, deployments that have a stand-alone keycloak can update until `9.0.3` without further modifications.
However, the charts need to be upgraded and they would not be backwards compatible (see https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#upgrading).
``` 
Warning: Merging destination map for chart 'keycloak'. The destination item 'livenessProbe' is a table
 and ignoring the source 'livenessProbe' as it has a non-table value of: httpGet:
  path: {{ if ne .Values.keycloak.basepath "" }}/{{ .Values.keycloak.basepath }}{{ end }}/
  port: http
initialDelaySeconds: 300
timeoutSeconds: 5
Error: render error in "renku/charts/keycloak/templates/statefulset.yaml": template: renku/charts
/keycloak/templates/statefulset.yaml:141:20: executing "renku/charts/keycloak/templates
/statefulset.yaml" at <.>: wrong type for value; expected string; got map[string]interface {}
```

ableuler:
> I tested a RenkuLab deployment with newer images `8.0.2` and `9.0.3` and it works fine.

Thanks for testing this 👍 ! I agree that we should upgrade both chart and image sooner rather than later. See also #1273.

ableuler:
[...]

PR #1687, opened 2020-11-13
Update deprecated set-env

pameladelgado:


charts/renku/requirements.yaml
```diff
@@ -29,6 +29,6 @@ dependencies:
   repository: "https://charts.bitnami.com/bitnami"
   condition: postgresql.enabled
 - name: keycloak
-  version: 4.10.1
-  repository: "@stable"
+  version: 4.10.2
```
rokroskar, line 32: why the change in this PR?

.github/workflows/pull-request-test.yml
```diff
@@ -94,6 +94,7 @@ jobs:
       run: |
         echo "$RENKUBOT_KUBECONFIG" > renkubot-kube.config
         helm repo add stable 	https://kubernetes-charts.storage.googleapis.com
+        helm repo add codecentric https://codecentric.github.io/helm-charts
```
rokroskar, line 97: This should be added to instructions that describe deploying renku from source - but I'm not sure it belongs in this PR?

rokroskar (changes requested):
Thanks @pameladelgado! I don't think we should be making changes to the keycloak version here - if it's inevitable then I suggest we do the change here but open an issue for adapting the docs.

charts/renku/requirements.yaml
pameladelgado: > why the change in this PR?

Because in the new codecentric repo that's the oldest chart available. The deployment does not work otherwise, so it is indeed a small inevitable update. Luckily, this is just a bugfix change in versions so it should not break things. Authorization seem to work just fine in that deployment.
It seems that we might want to work on #1003 sooner rather than later.

.github/workflows/pull-request-test.yml
pameladelgado: Yes, documentation should be updated.
I also did not included it here because I was not sure it belongs in this PR but I can certainly add it. Bitnami repo should also be added to the docs.

charts/renku/requirements.yaml
lorenzo-cavazzi: I was going to push the very same changes cause I had the issue with `keycloack` :smile: 

Some extra references for the reviewers:
* Deprecated: https://github.com/helm/charts
* New source: https://github.com/codecentric/helm-charts
* Problem with the version:
```
helm search repo keycloak --versions
NAME            CHART VERSION   APP VERSION     DESCRIPTION                                       
stable/keycloak 4.10.1          5.0.0           DEPRECATED - Open Source Identity and Access Ma...
stable/keycloak 4.10.0          5.0.0           Open Source Identity and Access Management For ...
```

.github/workflows/pull-request-test.yml
rokroskar: sure that's fine - please make an issue for it. Thanks!

rokroskar (approved):

PR #1708, opened 2020-11-19
Fix acceptance tests

lorenzo-cavazzi:
Update to support UI changes SwissDataScienceCenter/renku-ui#1123


rokroskar:
I guess 38c9ce7 is a leftover from starting from master?

lorenzo-cavazzi:
ah yes. Should I remove it? It will make its way into `development` later anyway

rokroskar:
yes, it will, but I'm not sure if it will cause an unnecessary merge conflict

lorenzo-cavazzi:
I can remove it just to be safe :smile:

lorenzo-cavazzi:
Alright, a good old `--force` should have fixed it.

rokroskar (approved):

PR #1715, opened 2020-11-19
chore(ci): increase automerge timeout to 1 day

ableuler:
Keep trying to automerge a PR every 5 minutes for 24 hours. The current setting of two hours seems to be too short or our tests sometimes.


ableuler:
This wasn't the reason for the automerge to not work properly.

PR #1737, opened 2020-11-25
docs: update interactive environments documentation

### LINKED ISSUE #1647: Add documentation for pinning a registry image
After we support custom `image` for new interactive environments in the UI, we should document how to use the feature.

We will need to update at least the "Interactive Environment" documentation pages.
An idea would be to also update our templates to include a few lines in the Readme file to help people use this feature and make them aware of its existence.
In that case, the change for the documentation should extend to the "Templates" pages, and it would be great to include in the `.gitlab-ci.yml` a few lines of commented code to enable the creation of images only in specific cases (e.g. on tags)

**Describe alternatives you've considered**
This issue is only a stub -- feel free to contribute with different ideas or to point out other places where the change would be required.

lorenzo-cavazzi:
Update the Interactive Environments documentation pages, including the feature to pin images (that requires this to the merged SwissDataScienceCenter/renku-ui#1109).

fix #1647

### DIFF
--- a/docs/user/interactive_basics.rst
+++ b/docs/user/interactive_basics.rst
@@ -43,11 +43,12 @@ What's in my Interactive Environment?
 
 * Any dependencies you specified via conda (``environment.yml``), using
   language-specific dependency-management facilities (``requirements.txt``,
-  ``install.R``, etc.) or installed in the ``Dockerfile``.
+  ``install.R``, etc.) or installed in the ``Dockerfile``. An exception to
+  this is if :ref:`project sets a specific image <renku_ini>`.
 
 * The renku command-line interface renku-python_.
 
-* The amount of CPUs, memory, and (possibly) GPUs that you configured before launch
+* The amount of CPUs, memory, and (possibly) GPUs that you configured before launch.
 
 For adding or changing software installed into your project's interactive environment,
 check out :ref:`customizing`
--- a/docs/user/interactive_basics.rst
+++ b/docs/user/interactive_basics.rst
@@ -20,7 +20,8 @@ but RenkuLab offers the following advantages:
 
 * Environments are defined using Docker, so they can be shared and reproducibly re-created.
 
-* Auto-saving of work back to RenkuLab, so you can recover in the event of a crash.
+* Auto-saving of work back to RenkuLab, so you can recover when your environment is shut down
+  (this happens automatically after 24 hours of inactivity).
 
 * A git client pre-configured with your credentials to easily push your changes
   back to the server.
@@ -122,9 +123,10 @@ A GitLab's CI/CD pipeline automatically builds a new image using the project's
   * Forking a project (in which the new build happens for the fork).
   * Pushing changes to the project.
 
-This is defined in the project's :ref:`.gitlab-ci.yml file <gitlab_ci_yml>`. If the project
-references a pinned image, the UI will not check for the image availability - that is
-usually provided by the project's maintainer and it doesn't change at every new commit.
+The pipeline is defined in the project's :ref:`.gitlab-ci.yml file <gitlab_ci_yml>`. If the
+project :ref:`references a specific image <renku_ini>` to use for all environments, the UI
+will not check for the image availability - that is usually provided by the project's
+maintainer and it doesn't change at every new commit.
 
 It may take a long time to build an image for various reasons, but if you've just created the
 project on RenkuLab from one of the templates, it generally takes less than a minute or two.
--- a/docs/user/interactive_basics.rst
+++ b/docs/user/interactive_basics.rst
@@ -15,40 +15,38 @@ of code before combining everything into a (reproducible) workflow.
 You can run JupyterLab or RStudio within a project independently from RenkuLab,
 but RenkuLab offers the following advantages:
 
-* environments hosted in the cloud with a configurable amount of resources
-  (memory, CPU, and sometimes GPU)
+* Environments hosted in the cloud with a configurable amount of resources
+  (memory, CPU, and sometimes GPU).
 
-* environments are defined using Docker, so they can be shared and reproducibly
-  re-created
+* Environments are defined using Docker, so they can be shared and reproducibly re-created.
 
-* auto-saving of work back to RenkuLab, so you can recover in the event of a
-  crash
+* Auto-saving of work back to RenkuLab, so you can recover in the event of a crash.
 
-* a git client pre-configured with your credentials to easily push your changes
-  back to the server
+* A git client pre-configured with your credentials to easily push your changes
+  back to the server.
 
-* the functionality provided by the renku-python_ command-line interface (CLI)
-  is automatically available
+* The functionality provided by the renku-python_ command-line interface (CLI)
+  is automatically available.
 
 
 What's in my Interactive Environment?
 -------------------------------------
 
-* your project, which is cloned into the environment on startup
+* Your project, which is cloned into the environment on startup.
 
-* your data (if the option ``Automatically fetch LFS data`` is selected)
-  files that are stored in git LFS*)
+* Your data files (if the option ``Automatically fetch LFS data`` is selected)
+  that are stored in git LFS*.
 
-* all the software required to launch the environment and common tools for
-  working with code (``git``, ``git LFS``, ``vim``, etc.)
+* All the software required to launch the environment and common tools for
+  working with code (``git``, ``git LFS``, ``vim``, etc.).
 
-* any dependencies you specified via conda (requirements.txt), using
+* Any dependencies you specified via conda (``environment.yml``), using
   language-specific dependency-management facilities (``requirements.txt``,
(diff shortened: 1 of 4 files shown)


docs/user/interactive_customizing.rst
```diff
@@ -164,6 +174,58 @@ these base ``Dockerfile`` s and add the ``renku``, ``git``, and ``jupyter``
...
+configurations stored in the ``.renku/renku.ini`` file.
+Once you do that, each time a user tries to start a new environment, those options will
+be pre-selected.
+
+.. note::
+
+  Manually modifying the ``renku.ini`` file is not recommended.
+  You can use the
+  `renku config command <https://renku-python.readthedocs.io/en/latest/commands.html#module-renku.cli.config>`_
+  form an interactive environment.
```
rokroskar, line 198:
```suggestion
  from an interactive environment, for example:
```

docs/user/interactive_customizing.rst
```diff
@@ -164,6 +174,58 @@ these base ``Dockerfile`` s and add the ``renku``, ``git``, and ``jupyter``
...
+and you probably want to choose it as default.
+
+It's possible to set a default value for all these options using the project
+configurations stored in the ``.renku/renku.ini`` file.
+Once you do that, each time a user tries to start a new environment, those options will
+be pre-selected.
+
+.. note::
+
+  Manually modifying the ``renku.ini`` file is not recommended.
```
rokroskar, line 195: Oh really? That's the only way I ever do it 😄

docs/user/interactive_customizing.rst
```diff
@@ -164,6 +174,58 @@ these base ``Dockerfile`` s and add the ``renku``, ``git``, and ``jupyter``
...
+  You can use the
+  `renku config command <https://renku-python.readthedocs.io/en/latest/commands.html#module-renku.cli.config>`_
+  form an interactive environment.
+
+    renku config set interactive.default_url "/tree"
+
+  We are working on adding a user friendly solution to set default options on
```
rokroskar, line 202:
```suggestion
  We are working on adding a user-friendly solution to set default options on
```

docs/user/interactive_customizing.rst
```diff
@@ -164,6 +174,58 @@ these base ``Dockerfile`` s and add the ``renku``, ``git``, and ``jupyter``
 parts to another base image that you might have.
 
 
+Renku project configurations
```
rokroskar, line 177: Maybe we can move this section before the Docker section? I imagine people might want to change this more often than they would want to dig into the Docker build itself.

rokroskar (changes requested):
This is great @lorenzo-cavazzi thanks!! Just a few minor comments. My future you (hunting for `renku.ini` options all over the docs) thanks you in advance 🙇

docs/user/interactive_basics.rst
```diff
@@ -15,40 +15,38 @@ of code before combining everything into a (reproducible) workflow.
...
+* Environments hosted in the cloud with a configurable amount of resources
+  (memory, CPU, and sometimes GPU).
 
-* environments are defined using Docker, so they can be shared and reproducibly
-  re-created
+* Environments are defined using Docker, so they can be shared and reproducibly re-created.
 
-* auto-saving of work back to RenkuLab, so you can recover in the event of a
-  crash
+* Auto-saving of work back to RenkuLab, so you can recover in the event of a crash.
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Good point. We should mention a more generic "environment shutdown". We can also mention that environments are shut down automatically after 24h of inactivity

docs/user/interactive_customizing.rst
lorenzo-cavazzi: ahah it may work for you, but I would say it's safer to use the `renku` command line to avoid messing up the config file :grin:

docs/user/interactive_customizing.rst
lorenzo-cavazzi: :+1: 
I'll fix it also on the other locations where we use the space

lorenzo-cavazzi:
All the point should be addressed

docs/user/interactive_customizing.rst
rokroskar: I always just assume the UI is buggy and move on 😆

rokroskar (approved):

docs/user/interactive_basics.rst
```diff
@@ -15,40 +15,39 @@ of code before combining everything into a (reproducible) workflow.
...
-* all the software required to launch the environment and common tools for
-  working with code (``git``, ``git LFS``, ``vim``, etc.)
+* All the software required to launch the environment and common tools for
+  working with code (``git``, ``git LFS``, ``vim``, etc.).
 
-* any dependencies you specified via conda (requirements.txt), using
+* Any dependencies you specified via conda (``environment.yml``), using
   language-specific dependency-management facilities (``requirements.txt``,
-  ``install.R``, etc.) or installed in the ``Dockerfile``
+  ``install.R``, etc.) or installed in the ``Dockerfile``.
```
ciyer, line 46: We should add to this, I think.

```
* Any dependencies you specified via conda (``environment.yml``), using language-specific dependency-management facilities (``requirements.txt``, ``install.R``, etc.) or installed in the ``Dockerfile``. (An exception to this is if the project uses a pinned image. [See the section below](linked to pinned image section) for more information.)
```

docs/user/interactive_customizing.rst
```diff
@@ -14,55 +14,118 @@ Important files
...
+Jupyter interface ``/tree``, or when you need support for a different interface,
+like R studio ``/rstudio`` or  ``/streamlit`` (not included in the standard Python template).
+
+The ``image`` is useful when you settle on a Docker image and you don't need to change it
+anymore. The benefit is particularly evident when building a new image takes a lot of time
+(e.g. you added big packages) or when you expect the project to be used by a lot of people
+over a short period of time (e.g. you use it in a presentation or a lecture).
+
+Even if it's common to start the environment with the default values, setting a default value
+doesn't prevent a user from changing it.
```
ciyer, line 77: I do not think a user can change the pinned image, or am I wrong?

ciyer (changes requested):
Lots of good improvements here. I think we need to be a bit clearer about the pinned image feature.

docs/user/interactive_customizing.rst
lorenzo-cavazzi: This is true in the current UI implementation, but I'm not sure if we want to change that or not.
The section is about all the options, but you are right, `image` is an exception. If you think it's better to specify it, I can add the text

docs/user/interactive_basics.rst
rokroskar: @ciyer I thought we should try to avoid using the word "pinned" - I don't think it's really a term that most of our users understand immediately. But maybe it's more widely used than I think?

docs/user/interactive_customizing.rst
ciyer: I think we should describe the current UI behavior in the documentation. If that changes, then we should update the documentation.

docs/user/interactive_basics.rst
ciyer: @rokroskar I wasn't involved with those discussions, so I don't know what arguments were presented. 

I just do not find "custom" very descriptive in this case. There image isn't any more or less customized than any other image used in an interactive environment -- what is unique in this case is that the image if pinned/fixed and may not reflect the contents of the project (Dockerfile, environment.yml, etc.).

In fact, the project that was created for testing was called "fixed-image" -- that's an indication to me that there might be a naming mismatch.

lorenzo-cavazzi:
I addressed the 2 points from the last review, but I didn't use the term `pinned image`.
@rokroskar @ciyer if we want to use a specific term (fixed/pinned/custom/<other> image) it's better to settle on a single one and keep it coherent both here in the docs and in the UI.

ciyer:
> I addressed the 2 points from the last review, but I didn't use the term `pinned image`.
> @rokroskar @ciyer if we want to use a specific term (fixed/pinned/custom/ image) it's better to settle on a single one and keep it coherent both here in the docs and in the UI.

I have argued my point. In the end, it is @rokroskar 's decision.

docs/user/interactive_basics.rst
rokroskar: @ciyer there wasn't really a discussion, it was just a suggestion I made [here](https://github.com/SwissDataScienceCenter/renku/pull/1737#discussion_r530839218) - I'm fine with using "pinned" but I thought we could maybe come up with something a bit less technical. If there isn't a better term then pinned is also OK. 

It's custom in the sense that it's not the one automatically provided by the system. It might be one of our standard images, one that was built by a previous CI run, or something that the user created completely on their own.

rokroskar:
@lorenzo-cavazzi I think go with "pinned" and if people respond with bewilderment we can revise :)

ciyer (approved):
👍

PR #1751, opened 2020-11-30
feat: add containerized renku deploy action

### LINKED ISSUE #377: Run integration tests renku ui/notebooks/gateway/...
Most of our recent test failures on Renku have been related to merged PRs in some of the other Renku repositories. Most of the times it is not immediately clear which PR caused the failure which leads time being wasted for debugging. 
Being able to run the the tests also on other repos will save us a lot of time. I think we have to find a solution for this, even if it means moving away from travis. Is GitLab CI/CD the way to go? (https://about.gitlab.com/features/github/)

rokroskar:
Adds two containerized github actions:

1. deploy renku given a collection of different versions
2. teardown a renku deployment

The first does almost everything needed to bring up a renku deployment with minimal boilerplate. (for the moment) you can do something like this:

```yaml
deploy-pr:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: renku build and deploy
      uses: ./actions/deploy-renku
      env:
        RANCHER_PROJECT_ID: ${{ secrets.CI_PROJECT_ID }}
        DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
        DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
        GITLAB_TOKEN: ${{ secrets.GITLAB_TOKEN }}
        KUBECONFIG: "${{ github.workspace }}/renkubot-kube.config"
        RENKU_NAMESPACE: ci-${{ github.event.number }}-renku
        RENKU_TMP_NAMESPACE: ci-${{ github.event.number }}-renku-tmp
        RENKU_RELEASE: ci-${{ github.event.number }}-renku
        RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
        RENKUBOT_RANCHER_APISECRET: ${{ secrets.RENKUBOT_RANCHER_APISECRET }}
        renku: "@${{ github.head_ref }}"
```

This will deploy the current PR version of renku and is used in this PR to replace the existing PR deploy action. 

The second is the teardown action which can be used very simply like this:

```yaml
cleanup:
    if: github.event.action == 'closed'
    runs-on: ubuntu-20.04
    steps:
    - uses: actions/checkout@v2
    - name: renku teardown
      uses: ./actions/teardown-renku
      env:
        KUBECONFIG: "${{ github.workspace }}/renkubot-kube.config"
        RENKU_NAMESPACE: ci-renku-${{ github.event.number }}
        RENKU_RELEASE: ci-renku-${{ github.event.number }}
        RENKU_TMP_NAMESPACE: ci-renku-${{ github.event.number }}-tmp
        RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
```

There is probably room in both of these actions to reduce the number of environment variables used... but I kept everything extra explicit for the time being. 

Addresses #1631 
closes #377 

/deploy


rokroskar:
So this basically works. The questions I have are:

1. do we want to bundle all of the additional setup in this action? I'm thinking here of e.g. the rancher project / namespace configuration and the gitlab application registration. If not inside this action then where? 
2. Where should the additional dependency constraints be specified? I thought we could do it via a comment but it turns out there is no action event trigger for that. One idea I had was to include the information about special dependencies in the description of the PR - we can then parse that in the action and extract the versions as needed.

actions/deploy-renku/entrypoint.sh
```diff
@@ -0,0 +1,54 @@
...
+# create the namespace in a Rancher project
+curl -s -H "Authorization: Bearer $RENKUBOT_RANCHER_BEARER_TOKEN" \
+        -X POST \
+        -d "name=${RENKU_NAMESPACE}" \
+        -d "projectId=${RANCHER_PROJECT_ID}" \
+        'https://rancher.renku.ch/v3/cluster/c-l6jt4/namespaces'
+
+curl -s -H "Authorization: Bearer $RENKUBOT_RANCHER_BEARER_TOKEN" \
+        -X POST \
+        -d "name=${RENKU_TMP_NAMESPACE}" \
```
aledegano, line 41: Where is this namespace used? I can't find...

rokroskar: Yes this is a bit obscured... this is used for the separate deployment of a jupyterhub + notebooks service to support anonymous user sessions. It's set up with the script executed [here](https://github.com/SwissDataScienceCenter/renku/pull/1751/files#diff-363792bff6f664b9158bd2432994427d2aeaf8e521c9bfa2c2bfb1452de10cbeR54)

rokroskar: ahaha no you're right - it's never passed to that script 👍  I thought it was 🤦

rokroskar: It just happens to be formed here the same way as the default value in the script, but it should be passed explicitly to avoid surprises. I'll do that.

aledegano: :+1:

rokroskar: I resolved this in 534edc8

ableuler:
This is great! I have one question: would it work to move the scripts and the Dockerfile out of the `/actions` directory into something like `charts/utils` and just point the action to the right Dockerfile? First, I think these scripts are very useful also outside the context of a github action, and second, when implementing the anonymous sessions, I made some assumptions about the location of `deploy_tmp_notebooks.py`. See for example in the script itself `chartdir = os.path.abspath(os.path.dirname(__file__))` or in the NOTES.txt chart template (might have taken some shortcuts there).

ableuler:
> So this basically works. The questions I have are:
> 
> 1. do we want to bundle all of the additional setup in this action? I'm thinking here of e.g. the rancher project / namespace configuration and the gitlab application registration. If not inside this action then where?

Can one action have several steps with different Dockerfiles? If yes we can put it into the same action, if not I would create a separate action for it. 

> 2. Where should the additional dependency constraints be specified? I thought we could do it via a comment but it turns out there is no action event trigger for that. One idea I had was to include the information about special dependencies in the description of the PR - we can then parse that in the action and extract the versions as needed.

Not sure I understand this question correctly. I guess you are talking about the PRs in the component repos and not the ones updating the dependencies in the renku chart. But what exact scenario do you have in mind here?

rokroskar:
Yes I wanted to put the scripts in some more generic location but I can't see an obvious way to do that. What I could do is link to them from an e.g. `charts/utils` directory. Also, I agree that there is a ton of setup done in one action here - my take on this was that I wanted a quick way to distribute the acceptance tests all across our repos. So this is super specific for our particular setup. If at some point we want to generalize these, we can do that, but I don't know of anyone else running renku accceptance tests atm. 

> Can one action have several steps with different Dockerfiles? If yes we can put it into the same action, if not I would create a separate action for it.

You can do this via action templates - see [here](https://github.com/SwissDataScienceCenter/.github/pull/1). But still, I think having it all in one action is pretty straightforward, there really isn't that many steps involved. I'd rather split it in more logical sections like deploy - test - teardown - the rancher stuff just happens to be a part of the "deploy" bit.

For the second question - yes, what I was thinking of was the frequent case where one component's PR depends on some branch of another component for testing, feature development etc. I think we can do it via PR comments, but that can be a different PR.

ableuler:
> Yes I wanted to put the scripts in some more generic location but I can't see an obvious way to do that. 

According to the [actions docs](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/metadata-syntax-for-github-actions#runsimage) it should be possible to point to any Dockerfile in the repo from `action.yaml`, so it seems to me that the Dockerfile, the entrypoint and the two scripts could live elsewhere.


> But still, I think having it all in one action is pretty straightforward, there really isn't that many steps involved. I'd rather split it in more logical sections like deploy - test - teardown - the rancher stuff just happens to be a part of the "deploy" bit.

My first reflex would be to keep the actions small and compose them into workflows rather than having one action that does everything, but it's true that a `deploy-renku` action does not make sense if for example Rancher is not configured properly, so if you think it's simpler having it all in one action, this would also be ok for me.

> For the second question - yes, what I was thinking of was the frequent case where one component's PR depends on some branch of another component for testing, feature development etc. I think we can do it via PR comments, but that can be a different PR.

Ok, I see what you mean. I'm not even sure we have to provide any automation for this case. For manual testing and development one can just specify several `@branchname` component versions. For `ci-XYZ` deployments of PRs in the component repos it's ok to require that they pass the tests with all other components being the same as on the tip of the renku development branch. If a component PR depends on a code change in another repo, this PR has to be merged and a release of that other component has to be tagged first. For me this is very much in the spirit of the phased rollouts that we want to adopt.

rokroskar:
> According to the actions docs it should be possible to point to any Dockerfile in the repo from action.yaml, so it seems to me that the Dockerfile, the entrypoint and the two scripts could live elsewhere.

"should be" is the operative word here - it doesn't work. It seems if you specify a relative path it thinks it's a docker registry and tries to pass it to `docker run` so the action fails. This seems to be a bug in github actions... So I've changed things back now and I added symbolic links to those scripts in the `helm-chart/utils` directory. I agree it's not very nice like this, but I'm not really sure if there is another option.

Panaetius:
> > According to the actions docs it should be possible to point to any Dockerfile in the repo from action.yaml, so it seems to me that the Dockerfile, the entrypoint and the two scripts could live elsewhere.
> 
> "should be" is the operative word here - it doesn't work. It seems if you specify a relative path it thinks it's a docker registry and tries to pass it to `docker run` so the action fails. This seems to be a bug in github actions... So I've changed things back now and I added symbolic links to those scripts in the `helm-chart/utils` directory. I agree it's not very nice like this, but I'm not really sure if there is another option.

Did you try using a qualified relative path like `file://../../Dockerfile` ? Maybe that works.

Also the path has to be relative to the action file, not the root of the repository, but you probably tried that already.

rokroskar:
Thanks for the suggestion @Panaetius - I tried all those options, see [here](https://github.com/SwissDataScienceCenter/renku/commit/e0abe41461bbd8435e1197b468a1c0112d8f9049#diff-767a4d1a14b40c37f59146c4ced2601cfbd4b4566a8527c8ec8114f324171422R5) and [here](https://github.com/SwissDataScienceCenter/renku/commit/1b196ddf0e5972a8a291afaa7f7462f80a4851f6#diff-767a4d1a14b40c37f59146c4ced2601cfbd4b4566a8527c8ec8114f324171422R5). It either seems like a bug or faulty documentation.

aledegano (dismissed):

lorenzo-cavazzi:
This currently seems to fail. Here is where I'm using the action: https://github.com/SwissDataScienceCenter/renku-ui/pull/1179

pameladelgado:
> This currently seems to fail. Here is where I'm using the action: [SwissDataScienceCenter/renku-ui#1179](https://github.com/SwissDataScienceCenter/renku-ui/pull/1179)

@lorenzo-cavazzi this could be because of the deprecated stable URL. There's a fix for this in #1814

rokroskar:
> This currently seems to fail. Here is where I'm using the action: [SwissDataScienceCenter/renku-ui#1179](https://github.com/SwissDataScienceCenter/renku-ui/pull/1179)

@lorenzo-cavazzi I left a comment in your PR - I renamed the helm chart directory in this PR, that's why it fails I believe.

actions/deploy-renku/entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# deploy renku - reads config from environment variables
+python3 /deploy-dev-renku.py
+
+# deploy anonymous notebooks
+helm repo add renku https://swissdatasciencecenter.github.io/helm-charts
+helm repo update
+
+# fetch the deployed renku-notebooks version
+NOTEBOOKS_VERSION=$(kubectl -n "$RENKU_NAMESPACE" get configmaps notebook-helper-scripts -ojson | jq -r ".metadata.labels.chart")
```
ableuler, line 53: This is much better than what I did in `/deploy-tmp-notebooks.py` (simply reading from `requirements.yaml`). Maybe you can just replace this in `/deploy-tmp-notebooks.py` and remove it here?

actions/deploy-renku/entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+python3 /deploy-tmp-notebooks.py --release-name "$RENKU_RELEASE" \
```
ableuler, line 54: Will this work for an unpublished notebook chart (ie for a PR on the notebooks service)? I _think_ not, but I'm not 100% sure right now. Publishing the chart when pushing the images would be the easy way out...

actions/deploy-renku/entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+        -X POST \
+        -d "name=${RENKU_TMP_NAMESPACE}" \
+        -d "projectId=${RANCHER_PROJECT_ID}" \
+        'https://rancher.renku.ch/v3/cluster/c-l6jt4/namespaces'
+
+# deploy renku - reads config from environment variables
+python3 /deploy-dev-renku.py
+
+# deploy anonymous notebooks
+helm repo add renku https://swissdatasciencecenter.github.io/helm-charts
```
ableuler, line 49: I think these 2 lines should not be necessary - unless we publish a chart in the above steps (see comment below).

ableuler (changes requested):
This looks good! Using symbolic links is a good workaround.

actions/deploy-renku/entrypoint.sh
rokroskar: I think you're right. I don't like the solution of pushing the chart on every PR commit though...

ableuler: Then the only other solution I see is deploying the anonymous notebooks deployment also from the local chart. How about adding an optional `chart-path` argument to `/deploy-tmp-notebooks.py` that - if provided - would replace the chart repo and version parameters in the resulting helm command?

rokroskar: To do that you need the logic from the other script, actually (clone, checkout, chartpress, etc). I think refactoring the `deploy-tmp-notebooks` script a bit and just calling it from the other one is the way to go.

rokroskar:
closing to restart the CI

Issue #1762, opened 2020-12-03
Run acceptance tests through helm


ableuler:
The acceptance tests are currently wired together through docker-compose. Running them therefore requires many environment variables to be defined. If the tests were run through helm (`helm test`), the necessary information or running the tests could just be taken from the chart values.

The steps here would be:
- creating a test user
- running the acceptance tests using this user
- deleting the test user (also in gitlab) after successful test execution

Also, for dev or ci deployments there should be an option to run the tests using another deployment (and an existing user from there) as identity provider.

aledegano:
Regarding the automated user creation and deletion I believe that we should have a discussion because there are several moving parts involved in that.
To start these are some facts, as I understand them:

- When registering an user for the first time, through the Renku UI an email verification is required. The acceptance tests do not handle this.

- Once the user is registered on Renku (Keycloak), Renku must be allowed as an application of Gitlab to enable OIDC authentication 

- As per renku#1804, currently removing an account requires multiple steps across different components

- When Renku is configured to use another deployment as Identity Provider (as for personal dev deployments), the registration of the user is automated in the tests, if the user already exists in the upstream provider (however there cannot be automatically created)

Given the above I only see two options:

1. Handle the test user manually, out-of-band with respect to the tests, documenting how a Renku operator is supposed to set it up
2. Create issues to tackle the above points, distributing them on the competent Renku sub-teams to get to a fully automatable user creation/destruction within the tests

Personally I lean toward the option one as I am not convinced that option 2 brings enough value for the effort I am expecting.

@SwissDataScienceCenter/yat what do you guys think?

PR #1780, opened 2020-12-09
feat: run acceptance tests through helm test

aledegano:
* Refactor container for acceptance test to include Selenium/chromium-driver

 * Modify acceptance tests spec to use chromium-driver within the same container

 * Add manifest for `helm test` to use the refactored container in a pod

 * Update values.yaml to request the necessary parameters to run the tests

 * Update documentation

Part of #1762.

### DIFF
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -16,37 +16,71 @@ normally of the form `test_[timestamp]`. This makes it easy to identify the test
 name collisions. If the test runs successfully to completion, the project is deleted at the end of the test.
 If the test fails, then the project is left around to aid debugging.
 
+## Running on Docker
 
-## Running over Docker
+The easiest way to run the tests is to use docker/docker-compose.
 
-The easiest way to run the tests is to use docker/docker-compose. The docker-compose environment is made up of
-two containers: one hosting selenium, and one running the tests.
+### Installation
 
+The `docker-compose` command has to be available: it is easiest to install it with `pip`.
 
-#### Installation
-The `docker-compose` command has to be available. For installation steps see: https://docs.docker.com/compose/install/.
+### Running
 
-#### Running
-For running with docker, the parameters to the test need to be provided as environment variables:
+To run through docker-compose, the tests parameters need to be provided as environment variables:
 
 ```
-docker-compose run -e RENKU_TEST_URL=https://renku-kuba.dev.renku.ch -e RENKU_TEST_FULL_NAME="<full user name>" -e RENKU_TEST_EMAIL=<email> -e RENKU_TEST_USERNAME=<username> -e RENKU_TEST_PASSWORD=<password> sbt
+docker-compose run \
+-e RENKU_TEST_URL=https://renku-kuba.dev.renku.ch \
+-e RENKU_TEST_FULL_NAME="<full user name>" \
+-e RENKU_TEST_EMAIL=<email> \
+-e RENKU_TEST_USERNAME=<username> \
+-e RENKU_TEST_PASSWORD=<password> \
+sbt
 ```
 
 By default, all tests are run. You can provide an argument to run a specific test:
-
 ```
-docker-compose run -e RENKU_TEST_URL=https://renku-kuba.dev.renku.ch -e RENKU_TEST_FULL_NAME="<full user name>"
--e RENKU_TEST_EMAIL=<email> -e RENKU_TEST_USERNAME=<username> -e RENKU_TEST_PASSWORD=<password> sbt
-/tests/docker-run-tests.sh *<test-class-name>*
+docker-compose run \
+-e RENKU_TEST_URL=https://renku-kuba.dev.renku.ch \
+-e RENKU_TEST_FULL_NAME="<full user name>" \
+-e RENKU_TEST_EMAIL=<email> \
+-e RENKU_TEST_USERNAME=<username> \
+-e RENKU_TEST_PASSWORD=<password> \
+sbt *<test-class-name>*
 ```
 __**IMPORTANT**__
 
 Once you do changes to the `Dockerfile` do not forget to build the image with `docker-compose build`.
 
+## Running through Helm test
+
+The `helm test` command can also be used to execute the tests against a Helm release of Renku.
+
+When testing a Renku release that relies on an external GitLab instance, the test user must already
+have an existing account on that GitLab instance with login credentials specified in the
+values.yaml file. Note that the same user (with the same username/password combination!) must also exist in the Keycloak instance which is part of the Renku release being tested. Alternatively, if the `.Values.tests.registerTestUser` is set to `true`, the test suit will try to register a new user with the provided details.
+```yaml
+tests:
+  ## User account for running `helm test`
+  testUser:
+   username: foo.bar
+   fullname: Foo Bar
+   email: foo.bar@example.com
+   password: FooBar
+```
+The tests can now be executed:
+```bash
+helm --namespace renku-foo test renku-foo --logs
+```
+
+The acceptance tests will run on the Kubernetes cluster where `renku-foo` is deployed, in the
+same namespace, in a pod named `renku-foo-acceptance-tests`. Once the tests are completed
+(successfully or otherwise) the container is terminated.
+
 ## Running using sbt
 
-#### Installation
+### Installation
+
 In order to run the tests you need to install:
 * Java Development Kit (JDK) >= 8 (see: https://adoptopenjdk.net/?variant=openjdk11&jvmVariant=openj9)
 * sbt >= 1.3.3 (use `sbt sbtVersion` to verify current version but run it not from the project root)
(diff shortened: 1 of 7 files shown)


acceptance-tests/src/test/scala/ch/renku/acceptancetests/tooling/AcceptanceSpec.scala
```diff
@@ -29,10 +28,11 @@ trait AcceptanceSpec
 
   private def getWebDriver: WebDriver =
     sys.env.get("DOCKER") match {
-      // NOTE We could also support running against the selenium container from the host machine:
-      // RemoteWebDriver.builder().url("http://localhost:4444/wd/hub").addAlternative(new ChromeOptions()).build()
       case Some(_) =>
-        RemoteWebDriver.builder().url("http://chrome:4444/wd/hub").addAlternative(new ChromeOptions()).build()
+        new ChromeDriver(
```
aledegano, line 32: Thank you @jachro :pray:

acceptance-tests/README.md
```diff
@@ -16,37 +16,71 @@ normally of the form `test_[timestamp]`. This makes it easy to identify the test
...
 __**IMPORTANT**__
 
 Once you do changes to the `Dockerfile` do not forget to build the image with `docker-compose build`.
 
+## Running through Helm test
+
+The `helm test` command can also be used to execute the tests against a Helm release of Renku.
+
+Suppose to have a deployed release of Renku called `renku-foo`, the `values.yml` file should
+contain the credential to an user that needs to be already present on `dev.renku.ch`
```
ableuler, line 60: Must that user already be present in keycloak of `renku-foo` too or is it automatically created there? In general I think it's good practice to assume that any person who is not an SDSC dev should be able to read and follow this documentation, so I'd not assume that everybody reading this docs runs their tests against an instance that uses `dev.renku.ch/gitlab` as GitLab backend. We could write something like "When testing a Renku release that relies on an external GitLab instance, the test user must already have an existing account on that GitLab instance with login credentials specified in the values.yaml file."

My preferred end state would be:
- For regular deployments that do not depend on another gitlab instance, the test user is created in the beginning through the keycloak API and (if tests are successful) destroyed at the very end together with all the projects that were created during tests.
- For a deployment which does rely on another gitlab instance, the GitLab credentials of and existing user for that instance must be provided through the values.yaml file.

This would allow us to get rid of the same user with the same password having to be present in two separate locations.

rokroskar:
Actually the user must already exist also in our instances because you need to be able to push to gitlab - for that you either need to create a password or create a private token because making tokens via the API is not possible.

ableuler:
> Actually the user must already exist also in our instances because you need to be able to push to gitlab - for that you either need to create a password or create a private token because making tokens via the API is not possible.

Good point, forgot about this. Could obtaining a token through the GitLab UI be part of the browser test flow?

rokroskar:
I don't see why not, though also not sure what you're testing in that case :) How does the clipboard work in selenium?

ableuler:
> I don't see why not, though also not sure what you're testing in that case :) How does the clipboard work in selenium?

I'd probably skip the clipboard and just read the password from the page into the scala code which can reuse it later.

charts/renku/templates/tests/test-renku.yaml
```diff
@@ -1,51 +1,42 @@
 apiVersion: v1
 kind: Pod
```
ableuler, line 2: Would we gain anything in terms of persisting error-state, etc from using a k8s job instead of a naked pod here? I actually don't think so but I'm not sure.

ableuler (changes requested):
Some smaller things to be changed and one question that came up.

charts/renku/values.yaml
```diff
@@ -34,6 +34,9 @@ global:
     password:
       value:
       overwriteOnHelmUpgrade: false
+    ## Skip Keycloak testing when running Helm test
```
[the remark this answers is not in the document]
aledegano: Yes, you're right, I've actually realized this when working on the next PR, but have not backported it here. I will.

charts/renku/templates/tests/test-renku.yaml
aledegano: I've tested and found out that while it is possible to use a job it seems we'd have nothing to gain from it, so I'd go for the simplest solution of the naked pod, which is also what's used in the Helm test documentation example.

acceptance-tests/README.md
aledegano: I have updated the README as you suggested.
I'll tackle the all story about user generation in a following PR (perhaps also after some team's discussion).

charts/renku/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku platform
 name: renku
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.7.2
+version: 0.7.2-3034fe0
```
pameladelgado, line 6: We normally update this only when making releases, I dont think is necessary to include this change in the PR.

charts/renku/values.yaml
```diff
@@ -131,8 +131,12 @@ keycloak:
   ## keycloak admin username/password is supplied.
   initRealm:
     image:
-      repository: renku/init-realm
-      tag: 'latest'
+      repository: degano/init-realm
```
pameladelgado, line 134: Same here, I guess you'd clean this and the tests images up before merging as well.
We can publish the official tests image after merging.

charts/renku/Chart.yaml
aledegano: This is the result of using `chartpress`, I'm including it in the PR so the tests can run correctly (as they need to pick a new image).
Once the review is done I'll reset them before merging.

charts/renku/values.yaml
aledegano: Yes, indeed.

rokroskar (changes requested):
Looks great! I think we need to add the test resource only if the tests are requested, otherwise there are potential chart rendering issues.

charts/renku/templates/tests/test-renku.yaml
```diff
@@ -1,51 +1,42 @@
...
+      - name: RENKU_TEST_FULL_NAME
+        value: "{{ .Values.tests.testUser.fullname }}"
+      - name: RENKU_TEST_USERNAME
+        value: "{{ .Values.tests.testUser.username }}"
+      - name: RENKU_TEST_EMAIL
+        value: "{{ .Values.tests.testUser.email }}"
+      - name: RENKU_TEST_PASSWORD
+        value: "{{ .Values.tests.testUser.password }}"
+      - name: RENKU_TEST_REGISTER
+        value: "1"
```
ableuler, line 28: It would be great if all test parameters listed [here](https://github.com/SwissDataScienceCenter/renku/tree/master/acceptance-tests#running-using-sbt) could also be specified through the chart. This will allow us to put all test configuration specific to an instance into the values file and remove some hard-coded defaults (see [this issue](https://github.com/SwissDataScienceCenter/renku/issues/1822)).

ableuler (changes requested):
Gave it another test and it works like a charm! If you could add the possibility to add all test-related environment variables through the chart, that would be great!

charts/renku/templates/tests/test-renku.yaml
aledegano: Done: it is now possible to specify all the env parameters of the acceptance tests through the Helm values.

charts/renku/templates/tests/test-renku.yaml
```diff
@@ -1,51 +1,42 @@
...
     env:
-      - name: RENKU_ENDPOINT
+      - name: RENKU_TEST_URL
         value: {{ template "http" . }}://{{ .Values.global.renku.domain }}
-    lifecycle:
-      preStop:
-        exec:
-          command: ["/tests/demo-cleanup.sh"]
+      - name: RENKU_TEST_FULL_NAME
+        value: "{{ .Values.tests.testUser.fullname }}"
```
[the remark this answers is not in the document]
aledegano: Done: the whole test is wrapped in a conditional now.

aledegano:
@rokroskar @ableuler I have implemented your comments.

charts/renku/values.yaml
```diff
@@ -589,22 +593,18 @@ notebooks:
...
-  #    password: IamBatman
-  #  - username: ckent
-  #    firstName: Clark
-  #    lastName: Kent
-  #    email: ckent@example.com
-  #    password: IamSuperman
+    repository: degano/tests
+    tag: '0.7.4-8497790'
+  ## User account for running `helm test`
+  #parameters:
```
rokroskar, line 601: sorry to ask for one more thing but could you add the rest of the parameters here to serve as documentation?

rokroskar (changes requested):
Looking good! Just a minor comment to add the rest of params to the values file.

ableuler (approved):

charts/renku/values.yaml
aledegano: Sure!

rokroskar (dismissed):

ableuler (approved):

PR #1782, opened 2020-12-10
feat: run GH action acceptance tests through helm test

aledegano:
The GH action running the acceptance tests on every PR opened on this repository is now ran through `helm test`.
The values of the chart are updated accordingly.    
The output artifacts of the tests (images and logs) are uploaded temporarily to a SWITCH S3 bucket from the test pod, then downloaded by the GitHub action to be served through the action artifact service (preserving the current behavior).

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -125,6 +125,7 @@ jobs:
                     --set tests.parameters.email="renku@datascience.ch" \
                     --set tests.parameters.password="$RENKU_BOT_DEV_PASSWORD" \
                     --set tests.parameters.anonAvailable="true" \
+                    --set tests.parameters.register="true" \
                     --set tests.resultsS3.host="os.zhdk.cloud.switch.ch" \
                     --set tests.resultsS3.bucket="dev-acceptance-tests-results" \
                     --set tests.resultsS3.filename="tests-artifacts-${GITHUB_SHA}" \
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -92,6 +92,8 @@ jobs:
         RENKUBOT_RANCHER_APISECRET: ${{ secrets.RENKUBOT_RANCHER_APISECRET }}
         CI_PROJECT_ID: ${{ secrets.CI_PROJECT_ID }}
         RENKU_BOT_DEV_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
+        OS_DEV_ACCESS: ${{ secrets.OS_DEV_ACCESS }}
+        OS_DEV_SECRET: ${{ secrets.OS_DEV_SECRET }}
       run: |
         echo "$RENKUBOT_KUBECONFIG" > renkubot-kube.config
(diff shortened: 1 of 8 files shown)


.github/workflows/pull-request-test.yml
```diff
@@ -162,37 +172,28 @@ jobs:
...
-                           -e RENKU_TEST_PASSWORD="$RENKU_BOT_DEV_PASSWORD" \
-                           -e RENKU_TEST_ANON_AVAILABLE="true" sbt
-    - name: Prepare artifact for packaging on failure
+        echo "$RENKUBOT_KUBECONFIG" > renkubot-kube.config
+        helm test $RELEASE --namespace $RELEASE --timeout 40m --logs
+    - name: Download artifact for packaging on failure
+      env:
+        OS_DEV_ACCESS: ${{ secrets.OS_DEV_ACCESS }}
+        OS_DEV_SECRET: ${{ secrets.OS_DEV_SECRET }}
       if: failure()
```
ableuler, line 185: Does the above `helm upgrade --install` command wait for the tests to pass/fail before returning and return a corresponding return value?

ableuler (changes requested):
> ..., then downloaded by the GitHub action to be served through the action artifact service (preserving the current behavior).

That's very convenient. Thanks for going the extra mile here! 👍 

Would be great if you could do the same transition from `docker-compose` based tests to helm tests also for our CD deployment running at dev.renku.ch, see [this workflow definition file](https://github.com/SwissDataScienceCenter/renku/blob/master/.github/workflows/deploy.yml). 

Note that templates for the values files for the CI and CD deployments are [here](https://github.com/SwissDataScienceCenter/ops-dev.renku.ch/blob/master/dev-ci-values.tmpl) and [here](https://github.com/SwissDataScienceCenter/ops-dev.renku.ch/blob/master/renku-values.tmpl). After changing those files we render the templates and store the rendered values file in the corresponding github secret.

.github/workflows/pull-request-test.yml
aledegano: Perhaps you mean the `helm test` command? In that case yes, it wait for the tests to complete (e.g. it really just waits for the pods to terminate) and return the corresponding return code.

.github/workflows/pull-request-test.yml
```diff
@@ -117,10 +119,18 @@ jobs:
                     --set notebooks.jupyterhub.auth.gitlab.clientId=${APP_ID} \
                     --set notebooks.jupyterhub.auth.gitlab.clientSecret=${APP_SECRET} \
                     --set global.anonymousSessions.enabled=true \
+                    --set tests.enabled="true" \
```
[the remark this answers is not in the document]
aledegano: I had started, now I've updated it: https://github.com/SwissDataScienceCenter/ops-dev.renku.ch/pull/108

.github/workflows/pull-request-test.yml
ableuler: Sorry yeah, of course the tests are running on `helm test` and not automatically after the install...

aledegano:
This branch diverged too much from master. I will open a new PR.

PR #1797, opened 2020-12-14
fix: patch 0.7.3 release

pameladelgado:
Downgrading notebooks, since the latest version has a bug.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,10 @@
 .. _changelog:
 
+0.7.4
+-----
+
+This is a patch release that includes a notebooks change and some improvements to the Knowledge Graph.
+
 0.7.3
 -----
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -1,5 +1,10 @@
 .. _changelog:
 
+0.7.4
+-----
+
+This is a patch release that includes a notebooks change and some improvements to the Knowledge Graph.
```
lorenzo-cavazzi, line 6: I'd change the sentence to `includes a change in notebooks` (or remove the `a` article). My English is not top-notch, this may not be really necessary :nerd_face:

lorenzo-cavazzi (approved):
I'd slightly re-phrase the sentence, but it may not be necessary.

CHANGELOG.rst
pameladelgado: right, the notebooks have not really changed is the version of them that changed in the release.
I'll leave it as is for the moment, we can update later if needed.

PR #1813, opened 2021-01-04
fix: publish-chart not to fail on accessing stable

jachro:


jachro:
Closing as the issue got properly fixed by #1814

PR #1850, opened 2021-01-15
chore: update the auto pr workflow

rokroskar:
This updates the automatic PR workflow such that it adds `/deploy` in the PR body to ensure that deployment + tests run.


ableuler (approved):

ableuler:
Guess we should get those changes also into `development`?

PR #1867, opened 2021-01-20
chore: modify CD workflow for deployment script

rokroskar:
This PR does the CD using the `deploy-dev-renku`script.

~~Note that it is currently configured to run from the branch of this PR for testing - once it works I'll switch the branch requirements on the workflow.~~

See [here](https://github.com/SwissDataScienceCenter/renku/actions/runs/499768279) for a successful deploy - acceptance tests are failing, but that seems to be a problem in general and not related to this PR. 

addresses #1631

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,54 +12,25 @@ on:
 
 
 jobs:
-  github-env:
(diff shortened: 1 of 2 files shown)


ableuler (approved):
LGTM 👍

PR #1872, opened 2021-01-21
Update cli-installation.rst

champost:
The $(conda env list | grep 'renku ' | awk '{print $2}')/bin/pip install command assumes that there is single conda environment named renku but there may be other *renku* environments locally. The proposed changes account for these situations.


ciyer:
Thanks a lot for this improvement; we really appreciate contributions like this! Due to the way our CI pipeline is set up, it will not run correctly on forks, so I have created a branch and brought in your commit there, so you will be properly credited for the contribution. I will close this PR in favor of #1878.

PR #1878, opened 2021-01-22
docs: update cli-installation.rst

ciyer:
The $(conda env list | grep 'renku ' | awk '{print $2}')/bin/pip install command assumes that there is single conda environment named renku but there may be other *renku* environments locally. The proposed changes account for these situations.


docs/user/cli-installation.rst
```diff
@@ -112,9 +112,9 @@ it installed, you can run
 .. code-block:: console
 
   conda create -y -n renku python=3.6
-  $(conda env list | grep renku | awk '{print $2}')/bin/pip install -e git+https://github.com/SwissDataScienceCenter/renku-python.git#egg=renku
+  $(conda env list | grep 'renku ' | awk '{print $2}')/bin/pip install -e git+https://github.com/SwissDataScienceCenter/renku-python.git#egg=renku
```
Panaetius, line 115: We should do `grep '^renku\s'` so it also covers the case of an env being name e.g. `my-renku` as well as `renku-env`

Panaetius (approved):

ableuler:
@ciyer can this one be merged?

pameladelgado:
@ciyer this seems to be ready to merge, anything against doing it?

PR #1890, opened 2021-01-26
fix: typo in anonymous notebook deploy

rokroskar:


actions/deploy-renku/deploy-dev-renku.py
```diff
@@ -220,11 +220,11 @@ def configure_requirements(tempdir, reqs, component_versions):
             check_notebooks_chart_version,
         )
 
-        renku_notebooks_version = component_versions.get("renku-notebooks") or check_notebooks_chart_version(
+        renku_notebooks_version = component_versions.get("renku_notebooks") or check_notebooks_chart_version(
```
ableuler, line 223: Shouldn't it be `renku-notebooks` here too? At the top of the script the different component names are defined are defined with hyphens.

ableuler: Ah, sorry, seen the replace in the code now.

ableuler (approved):

PR #1900, opened 2021-01-28
chore: upgrade keycloak dependency

ableuler:
/deploy

~(still work in progress, but I want a CI deployment)~ ready for review now!

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -73,7 +73,7 @@ jobs:
         RENKU_ANONYMOUS_SESSIONS: true
         RENKU_RELEASE: ci-renku-${{ github.event.number }}
         RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
-        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES_1900 }}
+        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
         RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
         RENKUBOT_RANCHER_BEARER_TOKEN: ${{ secrets.RENKUBOT_RANCHER_BEARER_TOKEN }}
         TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
@@ -113,7 +113,7 @@ jobs:
       if: failure()
       uses: ./actions/download-test-artifacts
       env:
-        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES_1900 }}
+        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
         TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
     - name: Upload screenshots on failure
       if: failure()
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -72,7 +72,7 @@ jobs:
         RENKU_ANONYMOUS_SESSIONS: true
         RENKU_RELEASE: ci-renku-${{ github.event.number }}
         RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
-        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
+        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES_1900 }}
         RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
         RENKUBOT_RANCHER_BEARER_TOKEN: ${{ secrets.RENKUBOT_RANCHER_BEARER_TOKEN }}
         TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
@@ -112,7 +112,7 @@ jobs:
       if: failure()
       uses: ./actions/download-test-artifacts
       env:
-        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
+        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES_1900 }}
         TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
     - name: Upload screenshots on failure
       if: failure()
--- a/helm-chart/README.rst
+++ b/helm-chart/README.rst
@@ -61,7 +61,7 @@ To run tests on the deployment, use
 Upgrading
 ---------
 Most information related to upgrading from one chart version to another is covered
-in the `values changelog file<https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md>`_.
+in the `values changelog file <https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md>`_.
 For upgrades that require some steps other than modifying the values files to be
 executed, we add some instructions here.
 
@@ -69,10 +69,10 @@ Upgrading to 0.8.0
 ******************
 This upgrade comes with an upgrade of the keycloak chart from ``4.10.2`` to ``9.8.1``! For
 details on this upgrade check the dedicated section in the
-`the keycloak chart docs<https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#upgrading>`_
-and the `keycloak docs<https://www.keycloak.org/docs/latest/upgrading/>`_.
+`the keycloak chart docs <https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#upgrading>`_
+and the `keycloak docs <https://www.keycloak.org/docs/latest/upgrading/>`_.
 
-- Before starting, make sure to check out `the values changelog for this upgrade<https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md#upgrading-to-renku-080-includes-breaking-changes>`_
+- Before starting, make sure to check out `the values changelog for this upgrade <https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md#upgrading-to-renku-080-includes-breaking-changes>`_
   and update your values file accordingly.
 
 - The upgrade of keycloak will perform an **irreversible database migration**. It is therefore recommended
--- a/helm-chart/README.rst
+++ b/helm-chart/README.rst
@@ -67,7 +67,7 @@ executed, we add some instructions here.
 
 Upgrading to 0.8.0
 ******************
-This upgrade comes with an upgrade of the keycloak chart from `4.10.2` to `9.8.1`! For
+This upgrade comes with an upgrade of the keycloak chart from ``4.10.2`` to ``9.8.1``! For
 details on this upgrade check the dedicated section in the
 `the keycloak chart docs<https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#upgrading>`_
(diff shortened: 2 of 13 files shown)


.github/workflows/pull-request-test.yml
```diff
@@ -112,7 +112,7 @@ jobs:
       if: failure()
       uses: ./actions/download-test-artifacts
       env:
-        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
+        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES_1900 }}
```
aledegano, line 115: I guess these values will be reverted before merging, right?

ableuler: Well... I'd find it the cleanest solution to create new values files and put them into new secrets every time we have to change the values. But currently this would break some assumptions in our component repos, so for the time being I'll just use this for testing, revert back once the PR is ready, update the secret to make the PR pass (thereby breaking all other CI pipelines) and merge this PR asap.

aledegano: ah, right! The classical conundrum of the global configuration... yeah, I don't see many solutions, the only other one would be to include a timestamp/version in these variables (e.g. CI_RENKU_VALUES_20210202), but then we'll have to deal with the garbage...
I like your approach better.

helm-chart/README.rst
```diff
@@ -56,3 +56,43 @@ To run tests on the deployment, use
...
+- If they have not been explicitly defined (and thus autocreated through a helm hook), make sure to get the value
+  of get the `keycloak-postgres-password` from the `renku-keycloak-postgres` secret and add it to your values file
+  as `global.keycloak.postgresPassword.value`. Do the same for `keycloak-password` from the `keycloak-password-secret`
+  and add it as `global.keycloak.postgres.value`.
+
+- Delete the two secrets which need to be recreated as well as the keycloak StatefulSet:
+.. code-block:: bash
+
+   kubectl delete secrets -n <namespace> keycloak-password-secret renku-keycloak-postgres
+   kubectl delete statefulsets.apps -n <namespace> <release-name>-keyclo
```
aledegano, line 90:
```suggestion
   kubectl delete statefulsets.apps -n <namespace> <release-name>-keycloak
```

aledegano, line 90:
```suggestion
   kubectl delete statefulsets.apps -n <namespace> <release-name>-key
```

helm-chart/README.rst
```diff
@@ -56,3 +56,43 @@ To run tests on the deployment, use
...
+- Before starting, make sure to check out `the values changelog for this upgrade<https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md#upgrading-to-renku-080-includes-breaking-changes>`_
+  and update your values file accordingly.
+
+- The upgrade of keycloak will perform an irreversible database migration. It is therefore recommended
+  to back up your postgres volume before performing this upgrade.
+
+- If they have not been explicitly defined (and thus autocreated through a helm hook), make sure to get the value
+  of get the `keycloak-postgres-password` from the `renku-keycloak-postgres` secret and add it to your values file
+  as `global.keycloak.postgresPassword.value`. Do the same for `keycloak-password` from the `keycloak-password-secret`
+  and add it as `global.keycloak.postgres.value`.
```
aledegano, line 84:
```suggestion
  and add it as `global.keycloak.password.value`.
```

helm-chart/README.rst
```diff
@@ -56,3 +56,43 @@ To run tests on the deployment, use
...
+`the keycloak chart docs<https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#upgrading>`_
+and the `keycloak docs<https://www.keycloak.org/docs/latest/upgrading/>`_.
+
+- Before starting, make sure to check out `the values changelog for this upgrade<https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md#upgrading-to-renku-080-includes-breaking-changes>`_
+  and update your values file accordingly.
+
+- The upgrade of keycloak will perform an irreversible database migration. It is therefore recommended
+  to back up your postgres volume before performing this upgrade.
+
+- If they have not been explicitly defined (and thus autocreated through a helm hook), make sure to get the value
```
aledegano, line 81: I would rephrase this to really highlight that the two passwords have to be backed up before deleting them and just mention that is not necessary later.
e.g.:
"[WARNING] Backup the secrets: `renku-keycloak-postgres.keycloak-postgres-password` and `keycloak-password-secret.keycloak-password` with the command 
[...]
This operation is not needed if the secrets were previously explicitly defined [...]"

helm-chart/renku/values.yaml
```diff
@@ -116,14 +118,55 @@ ingress:
...
+  # renku-keycloak-postgres
+  # - DB_ADDR: database host                            value: <RELEASE-NAME>-postgresql
+  # - DB_DATABASE: database name for keycloak           value: global.keycloak.postgresDatabase
+  # - DB_USER: database username for keycloak database  value: global.keycloak.postgresUser
+  # - DB_PASSWORD: database password for keycloak       value: global.keycloak.postgresPassword.value
+  # keycloak-password-secret
+  # - KEYCLOAK_USER: keyclaok admin username            value: global.keycloak.user
+  # - KEYCLOAK_PASSWORD: keycloak admin password        value: global.keycloak.password.value
+
+  extraEnvFrom: |
```
aledegano, line 150: Something's not very clear to me: the global value: `global.keycloak.postgresPassword.value` and the secret `keycloak-password-secret` are a redundant: I don't understand if both should be present, or either depending if the operators is performing an upgrade or a clean installation.
Perhaps we could clarify in the instructions?

helm-chart/README.rst
ableuler: oops - I've seen so many 7-letter namespaces (andreas, ci-1234) that I've come to believe that `<release-name>-keyclo` was generic...

helm-chart/README.rst
ableuler: 👍

helm-chart/renku/values.yaml
ableuler: The value of `global.keycloak.postgresPassword.value` is written into `keycloak-password-secret`. If `global.keycloak.postgresPassword.value` is missing, the actual password is autogenerated through helm. In this second case, we have to copy the password from the secret into the values file if we want the password to remain unchanged over the deletion of the secret and the subsequent `helm upgrade`.

helm-chart/README.rst
ableuler: 👍

helm-chart/README.rst
ableuler: fixed by querying the actual name first.

helm-chart/README.rst
ableuler: Tried to make things clearer.

helm-chart/renku/values.yaml
ableuler: I agree that the secrets handling could be documented better in general, but I'm not sure the instructions on how to upgrade the renku helm chart version is the best place for this.

aledegano: ok, thanks for the explanation.

aledegano (dismissed):

helm-chart/renku/values.yaml
```diff
@@ -116,14 +118,55 @@ ingress:
...
+      emptyDir: {}
+
+
   ## Create a demo user in keycloak? Note that the password for the demo
   ## user must be queried from kubernetes (see the rendered NOTES.txt
   ## template which is shown after a successul deployment).
   createDemoUser: false
 
-  ## Spawn a keycloak instance
-  enabled: true
```
pameladelgado, line 125: dont we need this anymore? what about deployments with external keycloak?

pameladelgado (commented):
Thanks for doing this!
I haven't tried this yet but it looks good. I just have a couple of comments/questions.
Have you already tried updating keycloak in an existing deployment? otherwise I could do it.

Also, the changelog notes are clear enough to me.

helm-chart/renku/templates/_keycloak-clients-users.tpl
```diff
@@ -46,7 +46,7 @@ Define clients and users for Keycloak
 [
   {{- if .Values.keycloak.createDemoUser -}}
   {
-    "username": "demo",
+    "username": "demo@datascience.ch",
```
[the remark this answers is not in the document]
ableuler: I agree, not the smartest detail for a demo user. However, in this PR I merely fixt the fact that the email should be used as username for our current keycloak settings. Older keycloak versions still tolerated using a non-email username, but now this fails.

helm-chart/renku/values.yaml
ableuler: It's still there, just moved a bit.

helm-chart/README.rst
```diff
@@ -56,3 +56,49 @@ To run tests on the deployment, use
...
+
+- Before starting, make sure to check out `the values changelog for this upgrade<https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md#upgrading-to-renku-080-includes-breaking-changes>`_
+  and update your values file accordingly.
+
+- The upgrade of keycloak will perform an **irreversible database migration**. It is therefore recommended
+  to **back up your postgres volume** before performing this upgrade.
+
+- **Warning: Persist keycloak-related secrets!**
+
+  If `global.keycloak.postgresPassword.value` and `global.keycloak.postgresPassword.value`
```
olevski, line 83: @ableuler I think you mean to say here the postgress password and the keycloak password. But you have just the same thing repeated.

helm-chart/README.rst
```diff
@@ -56,3 +56,49 @@ To run tests on the deployment, use
...
+  have not been explicitly defined in the values file (and thus have been autocreated by helm),
+  add them to the values file now.
+
+  * Get the `keycloak-postgres-password` from the `renku-keycloak-postgres` secret and add it as `global.keycloak.postgresPassword.value`.
```
olevski, line 87: @ableuler this is probably just me... But I overlooked here that the revised section in my values file should be:
```
global:
  keycloack:
    postgressPassword:
      value: actual_password_value
```

And not 
```
global:
  keycloack:
    postgressPassword: actual_password_value
```

Not sure if it is worth adding this as an example here. I could not make a code suggestion that included a multiline code block inside of it.

olevski (approved):
@ableuler you have my approval for this. I tested on my dev environment and following your instructions I did not lose (what little data) I had in my keycloak.

The proposals for edits in the documentation are very minor so you can do them here and merge or do them in a subsequent PR.

Oh and a few reminders:
- update the branch because in the meantime some change occurred on master
- revert back the `secrets.CI_RENKU_VALUES_1900` values when appropriate

aledegano (dismissed):

helm-chart/renku/templates/_keycloak-clients-users.tpl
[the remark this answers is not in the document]
pameladelgado: right, we could maybe add this as an optional parameter for the values file. What do you think? It could also be in another PR...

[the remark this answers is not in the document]
ableuler: Yeah, I think the cleanest would be that if the admin want's to automatically create users, all details of those users have to be provided explicilty, no assumptions from the chart side about those users.

helm-chart/README.rst
ableuler: 👍

ableuler:
@olevski thanks for the careful review - added all your suggestions.

olevski (approved):

PR #1920, opened 2021-02-01
Close development branch

ableuler:
- run CD deployment from master
- let auto-update PRs be created against master
- merge development into master

/deploy

### DIFF
--- a/helm-chart/renku/requirements.yaml
+++ b/helm-chart/renku/requirements.yaml
@@ -2,19 +2,19 @@ dependencies:
 - name: renku-core
   alias: core
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: 0.12.2
+  version: 0.12.3-93348f9
 - name: renku-ui
   alias: ui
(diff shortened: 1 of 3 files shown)


helm-chart/renku/requirements.yaml
```diff
@@ -2,27 +2,27 @@ dependencies:
 - name: renku-core
   alias: core
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-  version: 0.12.2
+  version: 0.12.3-93348f9
```
rokroskar, line 5: these should be latest released versions, no?

ableuler: My thinking was to just take over whatever is on development now, remove development so that people stop updating dependencies there and then figure out the right tagged version for each component in the release-prep PR that should follow shortly after.

rokroskar: ok sounds good!

rokroskar (approved):
🤖 🚀

PR #1923, opened 2021-02-02
[wip] Fix post-install jobs for external deployment

pameladelgado:
/deploy
See https://github.com/SwissDataScienceCenter/renku/pull/1516#discussion_r567697103

### DIFF
--- a/helm-chart/renku/templates/post-install-job-keycloak.yaml
+++ b/helm-chart/renku/templates/post-install-job-keycloak.yaml
@@ -60,3 +60,4 @@ spec:
         - name: realm-data
           secret:
             secretName: {{ template "renku.fullname" . }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


helm-chart/renku/templates/post-install-job-keycloak.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.keycloak.enabled }}
```
aledegano, line 1: I'm a bit confused: if you're adding an `if` shouldn't be there also an `end`?

pameladelgado: yes, good catch. In the rush I forgot to push it.
Should be ok now :)

aledegano (dismissed):

pameladelgado:
closing in favor of #1928

PR #1932, opened 2021-02-04
chore: fix chart publishing

ableuler:


pameladelgado (approved):
thanks for fixing this!

PR #1937, opened 2021-02-05
test: update acceptance tests to comply with ui changes

lorenzo-cavazzi:
Updating the acceptance test to support recent UI changes: SwissDataScienceCenter/renku-ui#1229
/deploy

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,12 +2,14 @@ name: Deploy and Test
 
 on:
   push:
+    branches: ['master']
+    # Note that branch patterns are ignored for tag pushes.
     paths:
     - helm-chart/**
     - scripts/**
     - acceptance-tests/**
     - .github/**
-    branches: ['master']
+    tags: ['*']
   workflow_dispatch:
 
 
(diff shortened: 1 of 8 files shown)


rokroskar:
Will this break others who are running against the last released version of renku-ui?

lorenzo-cavazzi:
This should be merged just after the corresponding UI PR, I added the "Do not merge" label as a reminder.
I'm testing it in a draft PR in the UI

rokroskar:
yes, but the acceptance tests run against the requirements specified in the master branch which only get updated after a tag. So this should get merged after the next _release_ of renku-ui not after the PR is merged.

lorenzo-cavazzi:
Good point! I'm still not used to this new paradigm, thanks for the reminder.
That is a bit annoying though. I guess the best solution would be to tag a new UI release after merging the PR.

rokroskar:
yes it's not ideal perhaps. Alternatively, you can continue running the tests in renku-ui from this branch.

lorenzo-cavazzi:
This is funny: the acceptance tests are failing, throwing the same error I would expect without the fix from the first commit.
If I run them manually on https://ci-renku-1937.dev.renku.ch, I don't get that error.
Since the GitHub action has a run time of in 6 seconds, I assume something is wrong. Maybe the run is cached? :thinking:

lorenzo-cavazzi:
Creating a different commit with the same changes from this PR, and pushing it again has worked fine #1960 
Sadly, I couldn't find the cause of the problem.

Closing the PR since it's not needed anymore.

PR #1939, opened 2021-02-05
chore: release 0.7.6

pameladelgado:
/deploy

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@
 0.7.6
 -----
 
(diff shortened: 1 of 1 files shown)


olevski:
@pameladelgado that is the correct renku-notebooks version

olevski (approved):

jachro (approved):
👍

ableuler (approved):
We should also bump the chart version in `helm-chart/renku/Chart.yaml`. But we can do that after master has been tagged to not restart the entire test pipeline.

pameladelgado:
> We should also bump the chart version in `helm-chart/renku/Chart.yaml`. But we can do that after master has been tagged to not restart the entire test pipeline.

ah right, good point :+1:

PR #1948, opened 2021-02-09
Make S3 config optional

pameladelgado:
Not all renkulab deployments have the option to configure an S3 bucket, this PR makes this section optional so that these deployments can still benefit from running tests via Helm.

/deploy


aledegano (dismissed):

aledegano:
I just realized that we have to add `tests.enabled: true` in few places when we merge this, otherwise every new deployment won't upload the tests by default...

pameladelgado:
yes, I was doing this for the CI values. You can include the same in your PR for CD.

aledegano:
> yes, I was doing this for the CI values. You can include the same in your PR for CD.

I've updated the GH action secret too.

aledegano (approved):

PR #1957, opened 2021-02-10
feat: renku login support (always forward gitlab to gateway)

mohammad-sdsc:
To authenticate git requests from CLI, we need to forward all request that go to `/gitlab` to the gateway.

Fixes: https://github.com/SwissDataScienceCenter/renku-python/issues/677

### DIFF
--- a/helm-chart/renku/templates/ingress.yaml
+++ b/helm-chart/renku/templates/ingress.yaml
@@ -52,12 +52,10 @@ spec:
           serviceName: {{ $keycloakFullname }}-http
           servicePort: {{ $keycloakServicePort }}
       {{- end }}
-      {{- if $gitlabEnabled }}
       - path: /gitlab
         backend:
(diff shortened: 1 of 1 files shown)


helm-chart/renku/templates/ingress.yaml
```diff
@@ -52,12 +52,10 @@ spec:
           serviceName: {{ $keycloakFullname }}-http
           servicePort: {{ $keycloakServicePort }}
       {{- end }}
-      {{- if $gitlabEnabled }}
```
pameladelgado, line 55: how does this work for deployments that dont have GitLab enabled and their path is not `/gitlab`?

mohammad-sdsc:
We've decided not to forward all `/gitlab` requests to gateway. We will add a specific endpoint for repositories that use renku CLI login to only forward requests that have a CLI token.

PR #1986, opened 2021-02-23
chore: remove some obsolete ui values from chart

ableuler:
/deploy


pameladelgado:
@ableuler would you mind moving this to another PR to get a different namespace? renkubot is not able to deploy because of the weave bug (I believe the namespace was deleted while holding some resources in it).

ableuler:
I've quickly restarted weave in our dev-cluster, removed the helm release and triggered the pipeline again. This should fix it.

pameladelgado (approved):
LGTM!
I would be good to have a unique `gitlabUrl` in the global section for all of the components but that can wait for a chart refactor.

PR #1988, opened 2021-02-25
chore: use new ingress anotation for cert-manager

aledegano:


pameladelgado (commented):
From what I understand the deprecated `ingress.class` annotation is still needed for k8s <1.18 and nginx-ingress, so I would keep it there for the moment.

aledegano:
> From what I understand the deprecated `ingress.class` annotation is still needed for k8s <1.18 and nginx-ingress, so I would keep it there for the moment.

uhm... that is strange because I have tested in my namespace in `dev`:

1. removing `ingress.class`
2. adding the new annotation
3. deleting the existing certificate

And it recreated the certificate correctly.

pameladelgado:
> > From what I understand the deprecated `ingress.class` annotation is still needed for k8s <1.18 and nginx-ingress, so I would keep it there for the moment.
> 
> uhm... that is strange because I have tested in my namespace in `dev`:
> 
>     1. removing `ingress.class`
> 
>     2. adding the new annotation
> 
>     3. deleting the existing certificate
> 
> 
> And it recreated the certificate correctly.

Right, but this annotation is for the ingress, not for the certificate. So to test this you would need to recreate the ingresses.

aledegano:
> > > From what I understand the deprecated `ingress.class` annotation is still needed for k8s <1.18 and nginx-ingress, so I would keep it there for the moment.
> > 
> > 
> > uhm... that is strange because I have tested in my namespace in `dev`:
> > ```
> > 1. removing `ingress.class`
> > 
> > 2. adding the new annotation
> > 
> > 3. deleting the existing certificate
> > ```
> > 
> > 
> > And it recreated the certificate correctly.
> 
> Right, but this annotation in particular is for the ingress, not for the certificate. So to test this you would need to recreate the ingresses.

Oh, right! I was confused!
I'll fix this and all other PRs! Sorry for the noise!

pameladelgado (approved):
Looks good! :+1:

PR #2013, opened 2021-03-10
charts: adapting resource requests for tests

ableuler:
After merging #2011 the tests run at pretty exactly at 2GB of memory and <= 1 CPU.

/deploy

### DIFF
--- a/helm-chart/renku/templates/tests/test-renku.yaml
+++ b/helm-chart/renku/templates/tests/test-renku.yaml
@@ -81,10 +81,10 @@ spec:
     resources:
       requests:
         memory: "3G"
-        cpu: "1500m"
+        cpu: "1000m"
         ephemeral-storage: "1G"
       limits:
         memory: "3G"
-        cpu: "1000m"
+        cpu: "1500m"
         ephemeral-storage: "2G"
 {{- end }}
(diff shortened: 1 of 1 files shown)


helm-chart/renku/templates/tests/test-renku.yaml
```diff
@@ -80,11 +80,11 @@ spec:
...
-        memory: "4G"
-        cpu: "2"
+        memory: "3G"
+        cpu: "1500m"
         ephemeral-storage: "1G"
       limits:
-        memory: "4G"
-        cpu: "2"
+        memory: "3G"
+        cpu: "1000m"
```
aledegano, line 88: Is it normal that the limit is lower than the request?

pameladelgado: is normally the other way around, probably was a typo...

pameladelgado:
what about this particular run? https://github.com/SwissDataScienceCenter/renku/runs/2077083776
do you know if it failed because of another non-resource-related reason?

helm-chart/renku/templates/tests/test-renku.yaml
ableuler: oh gosh, I seem to be unable to edit these properly through the gitlab web UI...

pameladelgado (approved):
LGTM

Issue #1977, opened 2021-02-17
Automatically deploy to staging on tag


ableuler:
Whenever we tag a release or a release candidate, the tagged version should be deployed automatically on a staging deployment which runs against the production gitlab instance.

pameladelgado:
we can use the existing `staging` dev deployment, we need to:
- move the namespace to Renku project on rancher
- extract, encrypt and save its values on the ops-dev repo
- same for PVs, PVCs
- automate the deployment on a CD pipeline, including running the acceptance tests
- ideally I would also make an upgrade to an rc/official release after a tag has been made and the charts are available (to verify the charts are properly built)

ableuler:
> * ideally I would also make an upgrade to an rc/official release after a tag has been made and the charts are available (to verify the charts are properly built)

Not sure I understand what you mean here. I would expect every tag to be deployed on this staging environment only after the tag has been made and the corresponding chart has been published. Did you have something else in mind?

PR #2019, opened 2021-03-11
feat: publish chart of tag push and deploy on staging

aledegano:
This is the first part of #1977: on a new tag push create and push the Renku Helm chart.
In a following PR we will deploy the new chart in Staging.
/deploy


aledegano:
I'll work some more on this PR, closing for now.

PR #2020, opened 2021-03-11
fix: include release breaking change in docs

pameladelgado:


CHANGELOG.rst
```diff
@@ -61,6 +61,18 @@ For changes to individual components, please check:
 * renku-notebooks:
   `0.8.10 <https://github.com/SwissDataScienceCenter/renku-notebooks/releases/tag/0.8.10>`__
 
+Upgrading from 0.7.7
+~~~~~~~~~~~~~~~~~~~~~
+
+
+* **Breaking change**  Keycloak chart dependency has been upgraded from ``4.10.2`` to ``9.8.1``, check out the instructions on how to upgrade aspects not covered by default in the Renku chart. Most notably, keycloak values are less nested, so at the level of the Renku chart values, keycloak.keycloak.X.Y becomes keycloak.X.Y.
+
+If the Renkulab deployment includes keycloak, the values file should be modified as follows:
```
ableuler, line 70: I'd just link to the chart readme here: https://github.com/SwissDataScienceCenter/renku/tree/master/helm-chart#upgrading-to-080 (and also fix the version number from 0.8.0 to 0.7.8 there too.

ableuler (changes requested):

CHANGELOG.rst
pameladelgado: >I'd just link to the chart readme

ok, I'll add the link

>also fix the version number from 0.8.0 to 0.7.8 there too

already did that

pameladelgado: oh wait, why is this here? I thought we were keeping everything in the changelog.md file...

pameladelgado: done, I also added other references

ableuler: thanks - it seems to be common practice to add upgrading instructions to the chart readme. On the other hand I like the idea of having a place where we just keep track of actual values changes. Not sure if we want to merge the two, but for the time being, I think referring to both places is the right thing to do.

ableuler (approved):

ableuler (approved):
👍

PR #2025, opened 2021-03-16
chore: expose uiserver service through ingress

lorenzo-cavazzi:
As suggested here SwissDataScienceCenter/renku-gateway#311 , we can start exposing the UI service through ingress.

This PR should supersede #1575

### DIFF
--- a/helm-chart/renku/requirements.yaml
+++ b/helm-chart/renku/requirements.yaml
@@ -7,6 +7,10 @@ dependencies:
   alias: ui
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
   version: 0.11.8
+- name: renku-ui-server
+  alias: uiserver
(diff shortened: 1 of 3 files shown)


helm-chart/example-configurations/minimal-values.tmpl
```diff
@@ -1,5 +1,5 @@
 #
-# Basic configuration file for Renku 0.7.1
+# Basic configuration file for Renku 0.7.8
```
ableuler, line 2: I think we should just get rid of the version tag here, no way we'll always remember to bump this.

ableuler (commented):
LGTM!

helm-chart/example-configurations/minimal-values.tmpl
lorenzo-cavazzi: Yes, I agree. I've removed it

ableuler (approved):

PR #2033, opened 2021-03-18
chore: fix minimal-values template

ableuler:


pameladelgado (approved):
Great, thanks for spotting this!

PR #2037, opened 2021-03-22
feat: use Slack notification action for Deploy workflow

aledegano:
Replace direct calls to Slack's API with notification action.

Gives more information, including link to the commit that triggered the workflow.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -34,15 +34,20 @@ jobs:
         RENKU_TESTS_ENABLED: true
         TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
         renku: "@master"
-    - name: Notify slack
-      if: success()
+    - uses: 8398a7/action-slack@v3
+      with:
+        status: custom
+        fields: job,ref
+        custom_payload: |
+          {
+            attachments: [{
+              color: '${{ job.status }}' === 'success' ? 'good' : '${{ job.status }}' === 'failure' ? 'danger' : 'warning',
+              text: `${process.env.AS_JOB} commit ${process.env.AS_COMMIT}: ${{ job.status }}.`,
+            }]
+          }
       env:
-        RENKU_SLACK_BOT_TOKEN: ${{ secrets.RENKU_SLACK_BOT_TOKEN }}
-      run: |
-        curl -X POST https://slack.com/api/chat.postMessage \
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -69,15 +74,20 @@ jobs:
...
+          }
       env:
-        RENKU_SLACK_BOT_TOKEN: ${{ secrets.RENKU_SLACK_BOT_TOKEN }}
-      run: |
-        curl -X POST https://slack.com/api/chat.postMessage \
-             -H "Authorization: Bearer $RENKU_SLACK_BOT_TOKEN" \
-             --data "channel=C9U45DL1H" \
-             --data "text=Acceptance tests on https://dev.renku.ch failed! :scream_cat:"
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      if: always()
```
ableuler, line 90:
```suggestion
      if: failure()
```

.github/workflows/deploy.yml
```diff
@@ -36,15 +36,20 @@ jobs:
...
+          }
       env:
-        RENKU_SLACK_BOT_TOKEN: ${{ secrets.RENKU_SLACK_BOT_TOKEN }}
-      run: |
-        curl -X POST https://slack.com/api/chat.postMessage \
-             -H "Authorization: Bearer $RENKU_SLACK_BOT_TOKEN" \
-             --data "channel=C9U45DL1H" \
-             --data "text=https://dev.renku.ch has been updated! :tada:"
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      if: always()
```
ableuler, line 52:
```suggestion
      if: failure()
```
I'd actually switch to a notification on failure only here too. I think letting us know that everything worked as intended is pointless.

ableuler (changes requested):

.github/workflows/deploy.yml
aledegano: Currently we do notify when a new version is deployed. But I'm ok to change it if you don't think there's value in it.

ableuler: Yes, I know. This is definitely not the main reason for our flood of notifications, but since we're touching it I'd propose to make that change now.

aledegano: Okay, agreed 👍

ableuler (approved):

PR #2043, opened 2021-03-24
docs: added doc on running sessions locally

### LINKED ISSUE #2040: instructions on how to reproduce renkulab environment locally
As discussed in https://renku.discourse.group/t/installation-instructions-incomplete/208, it would be good to have an easy way to re-build locally an adequate environment for a renkulab project. The packages listed in requirements.txt are installed on top of the standard installation, but it is not obvious how to reproduce the standard installation locally. If the information contained in the dockervile, environment.yml and requirements.txt is sufficient to achieve this, could some instructions be added to the docu how to use them?

rokroskar:
This PR adds the docs that demonstrate how a renku-generated image can be used to run a "renkulab" session on another machine. 

closes #2040

### DIFF
--- a/docs/user/interactive-running-locally.rst
+++ b/docs/user/interactive-running-locally.rst
@@ -1,0 +1,121 @@
+.. _local_interactive_environments:
+
+Running RenkuLab Interactive Sessions on Your Own Machine
+=========================================================
+
+Each Renku project contains configuration files (``Dockerfile``,
+``requirements.txt`` etc.) for building reproducible session images. We build
+these automatically when you push your project to the server and store them in
+the image registry connected to the Renku instance that you are using. The
+custom-built images are used whenever you launch an interactive session on
+RenkuLab.
+
+You can of course also use these same images to run your code in the correct
+environment anywhere that `Docker <https://www.docker.com>`_ is available.
+
+.. note::
+
+    All of the instructions below assume that you have cloned your project
+    locally and that you have `Docker
+    installed <https://docs.docker.com/get-docker/>`_. On Linux systems, Docker
+    requires root privileges - if you do not have root access to the host, you
+    might consider `rootless
+    Docker <https://docs.docker.com/engine/security/rootless/>`_.
+
+
+Building your own image
+-----------------------
+
+If you are working on a project that has not yet been pushed to the server,
+or you just prefer to build the image locally, you can certainly do so
+with Docker. To build the image, run this command from the root directory
+of your project:
+
+.. code-block::
+
+    $ docker build -t <imageName>:<tag> .
+
+The ``imageName`` and ``tag`` can be anything you want, but you might want to
+try to use something you will remember. To see which images you have in
+your local Docker registry, run
+
+.. code-block::
+
+    $ docker images
+
+
+Using the images built on RenkuLab
+----------------------------------
+
+If your repository on RenkuLab is public, the easiest way to spin up your
+runtime environment locally is by pulling the image that has already been built.
+First, make sure that your project has been pushed and that the image build is
+complete. If you are not sure, the easiest way to check is to navigate to your
+project in RenkuLab and go to launch a new session - if the "Start environment"
+button is available, it means the image has been built already.
+
+To pull the image, you need to construct the image name. This is typically the
+format to expect: ``registry.<renku-host>/<namespace>/<project>:<commit-sha>``.
+Lets break down the meaning of these terms in brackets:
+
+* ``renku-host`` is the hostname of the RenkuLab instance, e.g. ``renkulab.io``
+* ``namespace`` is either the username or the group name of the project owner
+* ``project`` is the project name
+* ``commit-sha`` is the first 7 characters of the ``commit-sha``
+
+``namespace`` and ``project`` are easy - if your repository remote is something
+like ``https://renkulab.io/gitlab/rok.roskar/flights-tutorial.git`` then the
+``<namespace>/<project>`` is ``rok.roskar/flights-tutorial``.
+
+You can get the ``commit-sha`` by running ``git log`` in your repository, e.g.
+
+.. code-block::
+
+    $ git log
+
+    commit 53cbd6a11dc5b0d139b2563b5aad83ae4f676b18 (HEAD -> master, origin/master, origin/HEAD)
+    Author: Rok Roskar <rok.roskar@sdsc.ethz.ch>
+    Date:   Tue Nov 3 17:15:31 2020 +0100
+
+    renku graph generate
+
+The ``commit-sha`` here is ``53cbd6a11dc5b0d139b2563b5aad83ae4f676b18`` so the full
+name of this image, including the tag would be
+``registry.renkulab.io/rok.roskar/flights-tutorial:53cbd6a``.
+
+
+Launching the interactive environment locally
+---------------------------------------------
+
+Finally, if we wanted to launch an environment locally with the image that was
+built for this commit we can do
+
+.. code-block::
+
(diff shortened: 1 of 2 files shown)


docs/user/interactive-running-locally.rst
```diff
@@ -0,0 +1,121 @@
...
+the image registry connected to the Renku instance that you are using. The
+custom-built images are used whenever you launch an interactive session on
+RenkuLab.
+
+You can of course also use these same images to run your code in the correct
+environment anywhere that `Docker <https://www.docker.com>`_ is available.
+
+.. note::
+
+    All of the instructions below assume that you have cloned your project
```
lorenzo-cavazzi, line 18: I would add `jupyter` to the local requirements

docs/user/interactive-running-locally.rst
```diff
@@ -0,0 +1,121 @@
...
+----------------------------------
+
+If your repository on RenkuLab is public, the easiest way to spin up your
+runtime environment locally is by pulling the image that has already been built.
+First, make sure that your project has been pushed and that the image build is
+complete. If you are not sure, the easiest way to check is to navigate to your
+project in RenkuLab and go to launch a new session - if the "Start environment"
+button is available, it means the image has been built already.
+
+To pull the image, you need to construct the image name. This is typically the
```
lorenzo-cavazzi, line 57: There is an easier way to get that information from the UI, not sure if we want to include it since it requires launching the environment.
Anyway, once that is up and running, the user can click the green icon on the left to see the image URL

Screenshot_20210325_094805

docs/user/interactive-running-locally.rst
```diff
@@ -0,0 +1,121 @@
...
+The ``commit-sha`` here is ``53cbd6a11dc5b0d139b2563b5aad83ae4f676b18`` so the full
+name of this image, including the tag would be
+``registry.renkulab.io/rok.roskar/flights-tutorial:53cbd6a``.
+
+
+Launching the interactive environment locally
+---------------------------------------------
+
+Finally, if we wanted to launch an environment locally with the image that was
+built for this commit we can do
```
lorenzo-cavazzi, line 91: It's probably obvious, but I would specify that the command should be executed from the cloned project main folder

lorenzo-cavazzi (commented):
Great addition to the documentation! All works well :tada: 
I've added a couple of (minor) comments to clarify some aspects -- even though they may not be needed.

docs/user/interactive-running-locally.rst
rokroskar: everything runs in docker, so no need to have jupyter installed locally

docs/user/interactive-running-locally.rst
rokroskar: ah I forgot about that! Do you think we could add a button somewhere (maybe on the page listing the project commits) that would let you copy the image name? Also here it would be nice to be able to click to have it sent to the clipboard

docs/user/interactive-running-locally.rst
lorenzo-cavazzi: ops, you are right :sweat_smile:

docs/user/interactive-running-locally.rst
lorenzo-cavazzi: Adding the clipboard copy button is straightforward. Adding the same information in the commit list (you mean the one in the following screenshot?) is a bit more tedious since we need to verify that the entry in the registry exists, so I wouldn't do that for all the commits. We can add another button to request the URL for a specific commit, but it won't be very intuitive.
I'm wondering if it makes sense to add than info (for the latest commit) in the `Settings` page :thinking: 
Once we support changing branch, that would cover most of the cases.

Screenshot_20210325_135744

rokroskar: my initial thought was that if we just had a button to copy the image name, we would make clear that it doesn't on its own guarantee that the image is there.. but I can see that being annoying. It would be nice to not have to go to the settings page for this, but just have it readily available (along with a button to copy the renku clone command) from some front page of the project. Maybe the two locations in red in this screenshot?

Screenshot 2021-03-25 at 14 11 31

In any case, that's a bit off topic here, but if we're going to add something like this I would not complicate the instructions more at this point - it also requires the user to launch an environment first before they can see the image name, which is sort of annoying imho. 

btw, I would also be fine with your proposal to have that on the settings page, although it doesn't really sound like a setting to me.

rokroskar:
Thanks @lorenzo-cavazzi for the review! I will make a follow-up issue for figuring out where the image button should go.

docs/user/interactive-running-locally.rst
lorenzo-cavazzi: > In any case, that's a bit off topic here

Sure, I can open an issue in the UI repo to track this

> it doesn't really sound like a setting to me

That's true, I guess the setting page is not the place.
I wouldn't put this info next to the project name, it seems too prominent, and it would require polling the whole time.
I think an overview tab would work well.
Actually, the `Status` may be a good place. I'm also fine with creating a new one `Registry`. Anyway, let's move this discussion to the UI, I'll link here the new issue.

lorenzo-cavazzi (approved):
:+1:

docs/user/interactive-running-locally.rst
lorenzo-cavazzi: SwissDataScienceCenter/renku-ui#1289

rokroskar: haha you beat me to it! thanks!

schymans:
This is awesome, thank you, @rokroskar! I just tried it, but get an access forbidden: 
```
$ sudo docker pull registry.renkulab.io/gitlab/ba_math_gen-16/et-data:e0796781
[sudo] password for user1: 
Error response from daemon: Get https://registry.renkulab.io/v2/gitlab/ba_math_gen-16/et-data/manifests/e0796781: denied: access forbidden
```
Did I misunderstand or do I need to create a second key for docker? I can push to this repo using git with no problems. Perhaps a full example of pulling and running the docker container would be helpful in the docs. Would save a lot of time to people that don't use docker regularly.
There is also something missing here in the documentation: 
```
Finally, if we wanted to launch an environment locally with the image that was built for this commit. From the project’s root directory, run

Replace ...
```
The command is missing in the above.

erbou:
I found the easiest is to create an API token in gitlab view : profile / settings / access tokens, then:

```
docker login https://registry.renkulab.io
username:
password:
```

rokroskar:
@schymans good point, thanks! I'll add instructions for private repos, but it's basically what @erbou has written above.

rokroskar:
😱 somehow none of the code blocks get rendered in the hosted docs... oops, will fix that asap

rokroskar:
@schymans the code blocks should be fixed in the docs (https://github.com/SwissDataScienceCenter/renku/pull/2050)

schymans:
Thank you, @rokroskar ! I just tried it out, and it works, although it gives an error in the process:

```
repoName=$(basename -s .git `git config --get remote.origin.url`);     sudo docker run --rm -ti -v ${PWD}:/work/$repoName     --workdir /work/$repoName -p 8888:8888     registry.renkulab.io/stanislaus.schymanski/et-data-fork:44054eec jupyter lab
fatal: repository '//renku-env.git' does not exist
find: ‘/tmp/renku-env’: No such file or directory
Ahhhhhhhh! You have found a bug. 🐞
```
Is this related to my repo or to the docker build? It still seems to run, although I am unable to create a new notebook ("Invalid response: 403 Forbidden", then "Cannot read property 'path' of undefined").

rokroskar:
@schymans strange - can you share the Dockerfile?

PR #2047, opened 2021-03-25
fix: disable minio explicitly

pameladelgado:
/deploy

### DIFF
--- a/helm-chart/gitlab/Chart.yaml
+++ b/helm-chart/gitlab/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for the Renku Gitlab server
 name: gitlab
-version: 0.5.2
+version: 0.5.3
--- a/helm-chart/gitlab/values.yaml
+++ b/helm-chart/gitlab/values.yaml
@@ -44,7 +44,8 @@ service:
     annotations:
       prometheus.io/scrape: "true"
       prometheus.io/path: /gitlab/-/metrics
-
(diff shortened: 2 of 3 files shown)


ableuler:
What are you trying to achieve with this PR?

ableuler:
As far as I can tell this won't do anything for two reasons:
1. We still use gitlab `0.5.2` in the renku requirements
2. I don't see`.Values.minio.enabled` being accessed anywhere in the templates of the gitlab chart, so I don't see how it would change the behaviour of this chart.

pameladelgado:
@ableuler sorry, the PR wasn't ready, that's why I didnt include much description.
The minio charts were published as part of the renku charts, even if you deleted it from the gitlab requirements previously. This PR is disabling it explicitly so that those charts are not included in the renku charts and deployments can succeed in k8s versions >1.16.
This PR is ready for review now: I have pushed a new gitlab chart `0.5.3` as specified in this PR and I have published the renku chart `0.7.9-rc1` which achieves the desired effect.

ableuler:
But minio used to be a dependency of the renku chart, not one of the gitlab chart.

pameladelgado:
I'm closing this, the issue seen on upgrading staging is not due to a problem with the new charts but rather with what is deployed currently there. I will fix this separately.
Regarding minio being included in renku charts, I did some digging and it might have been a problem of packaging/mistagging the charts, some versions like `0.7.4` have it while other dont. After `0.7.5` when the charts were getting published with the GH action this does not seem to be a problem anymore.

PR #2055, opened 2021-03-30
Fix dead link to helm installation documentation.

dunatotatos:


dunatotatos:
Not sure why it fails. I doubt we can blame the commit ^^

rokroskar:
Hi @dunatotatos definitely not your fault - it's because our CI requires repo write permissions. I will open another PR with your commit to make this go through. Thanks for the correction!

rokroskar:
actually, turns out that pushing the same commit to a different branch does the trick :) So I'll go ahead and approve/merge this PR instead of the other one.

rokroskar (approved):
Thanks @dunatotatos for fixing this!

PR #2060, opened 2021-04-06
docs: update the templates page

lorenzo-cavazzi:
The templates page contained outdated information about custom template repositories.
This PR already includes info for the incoming sharable links feature currently available here SwissDataScienceCenter/renku-ui#1288 and ready to be merged


docs/user/templates.rst
```diff
@@ -316,20 +319,174 @@ renku-specific variables that are always available in templates, namely:
...
 **always** specify a tag (or a commit) when creating a new project from a
 custom repository. You can find further details in
 `renku init docs <https://renku-python.readthedocs.io/en/latest/commands.html#use-a-different-template>`_.
 
-If you are using the UI through a RenkuLab instance, you can ask the
-administrators to include your repository in the official repository list
-available in the
+If you are using a RenkuLab instance, you can change the `Template source`
+to ``Custom`` on the project creation page. You will be able to insert
+a URL pointing to your template repository.
```
ciyer, line 333:
```suggestion
If you are using a RenkuLab instance, you can use a custom template repository by
changing the `Template source` to ``Custom`` on the project creation page. There, 
you will be able to insert a URL pointing to your template repository.
```

docs/user/templates.rst
```diff
@@ -316,20 +319,174 @@ renku-specific variables that are always available in templates, namely:
...
+
+  .. image:: ../_static/images/templates_url_error.png
+    :width: 100%
+    :align: center
+    :alt: Error fetching custom templates
+
+If you think your template may be useful for the broader community, you can
+have more visibility by including it in the
+`community-contributed project templates repository <https://github.com/SwissDataScienceCenter/contributed-project-templates>`_.
+Feel free to push a pull request and we will validate it.
```
ciyer, line 368:
```suggestion
Feel free to open a pull request and we will validate it.
```

docs/user/templates.rst
```diff
@@ -316,20 +319,174 @@ renku-specific variables that are always available in templates, namely:
...
+
+Most of the time, you want your repository to be immediately available
+to the target audience, maybe even having some variables set to specific
+values. We know manually entering the same values, again and again, is a
+tedious process, so we come up with a solution to automatically pre-fill
+some of the fields in the project creation page.
+
+You can create a link containing metadata embedded in a query parameter.
+Once you enter that link in a browser, the UI will automatically fetch
+the required data and pre-fill the fields for you.
```
ciyer, line 387:
```suggestion
There are times where you may have many users that each need to create their own
repository, but some of the parameters are shared across all the repositories. An example
could be in a course where each student needs their own repository, but all should use the
same template and metadata.

For this case, it is possible to generate a project-creation link containing 
metadata embedded in a query parameter. When that link is entered in a 
web browser, the UI will  automatically fetch the required information and 
pre-fill the fields.
```

docs/user/templates.rst
```diff
@@ -316,20 +319,174 @@ renku-specific variables that are always available in templates, namely:
...
+
+To create a shareable link, start by filling in all the fields as you would
+do when creating a new project. Instead of clicking on `Create project`,
+click on the dropdown on the right side of the same button and then on
+`Create link`.
+
+You should see a modal where you can select which fields to include in the
+metadata. Some may not be clickable (no value provided), and others are
+deselected by default. The URL updates in real-time, and you can copy it
+for future reference or share it with others.
```
ciyer, line 397:
```suggestion
You should see a modal dialog where you can select which fields to include in the
metadata. Some may not be clickable (no value provided), and others are
deselected by default. The URL updates in real-time and can be copied to
share with others.
```

docs/user/templates.rst
```diff
@@ -316,20 +319,174 @@ renku-specific variables that are always available in templates, namely:
...
+  restrictive one.
+  
+  It would be best to prefer fixed references for custom template
+  repositories, especially when selecting a template and providing values for
+  variables. This means commits and tags are a good choice, while branches
+  are not. Otherwise, the template or the variables may change in a later
+  version, resulting in a corrupted link.
+
+
+Programmatically compute shareable links
```
ciyer, line 420: This documentation is great and easy to understand, but I wonder if our user documentation is the best place to have it. Would it maybe be better to put it on Discourse and just say:

```
If you need to generate project-creation links yourself, you can see an example of how to do this in [Discourse](link to post).
```

docs/user/templates.rst
```diff
@@ -316,20 +319,174 @@ renku-specific variables that are always available in templates, namely:
...
+If you are working in a dedicated RenkuLab deployment and your local
+community needs the templates, you should contact the administrators to
+include your repository in the RenkuLab template source through the
 `renku-values file <https://renku.readthedocs.io/en/latest/admin/index.html#create-a-renku-values-yaml-file>`_.
 
-We are currently working on adding full support for custom template
-repositories in RenkuLab so that you can manually specify external resources
-not included in the configuration file.
+
+Create shareable links with pre-filled fields
```
ciyer, line 376:
```suggestion
Create shareable project-creation links with pre-filled fields
```

ciyer (changes requested):
I'm glad to see these feature getting documented. I have some suggestions for alternative wording. Also, I'm not sure if we want to include detailed instructions for programmers in this documentation: maybe Discourse is a better place for that.

docs/user/templates.rst
lorenzo-cavazzi: Maybe that part is a bit too technical, but I would still include it here since it was explicitly requested by one of the first users of this feature.

lorenzo-cavazzi:
Thanks for taking the time to go through the whole text :pray: 
I ended up applying all the suggested changes -- I had to create a separate commit for a couple of them since they required a few additional bits to comply with `.rst` formatting.

ciyer (approved):
This is ready to roll out!

PR #2061, opened 2021-04-06
chore: include community templates

lorenzo-cavazzi:
Add community templates to the RenkuLab template list.

/deploy


rokroskar (approved):
This looks good to me for the time being - when/if the list of templates grows larger we'll want some other way of presenting them, but that can wait. Thanks @lorenzo-cavazzi !

PR #2064, opened 2021-04-08
Support deploy command for specifying components versions

lorenzo-cavazzi:
Extends the `/ deploy` command with the possibility to add more information on the same string to personalize the deployment. Specifically:

- A reference for any component (e.g. `renku-ui=0.11.8` or `renku-notebooks=debug-with-vscode-k8s`).
- A `#notest` string that falsifies a `test-enabled` output variable, meant to be used to disable tests.

I guess the `check-pr-description` action could have been more flexible. Moving the logic to a python script would simplify improving it, but I think for the moment being the shell script is fine.

/deploy #notest renku-ui=0.11.8 renku-notebooks=debug-with-vscode-k8s


ableuler (approved):
Very nice! It's true that there is potential for some refactoring here later on. Especially the fact that we're hardcoding the list of renku components in more and more places will make it a bit annoying to add further components later on. But for now I'd say let's merge this! 👍 

@jachro With this action it will be straight forward to also test every PR on renku-graph against the heads of all main branches of all the other components.

rokroskar:
Is it obvious which changes the other components need to implement to take advantage of this? If any?

lorenzo-cavazzi:
@rokroskar this is an example for the UI https://github.com/SwissDataScienceCenter/renku-ui/pull/1302
It should be easy to adapt it to the other repos. I can also do that

rokroskar:
great, thanks for doing that - I was wondering if we should add a doc somewhere but it's probably fine like this

PR #2066, opened 2021-04-08
docs: describe RenkuLab homepage customization

ciyer:
Changes to the `values.yaml` file and documentation to support homepage customization. 

Do not merge until https://github.com/SwissDataScienceCenter/renku-ui/pull/1274 is merged to master.

### DIFF
--- a/docs/admin/homepage.rst
+++ b/docs/admin/homepage.rst
@@ -1,0 +1,85 @@
+.. _admin_homepage:
+
+Configuring the RenkuLab Homepage
+---------------------------------
+
+Homepage
+~~~~~~~~
+
+The homepage is shown at the root URL of a RenkuLab instance to a non-logged-in visitor.
+(A logged-in visitor sees the ``Welcome`` page.)
+By default, this page shows an explanation of the features of RenkuLab, but its content
+is configurable by altering the ``renku/values.yaml`` file.
+
+Configuring the content
+~~~~~~~~~~~~~~~~~~~~~~~
+
+There two pieces of information that most deployments will want to customize: the
+link to the tutorial and the list of projects to highlight on the homepage.
+
+ui.homepage.tutorialLink
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+``ui.homepage.tutorialLink`` contains the URL that is used when a user navigates to
+the tutorial. The default is to take the user to the tutorial in the documentaion:
+https://renku.readthedocs.io/en/latest/tutorials/01_firststeps.html. If your
+deployment has a custom tutorial, for example, one hosted on the deployment itself,
+you can replace the link to the tutorial.
+
+ui.homepage.projects
+^^^^^^^^^^^^^^^^^^^^
+
+A deployment may want to highlight certain projects that are hosted on it.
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi (commented):
While reading the docs, I've found a couple of details _possibly_ to adjust -- not sure though since my English is not great. Disregard them if you think it's already correct.

docs/admin/homepage.rst
```diff
@@ -0,0 +1,85 @@
+.. _admin_homepage:
+
+Configuring the RenkuLab Homepage
+---------------------------------
+
+Homepage
+~~~~~~~~
+
+The homepage is shown at the root URL of a RenkuLab instance to a non-logged-in visitor.
+(A logged-in visitor sees the ``Welcome`` page.)
```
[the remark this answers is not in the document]
ciyer: This is a matter of taste. Both are correct. I accepted the other correction fixing the missing verb.

helm-chart/renku/values.yaml
```diff
@@ -460,6 +460,22 @@ ui:
     ## is available under https://5bcn9bqff4qt.statuspage.io/api/v2/summary.json, so
     # id: 5bcn9bqff4qt
 
+  ## Home settings allow configuration of the home page shown on RenkuLab when the user
+  ## is not logged in.
+  # homepage:
+  #   custom:
+  #     enabled: true
+  #     main:
+  #       contentMd: |
```
ableuler, line 469:
```suggestion
  #       # Custom homepage content as single string markdown
  #       contentMd: |
```

helm-chart/renku/values.yaml
```diff
@@ -460,6 +460,22 @@ ui:
...
+  #         ## RenkuLab @ Yoyodyne
+  #         Welcome to the **Yoyodyne** RenkuLab instance!
+  #       backgroundImage:
+  #         url: https://eoimages.gsfc.nasa.gov/images/imagerecords/79000/79803/earth_night_rotate_lrg.jpg
+  #   tutorialLink: https://renku.readthedocs.io/en/latest/tutorials/01_firststeps.html
```
ableuler, line 474: Do we expect users to modify this? Can't it just be set automatically to https://renku.readthedocs.io/en/<renku-version>/tutorials/01_firststeps.html through chartpress?

helm-chart/values.yaml.changelog.md
```diff
@@ -7,6 +7,10 @@ Please follow this convention when adding a new row
 
 ----
 
+## Upgrading to Renku 0.7.10
```
ableuler, line 10: Thanks for adding this line already. 0.7.10 is already past, os I guess we should just add **unreleased** as title here and modify it when we do a release.

ableuler (changes requested):

ableuler:
Just a few minor comments from my side, makes a lot of sense otherwise.

helm-chart/renku/values.yaml
ciyer: I personally would prefer to link to a tutorial that is a project on the RenkuLab instance, rather than link to a tutorial on an external site. In that case, it needs to be configurable.

PR #2088, opened 2021-04-22
chore: add acceptance test for creating a dataset with file

Panaetius:
/deploy


ableuler (commented):
Definitely very good to test this functionality. Two things:
 - From looking at the test logs of this ci-deployment it seems to me that this scenario was not executed. Are things wired together correctly?
 - We're already creating quite a few datasets during the tests, so I wondering if it's better to test this functionality by adding to an existing scenario or by creating a new one. 

Maybe @jachro can have a look at this too?

PR #2091, opened 2021-04-24
chore: add a swagger UI

rokroskar:
This PR adds a swagger UI to the renku deployment to allow testing of API endpoints directly in a live deployment. 

It configures the core and notebooks service specs to be shown - each can be selected via a drop-down. 

### Authentication

Some endpoints (e.g. those of the core service) receive tokens via the gateway. The core service api spec is configured for login via Keycloak - this will retrieve a JWT token to inject into the requests that get routed through the gateway. You must also log in to the renku instance you are testing so that the gateway there has actual tokens to send (the login in swagger doesn't do this for you). For endpoints that do not go through the gateway the spec should define the proper authorization methods (e.g. an API token). 

You can access the api docs under https://ci-renku-2091.dev.renku.ch/swagger

/deploy renku-core=@2029-api-docs #notest

### DIFF
--- a/helm-chart/renku/templates/swagger.yaml
+++ b/helm-chart/renku/templates/swagger.yaml
@@ -1,9 +1,10 @@
+{{- if .Values.swagger.enabled }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "renku.fullname" . }}-swagger
 spec:
-  replicas: {{ .Values.replicaCount }}
+  replicas: 1
   strategy:
     type: Recreate
   selector:
@@ -48,20 +49,6 @@ spec:
             httpGet:
               path: /swagger
               port: http
-          resources:
-            {{- toYaml .Values.resources | nindent 12 }}
-      {{- with .Values.nodeSelector }}
-      nodeSelector:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
-    {{- with .Values.affinity }}
-      affinity:
-        {{- toYaml . | nindent 8 }}
-    {{- end }}
-    {{- with .Values.tolerations }}
-      tolerations:
-        {{- toYaml . | nindent 8 }}
-    {{- end }}
 ---
 apiVersion: v1
 kind: Service
@@ -77,3 +64,4 @@ spec:
   selector:
     app.kubernetes.io/name: {{ include "renku.name" . }}-swagger
     app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
--- a/helm-chart/renku/templates/swagger.yaml
+++ b/helm-chart/renku/templates/swagger.yaml
@@ -1,0 +1,69 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "renku.fullname" . }}-swagger
+spec:
+  replicas: {{ .Values.replicaCount }}
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "renku.name" . }}-swagger
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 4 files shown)


aledegano:
Very cool idea!
Unfortunately it would seem that there's an issue with the Notebook APIs: 
notebook_swagger_error

rokroskar:
@aledegano yes I need to follow up with @olevski about that - I added it as a placeholder for now basically. AFAIK it should work, but in any case I'd like to make it like the core API page where the auth goes through the gateway so you just need to authenticate with keycloak

helm-chart/renku/templates/swagger.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "renku.fullname" . }}-swagger
+spec:
+  replicas: {{ .Values.replicaCount }}
```
ableuler, line 6:
```suggestion
  replicas: {{ .Values.swagger.replicaCount }}
```

helm-chart/renku/templates/swagger.yaml
```diff
@@ -0,0 +1,73 @@
...
+            httpGet:
+              path: /swagger
+              port: http
+          readinessProbe:
+            httpGet:
+              path: /swagger
+              port: http
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+      {{- with .Values.nodeSelector }}
```
ableuler, line 47:
```suggestion
      {{- with .Values.swagger.nodeSelector }}
```
same for the values further down.

ableuler (changes requested):
Really nice to finally have this. I think it's justified to create issues in the repos that don't have an openAPI spec yet.

two points:
- please create a small swagger section in the renku values file and set appropriate resource requests / limits.
- is there a way of injecting renku client ID and secret into the swagger deployment? That would be super convenient and safer as we wouldn't have to copy paste application secrets around.

rokroskar:
Thanks for the review @ableuler!

Good point on automatically configuring the oauth client - turns out it's shockingly simple: https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/oauth2.md

I tried setting those two environment variables now and it seems to work - afaict it doesn't expose the secret to the user, but this needs to be understood carefully.

rokroskar:
Actually that doesn't work - somehow the secret needs to be sent to the user's browser of course... so it's exposed. We probably don't want that.

rokroskar:
btw @aledegano the notebooks API shows up now...

aledegano:
> btw @aledegano the notebooks API shows up now...

Great! Confirm that!

ableuler:
ok, I see. So swagger can mimic a confidential client, but not really act as one. I see two options:
- we enable swagger (preconfigured with client id + secret) on dev deployments only (disabled per default)
- we use the pixie (PKCE) flow for non-confidential clients. it looks like both keycloak and swagger support this off the shelf.

rokroskar:
I don't see PKCE as an option from the oidc config that swagger fetches - or does it go by some other name there?

ableuler:
Looks like it's disabled by default: https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/oauth2.md#oauth-20-configuration

helm-chart/renku/templates/swagger.yaml
rokroskar: I removed these options, they won't really be needed

helm-chart/renku/templates/swagger.yaml
rokroskar: I hard-coded this to 1

rokroskar:
@ableuler I've configured the extra keycloak client as we discussed, implementing the PKCE flow. I've also added a configuration section that allows you to turn swagger off if needed.

ableuler (approved):
LGTM!

PR #2096, opened 2021-04-26
feat(chart): add renku-cli client to keycloak

ableuler:
/deploy

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -45,7 +45,10 @@ global:
       value:
       overwriteOnHelmUpgrade: false
   gateway:
+    ## The client secret for the renku client application registered in keycloak.
     clientSecret: # use `uuidgen -r`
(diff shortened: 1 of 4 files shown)


helm-chart/renku/values.yaml
```diff
@@ -45,7 +45,10 @@ global:
       value:
       overwriteOnHelmUpgrade: false
   gateway:
+    ## The client secret for the renku client application registered in keycloak.
     clientSecret: # use `uuidgen -r`
+    ## The client secret for the renku-cli client application registered in keycloak.
+    cliClientSecret: some-dummy-secret # use `uuidgen -r`
```
mohammad-sdsc, line 51: Is it ok to have this dummy value in this file? Or should we add it to example configuration files?

ableuler: No, it shouldn't be here - thanks for checking.

mohammad-sdsc (approved):
Thank you!

PR #2099, opened 2021-04-27
refactor: tests to use given CLI version

### LINKED ISSUE #2095: Install renku-python as part of test execution
We are currently including the latest renku-python version (head of master) in the acceptance test image. This can lead to inconsistencies with the version of renku-core. We should exclude renku-python from the acceptance test image all together and instead pip install the appropriate version of renku python from pypi during the test execution itself. The version to be installed should be passed in as a an argument to the tests / read from an env variable. In absence of this argument, the "latest_version" for this instance published on `http(s)://<renku-host>/api/renku/version` should be picked.

jachro:
closes #2095 
/deploy


rokroskar:
I'm confused - we went through the trouble of taking this sort of setup out of the code before because it made it more difficult to understand and debug errors arising from conflicting versions. That was the reason for building the renku version into the image. What has changed that will avoid the problems this approach caused before?

jachro:
To be honest I've stopped following the long discussion that was going on these days on Slack but I think in the end we got more or less what we had before with the difference that we don't stick the master commit version of CLI to the tests but rather we take the version Core is running on a particular env and use it in the tests.

jachro:
I think it's a bit safer than before when we were for instance testing release candidate with a fixed version of CLI, we wanted to use that particular version in the tests rather than the CLI master.

ableuler:
> I'm confused - we went through the trouble of taking this sort of setup out of the code before because it made it more difficult to understand and debug errors arising from conflicting versions. That was the reason for building the renku version into the image. What has changed that will avoid the problems this approach caused before?

I might be ignorant about some aspects of how this used to work in the past. All we're trying to do here is make sure that the renku-cli version used in the test matches the renku-python version of the core service. This is currently not the case.

rokroskar:
A better approach would be to read the version from the requirements file, which is what used to be done in the CI script. Because if you are trying to do an acceptance test on a non-released version now you will get e.g. `version: 0.15.6dev9` - `pip` won't really know what to do with that.

rokroskar:
@SwissDataScienceCenter/renku-python-maintainers maybe we can have the commit-sha part of the version string? Is there any reason not to?

ableuler (approved):
LGTM! Thanks for the quick implementation!

jachro:
You're most welcome 🙂

PR #2106, opened 2021-04-29
fix: waiting for redirect frequency tweaked

jachro:
/deploy

### DIFF
--- a/acceptance-tests/src/test/scala/ch/renku/acceptancetests/tooling/Grammar.scala
+++ b/acceptance-tests/src/test/scala/ch/renku/acceptancetests/tooling/Grammar.scala
@@ -105,15 +105,16 @@ trait Grammar extends WebElementOps with Eventually {
 
     @scala.annotation.tailrec
     def asLongAsBrowserAt[Url <: BaseUrl](page: Page[Url], attempt: Int = 1)(implicit baseUrl: Url): Unit = {
-      val maxAttempts = 120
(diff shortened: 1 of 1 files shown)


acceptance-tests/src/test/scala/ch/renku/acceptancetests/tooling/Grammar.scala
```diff
@@ -105,12 +105,10 @@ trait Grammar extends WebElementOps with Eventually {
 
     @scala.annotation.tailrec
     def asLongAsBrowserAt[Url <: BaseUrl](page: Page[Url], attempt: Int = 1)(implicit baseUrl: Url): Unit = {
-      val frequencyFactor = 12
-      val maxAttempts     = 10 * frequencyFactor
+      val maxAttempts = 120
 
       if (attempt <= maxAttempts && (currentUrl startsWith page.url)) {
-        sleep((patienceConfig.timeout.millisPart / frequencyFactor) millis)
-        reloadPage()
```
ableuler, line 113: I think not reloading the page while waiting for a redirection is actually better 👍

ableuler (changes requested):
Small detail: I think your waiting indication is still false:
`${((patienceConfig.timeout.millisPart millis) * attempt).toSeconds}s`
After your changes in this PR I believe it's simply the number of requests in seconds.

ableuler (approved):

PR #2107, opened 2021-04-30
feat(test): use session PVCs in acceptance tests

olevski:
closes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/588

/deploy

### DIFF
--- a/actions/deploy-renku/entrypoint.sh
+++ b/actions/deploy-renku/entrypoint.sh
@@ -15,6 +15,21 @@ printf "%s" "$RENKU_VALUES" | sed "s/<replace>/${RENKU_RELEASE}/" > $RENKU_VALUE
 # enable anonymous notebooks
 yq w -i $RENKU_VALUES_FILE "global.anonymousSessions.enabled" "true"
 
+# enable user session PVCs
+yq w -i $RENKU_VALUES_FILE "notebooks.userSessionPersistentVolumes.enabled" "true"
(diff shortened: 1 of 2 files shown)


rokroskar:
I think the issue is more RBAC than the values themselves - we need to allow the renku-bot user to make PVs. Once that's done, these values can be added to the same values that all deployments use.

olevski:
> I think the issue is more RBAC than the values themselves - we need to allow the renku-bot user to make PVs. Once that's done, these values can be added to the same values that all deployments use.

I agree Rok. I figured out what kind of role we need to create and assign to renkubot in Rancher to make this work. Now just waiting for the tests to finish - the deployment succeeded. And I just launched a session from this deployment to confirm that PVs are indeed enabled and used.

rokroskar:
Awesome! 🚀

olevski:
So I tested out not having a clusterrole and a clusterrole binding for renku notebooks for the user sessions PVs. This is what @ableuler suggested in the issue originally and everything works fine because the storage plugin takes care of creating the pv once the renku notebooks service creates a pvc.

**NOTE**: This should only be merged after the following things are done:
- renku-notebooks 0.8.13 is released which will include this change required for the proper deployment of renku in acceptance tests (SwissDataScienceCenter/renku-notebooks#641)
- This PR is merged: https://github.com/SwissDataScienceCenter/ops-dev.renku.ch/pull/157

PR #2118, opened 2021-05-07
feat: endpoint to forward gitlab traffic through gateway

### LINKED ISSUE #2116: An endpoint to access gitlab with keycloak access token
As a Renku API user, I want to have an endpoint to access private gitlab repositories so that I can use a keycloak (renku) access token to authenticate with gitlab.

**Additional context**
This is a continuation of `renku login` in the CLI. It can now receive a keycloak token from the gateway and use it to access secured KG endpoints. Now, we want to use the same token to access private gitlab repos without setting up an SSH key.

In some renku deployments, a `/gitlab` endpoint exists which forwards requests directly to gitlab. A Keycloak access token cannot be used to authenticate with this endpoint because requests to this endpoint don't pass through the gateway. On the other hand, we don't want to forward all `/gitlab` traffic through the gateway because it's not efficient.

**Acceptance criteria**
- A `/repo` endpoint is added to renku
- Requests passed to this endpoint are forward to the gateway

mohammad-sdsc:
Adds a `/repo` endpoint to allow authenticating gitlab requests using renku token

Fixes #2116


helm-chart/renku/templates/ingress.yaml
```diff
@@ -62,6 +62,10 @@ spec:
           serviceName: {{ $gitlabFullname }}
           servicePort: {{ $gitlabServicePort }}
       {{- end }}
+      - path: /repo
```
ableuler, line 65: Should we call it `/repos`? We use the plural form in most or all our URLs elsewhere.

ableuler (approved):
👍

PR #2127, opened 2021-05-17
fix: patch docker command for running sessions locally

gavin-k-lee:
Small patch to add `ip` argument to the `jupyter lab` command when running sessions locally.

Especially relevant for RStudio users.


rokroskar (approved):
thanks!!

PR #2132, opened 2021-05-20
Configure the Renku realm to use the Renku keycloak theme

### LINKED ISSUE #2022: Enable custom theme on keycloak
Following the keycloak upgrade #1900 the custom theme for keycloak screen is not working properly, we need to enable/fix this.

ciyer:
Fix #2022

/deploy

### DIFF
--- a/acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
+++ b/acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
@@ -77,7 +77,7 @@ object LoginPage
   }
 
   def logInButton(implicit webDriver: WebDriver): WebElement = eventually {
-    find(cssSelector(".btn.btn-primary.btn-block.btn-lg")) getOrElse fail("Log In button not found")
+    find(cssSelector("#kc-login")) getOrElse fail("Log In button not found")
   }
 }
 
--- a/actions/deploy-renku/Dockerfile
+++ b/actions/deploy-renku/Dockerfile
@@ -7,7 +7,8 @@ RUN apk add python3 docker jq && \
     pip3 install -U pip \
     chartpress==1.0.0 \
     kubernetes \
-    ruamel.yaml
+    'ruamel.yaml<0.17.10' \
+    'ruamel.yaml.clib<0.2.4'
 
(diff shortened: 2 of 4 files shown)


ciyer:
Switching to the new renku-theme **fixes** the "internal server error" problem that was occurring before. The change to `init-realm.py` **does not** automatically activate the renku theme (even though that works locally). ATM, an admin needs to switch the theme to get the renku theme.

rokroskar:
@ciyer what's the status here? (it's still a draft) 

Where was the "internal server error" occurring?

ciyer:
> Where was the "internal server error" occurring?

The old Keycloak theme would cause an "internal server error" when you tried to get to the log in screen.

rokroskar:
can we adjust the acceptance test here as well? 

20210622090751-LoginSpec

ciyer:
> can we adjust the acceptance test here as well?

The Renku login page is not activated by default. A Keycloak admin needs to actively turn it on, so the tests do not need to be modified.

rokroskar:
hmm... then the test needs to work with both cases - we need to be able to run the acceptance tests in production as well

rokroskar:
so you enabled it by hand for the deployment in this PR?

ciyer:
> so you enabled it by hand for the deployment in this PR?

Yes.

ableuler:
> > so you enabled it by hand for the deployment in this PR?
> 
> Yes.

Are you sure that's true? It seems to me that the CI deployment is older than the commit which adds this:

image

The realm is only created at first deploy, but on a fresh redeploy I'd expect the realm to be created with the right template.

ciyer:
> hmm... then the test needs to work with both cases - we need to be able to run the acceptance tests in production as well

I have added a commit that fix the acceptance tests to work with both login screens, the keycloak default and the Renku screen.

ableuler:
I deleted everything in those namespaces (except the TLS secret) and redeployed, now the renku-theme was activated from the start.

ableuler:
> can we adjust the acceptance test here as well?
> 
> 20210622090751-LoginSpec

Did you test this on a keycloak instance which has upstream identity providers configured?

rokroskar (approved):
Thanks @ciyer!!

ciyer:
> Did you test this on a keycloak instance which has upstream identity providers configured?

Yes, that also works.

PR #2140, opened 2021-05-28
feat: acceptance-tests to work with a given personal access token

jachro:
/deploy


rokroskar:
Should we make a follow-up issue to use this in our CI scripts?

jachro:
We could but as we've got everything set up for the `credential way` it's
not maybe that necessary. I mean we won't gain anything by switching. When
I was adding the access token option I was thinking more about DEVs.

PR #2142, opened 2021-05-28
Use Amalthea in renku-notebooks

olevski:
/deploy


ableuler:
/deploy renku-notebooks=k8s-operator-poc

@olevski is that already expected to work? anyway... I'm feeling lucky :)

olevski:
> > /deploy renku-notebooks=k8s-operator-poc
> 
> @olevski is that already expected to work? anyway... I'm feeling lucky :)

It will not work because I am pretty sure that the `values.yaml` file generated by the action that deploys PRs is incompatible with the changes required by the operator. But you are welcome to try it.

rokroskar:
you could insert a new secret with the correct values _or_ modify the action for this PR _or_ if it's just a few values you need to add you can use the `extra-values` argument to `/deploy`

ableuler:
ok, quickly trying to deploy this already went pretty far. What's still missing here are some adaptations to the script which creates the renku realm in order to create the necessary client.

ableuler:
Also, I think it would be nice if the default scope of amalthea wasn't the `default` namespace but the namespace that amalthea itself is deployed in. This seems a more reasonable default, not just for the Renku case.

ableuler:
See https://github.com/SwissDataScienceCenter/renku/issues/2184 and https://github.com/SwissDataScienceCenter/amalthea/issues/40.

olevski:
Ok this is ready for review. The other test that persistently failed is now passing. Now there is a new test that fails that never failed before related to dataset search functionality. However this has nothing to do with the notebook service at all. So I think it is just a flaky test. Also this test was passing in earlier reruns of the same or very recent commits.

helm-chart/renku/templates/ingress.yaml
```diff
@@ -68,8 +68,8 @@ spec:
           servicePort: {{ $gatewayServicePort }}
       - path: /jupyterhub
```
ableuler, line 69: I think when we merge this PR jupyterhub will actually be completely gone.

ableuler (commented):
Can you please also remove all traces in the admin docs about installing the anonymous sessions extension? https://github.com/SwissDataScienceCenter/renku/blob/master/docs/admin/anonymous-sessions.rst

olevski:
Ok the problem with the deployment not working was because we changed the host IP on which the werkzeug server for the git sidecar in renku notebooks was listening to from `0.0.0.0` to `127.0.0.1`. This somehow made the readiness/liveness probes for gitsidecar because the connection to the git sidecar was refused. I tried to put the host in the readiness/liveness probes but this did not help. Only switching the host ip back to `0.0.0.0` seems to work.

ableuler (changes requested):
The acceptance test for launching a private session seems to be consistently failing with 
image

I believe the reason for this is that after opening the tab there's first a login loop to the oidc provider happening, so the window named `JupyterLab` doen't show up right after opening the tab.

ableuler (dismissed):

rokroskar:
Aren't we making a release of Renku soon? We probably want to wait with this PR until that release is done?

ableuler:
> Aren't we making a release of Renku soon? We probably want to wait with this PR until that release is done?

Renku release is out, let's goooo! 🚀

ableuler (approved):

PR #2166, opened 2021-06-20
feat(docs): new design for renku docs

vfried:
New design for docs

image

/deploy


rokroskar:
This looks really great!! Do you think it would be possible to apply these changes to the [renku sphinx theme](https://github.com/SwissDataScienceCenter/renku-sphinx-theme)? That way the other repos that might publish their own documentation pages can use the same theme

vfried:
Yes, that's a good point... i spoke with ralph this morning to apply this to renku python... at the moment we are not using the renku theme, we are using the read the docs one... https://github.com/readthedocs/sphinx_rtd_theme we could fork these theme and apply this design there.

rokroskar:
I know we're not using it now, but we used to use it so the repo is already there. It's fine if it's completely overwritten, but having it there would make it easier to maintain.

vfried:
So... shall i fork the RDD docs and apply this there?

vfried:
I will see how i can overwrite the theme in the renku repo then.

rokroskar:
I _think_ you should be able to copy over the static directory into the theme [here](https://github.com/SwissDataScienceCenter/renku-sphinx-theme/tree/master/renku_sphinx_theme/static) and set the RTD theme as the base [here](https://github.com/SwissDataScienceCenter/renku-sphinx-theme/blob/master/renku_sphinx_theme/theme.conf#L2)

ciyer (changes requested):
This looks very nice! 🎉 

The suggestions Rok made make sense to me as well. As part of that conversion, I found a couple of files that should probably be removed (see comments), and, as we discussed, I think it would be nice if the css were autogenerated from the scss files (which might require changes to the config and the makefile), but if that is complicated to do, it should be at least documented that the css files should not be edited and how they can be generated.

I will do a full review after the restructuring, but I have already looked at the changes, so I do not expect anything major -- it looks very good!

vfried:
Changes are now in the renku-sphinx repo.

For testing:
image

ciyer (approved):
👍

PR #2171, opened 2021-06-24
chore(docs): update git instructions

rokroskar:
The instructions for how to save and push back changes pre-dated `renku save`...


ciyer (approved):
👍

PR #2177, opened 2021-06-29
Minimal header tests

ciyer:
Adapt tests for https://github.com/SwissDataScienceCenter/renku-ui/pull/1385

PR #2178, opened 2021-06-30
fix: css for the Login button on the provider page

jachro:
/deploy

### DIFF
--- a/acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
+++ b/acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
@@ -110,7 +110,9 @@ object ProviderLoginPage
   }
 
   def logInButton(implicit webDriver: WebDriver): WebElement = eventually {
-    find(cssSelector(".btn.btn-primary.btn-block.btn-lg")) getOrElse fail("Log In button not found")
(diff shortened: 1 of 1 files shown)


joke1196 (approved):

acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/LoginPage.scala
```diff
@@ -110,7 +110,9 @@ object ProviderLoginPage
   }
 
   def logInButton(implicit webDriver: WebDriver): WebElement = eventually {
-    find(cssSelector(".btn.btn-primary.btn-block.btn-lg")) getOrElse fail("Log In button not found")
+    find(cssSelector(".btn.btn-primary.btn-block.btn-lg"))
```
ciyer, line 113: I think the better solution is the one here:

https://github.com/SwissDataScienceCenter/renku/pull/2177/commits/9905efaf7387c0016a2de09b59646d863b098fb6

Using the id will work for both versions of the page (Renku styled and un-styled), so it is unnecessary to support both versions.

jachro: @ciyer I wasn't aware of the change. I guess you can maybe replace our change with your version when merging.

PR #2179, opened 2021-06-30
chore: move actions to their own repo

ableuler:
Actions have been moved to their own repository and a first release has been tagged there. See https://github.com/SwissDataScienceCenter/renku-actions


rokroskar (changes requested):
the use of these actions in the renku repo needs to be adapted as well :)

ableuler:
> the use of these actions in the renku repo needs to be adapted as well :)

yeah, I realized this right after opening the PR and added a comment...

.github/workflows/deploy.yml
```diff
@@ -2,87 +2,86 @@ name: Deploy and Test
...
 on:
   push:
-    branches: ['master']
+    branches: ["master"]
     paths:
-    - helm-chart/**
-    - scripts/**
-    - acceptance-tests/**
-    - .github/**
+      - helm-chart/**
```
ableuler, line 7: vscode - at least for me - has some strong ideas about how YAML should be formatted. Do we actually have some rules defined for this?

ableuler:
PR is ready, giving it a few more days for people who are relying `@master` versions of actions in this repo to adapt.

rokroskar (dismissed):

PR #2183, opened 2021-07-02
Changes for the iframe-based collaboration page

ciyer:


acceptance-tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
```diff
@@ -45,38 +50,14 @@ trait Collaboration {
...
-    tf enterValue title
-    click on projectPage.Collaboration.Issues.NewIssue.markdownSwitch sleep (1 second)
-    projectPage.Collaboration.Issues.NewIssue.descriptionField enterValue description
-    click on projectPage.Collaboration.Issues.NewIssue.createIssueButton sleep (1 second)
+  def `verify that the GitLab issues page link is visible`: Unit = {
+    Then("they should see a link to the GitLab issues page")
+    verify userCanSee projectPage.Collaboration.Issues.gitLabIssuesLink
   }
 
   def `verify branch was added`: Unit = {
```
vfried, line 63: I see you left this here... will we show the banner for merge requests inside renku? otherwise we can maybe delete this code.

vfried (commented):
Things look good to me, there is a small comment but after you answer that i can aprove.

acceptance-tests/src/test/scala/ch/renku/acceptancetests/workflows/Collaboration.scala
ciyer: You are right! I have removed this method since the banner is gone.

vfried (approved):
Since we merged the UI stuff to master, shouldn't we merge the UI branch to master here as well?

ciyer:
> Since we merged the UI stuff to master, shouldn't we merge the UI branch to master here as well?

We cannot because the UI is deployed as a canary -- the default UI is still the old UI.

PR #2185, opened 2021-07-05
Update Dockerfile

travissarles:
Bintray has been discontinued so we're replacing it with scala-sbt


travissarles:
Duplicates: https://github.com/SwissDataScienceCenter/renku/pull/2186

PR #2190, opened 2021-07-07
feat: add new client for amalthea

### LINKED ISSUE #2184: Add JupyterServer client to Renku realm in keycloak post-install script
The keycloak post-install script has to create a separate client in the Renku realm for the JupyerServers run by Amalthea.

olevski:
closes #2184 

Note: This will be merged into the operator POC branch - not master.

I think to enable the automated acceptance tests to run with the POC I need to change a secret that may affect all other deployments. So I will do this in a separate PR.

For the time being this, with the poc branches on the gateway and notebooks, as well as the latest version of amalthea are deployed at https://tasko.dev.renku.ch.

/deploy renku-notebooks=k8s-operator-poc renku-gateway=notebooks-operator-poc


ableuler (changes requested):
Nice, quite a bit of configuration and hacky code going away. Can you also adapt the [script to create a values file](https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/example-configurations/make-values.sh)?

helm-chart/renku/values.yaml
```diff
@@ -53,6 +44,9 @@ global:
     gitlabClientId: renku-ui
     ## The client secret for authentication against gitlab
     gitlabClientSecret:
+    ## The secret for the keycloak client that will be used by amalthea
```
ableuler, line 47: I'd put this into `.Values.notebooks.jupyterServerClientId` and `.Values.notebooks.jupyterServerClientSecret` respectively. The init-realm script can pick it up from there as well so it doesn't really need to be a global value and the gateway is not involved here.

ableuler (changes requested):

ableuler:
> I think to enable the automated acceptance tests to run with the POC I need to change a secret that may affect all other deployments. So I will do this in a separate PR.

For the sake of testing, can you just provide this secret as an extra value [here](https://github.com/SwissDataScienceCenter/renku/pull/2142#issue-656764844)?

ableuler (approved):

PR #2207, opened 2021-07-20
(fix) filename used for the archive containing the artifacts of the tests for Staging

aledegano:


.github/workflows/publish-deploy-staging.yml
```diff
@@ -102,7 +102,7 @@ jobs:
         uses: ./actions/download-test-artifacts
         env:
           RENKU_VALUES: ${{ secrets.STAGING_VALUES_FILE }}
-          TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
+          TEST_ARTIFACTS_PATH: "staging-test-artifacts"
```
olevski, line 105: @aledegano doesn't this change require an additional change where these artifacts are created? How do you know the same name is used when the artifacts are created? You are probably not wrong - I am just confused most likely. Just want to make sure though.

aledegano: Yeah, the problem is that the artifact are already created with this name (which is hardcoded in the values file for staging), so the "wrong" name was here and we were never able to fetch the tests artifact because of that.

olevski (approved):

pameladelgado (approved):

PR #2221, opened 2021-07-29
(chore) promote Ingress API version to GA

aledegano:


rokroskar:
do we need to add a note about k8s version compatibility?

aledegano:
> do we need to add a note about k8s version compatibility?

I'm not sure about the format of the notes, can you please tell me if the way I wrote them is good enough?

rokroskar:
Sorry I should have been clearer - I think it should go here: https://github.com/SwissDataScienceCenter/renku/blob/master/docs/admin/index.rst#officially-supported-versions

When we do a release we can add it to the changelog as well.

rokroskar (approved):

PR #2329, opened 2021-10-28
feat: inject custom CA certificates in keycloak

olevski:
Should be reviewed and merged after https://github.com/SwissDataScienceCenter/renku/pull/2330

This adds a library chart that provides yaml snippets for:
- init container to generate ca certs (and optionally inject custom ones)
- volume to store generated certs
- volume mounts to mount the certs in various services

The Dockerfile and script to generate the certs I copied over from the gitlab cloud native helm chart: https://gitlab.com/gitlab-org/build/CNG/-/blob/master/alpine-certificates/scripts/bundle-certificates

In addition I setup the values.yaml portion for this also in a similar way.

I would like to get this merged and approved before the other (child PRs) for the different services are reviewed.

/deploy #persist


chartpress.yaml
```diff
@@ -34,3 +35,17 @@ charts:
         valuesPath: keycloak.initRealm.image
         paths:
           - scripts/init-realm
+      certificates:
+        contextPath: certificates
+        dockerfilePath: certificates/Dockerfile
+        valuesPath: global.certificates.image
+        paths:
+          - certificates
```
ableuler, line 43: I think this should be moved down to the `helm-chart/certificates` chart

olevski: @ableuler I am not sure what you mean? I should move the `Dockerfile` so that it is located at `helm-chart/certificates`?

ableuler (changes requested):
Looks good and seems to work, very minor concerns only from my side.

chartpress.yaml
ableuler: So the chartpress file defines two charts, but most of the specification that should belong to the certificates chart seems to be underneath the renku chart. Anyway, as I later during the review noticed, I think there should be two separate chartpress files anyway.

helm-chart/renku/values.yaml
```diff
@@ -88,7 +101,6 @@ ingress:
   ## Annotations for the created ingress
   annotations:
     ## The ingress class
-    cert-manager.io/cluster-issuer: letsencrypt-production
```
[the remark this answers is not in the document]
olevski: Yes I will remove this change.

chartpress.yaml
```diff
@@ -19,6 +19,7 @@ charts:
       - helm-chart
       - acceptance-tests
       - scripts/init-realm
+      - certificates
```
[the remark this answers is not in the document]
olevski: No it should not. Because when stuff under `certificates` is updated, then this leads to a new certificates image - this is a part of the renku helm chart. Because the certificates image is specified in the `global` section of the renku helm chart.

The certificates library helm chart has no values. It uses whatever is specified under `global.certificates` in any of the other true helm charts that it operates in.

olevski:
@ableuler this is ready for review

ableuler:
Note: this PR should add the valuesPath back which I commented out in https://github.com/SwissDataScienceCenter/renku/pull/2359

olevski:
The acceptance tests are failing because the dataset import step fails: `renku dataset import --name flight-data 10.7910/DVN/WTZS4K` - most likely because the service that provides this dataset is offline.

helm-chart/renku/requirements.yaml
```diff
@@ -36,6 +36,9 @@ dependencies:
   version: 16.0.4
   repository: "https://codecentric.github.io/helm-charts"
   condition: keycloak.enabled
+- name: certificates
+  version: 0.11.0-rc2
```
ableuler, line 40: Should we version the charts independently and tag `0.1.0`?

ableuler (commented):
Looks good, thanks a lot for the effort and sorry for stalling the PR for so long. The only question I'd raise is wether we want to version the library chart independently, even though it "lives" in the Renku repo. What do you think?

helm-chart/renku/requirements.yaml
olevski: @ableuler I am not really sure how we can version the certificates library chart independently if it is part of the same chartpress file as the renku chart. Also even if we used separate chartpress files for the renku chart and for the certificates library chart then we would have to setup github actions that trigger on a specific tag formatted in such a way so that it does not trigger a renku release but it triggers a certificates helm chart release? I was trying to think about all of this but then decided that versioning the certificates helm chart and renku together is the easiest thing to do.

Let me know what you think. And if you do think that they should not be released together then how would you setup the CI for this?

ableuler: The certificates library chart does already have its own chartpress file in the certificates folder, no? I think the easiest would be to not bother about building and publishing the library chart automatically and just run `chartpress --version `X.Y.Z` --publish` manually from the certificates folder whenever we change that chart, and then update the chart dependency in the renku chart requirements.yaml accordingly.

olevski:
> Looks good, thanks a lot for the effort and sorry for stalling the PR for so long. The only question I'd raise is wether we want to version the library chart independently, even though it "lives" in the Renku repo. What do you think?

@ableuler thanks for the review. It does make a lot of sense to version the library chart for certificates independently from renku. I dont foresee that it will change that often. And every time we publish a new version of the library chart we need to update it in all child helm charts. So I will just make sure the chartpress setup makes sense so that this really is the case.

helm-chart/renku/requirements.yaml
olevski: So now the images are built and published manually. Chartpress would not work because it expects the image versions to be updated in the helm chart but the library helm chart has not values file and it crashes chartpress.

If I put the images in the renku helm chart chartpress file then new images for the certificates are built and updated in the values file on every renku release. And I think this is confusing and unnecessary. This kinda pushes us toward the certificates releases being tied to renku - which I really dont want. Ideally (and I think this will be the case) the images for the certificates rarely change.

olevski:
@ableuler this is ready for review btw. The tests have been flaky but as long as renku can be deployed then I consider that that confirms this PR is good to go. This does not really affect launching of sessions or any of the other things the acceptance tests are checking. This just affects the keycloak post install job and keycloak itself.

certificates/README.md
```diff
@@ -4,12 +4,26 @@ This is directly inspired and copied from the Gitlab Cloud Native Helm chart. Se
...
+
+2. Navigate to the `certificates` folder and build the image, tagging with the same tag (i.e. `X.X.X`) as the certificates chart.
+
+```
+docker build -t renku/certificates:X.X.X .
+docker push renku/certificates:X.X.X
+```
+
+3. Publish the certificates chart: `chartpress --publish-chart --tag X.X.X`
+4. Update the tag of the certificates image and the certificates chart in the main Renku Helm chart and all dependent Helm charts
```
ableuler, line 29: I think a description of this manual process is perfectly fine here 👍

ableuler (approved):
Looks good! 👍

PR #2330, opened 2021-10-28
feat: library Helm chart for CA certificates

olevski:
This adds a library chart that provides yaml snippets for:

- init container to generate ca certs (and optionally inject custom ones)
- volume to store generated certs
- volume mounts to mount the certs in various services

The Dockerfile and script to generate the certs I copied over from the gitlab cloud native helm chart: https://gitlab.com/gitlab-org/build/CNG/-/blob/master/alpine-certificates/scripts/bundle-certificates

This should be reviewed, merged and the helm chart for this released before all other PRs can proceed.


certificates/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Taken from https://gitlab.com/gitlab-org/build/CNG/-/blob/master/alpine-certificates/Dockerfile
+ARG FROM_IMAGE=alpine
+ARG ALPINE_VERSION=3.14
+FROM $FROM_IMAGE:$ALPINE_VERSION
+
+ARG CA_PKG_VERSION=20191127-r5
```
ableuler, line 6: Is there something newer to start with?

certificates/README.md
```diff
@@ -0,0 +1,5 @@
+# CA Certificates Handling for Renku
+
+This is directly inspired and copied from the Gitlab Cloud Native Helm chart. See https://gitlab.com/gitlab-org/build/CNG/-/tree/master/alpine-certificates.
+
+The image built from this reop should be added as an `initContainer` to all k8s deployments, pods, statefulsets, etc from the Renku Helm chart.
```
ableuler, line 5:
```suggestion
The image built from this repo should be added as an `initContainer` to all k8s deployments, pods, statefulsets, etc from the Renku Helm chart. 
```

ableuler (changes requested):
This looks good, just two very minor comments. One thing I was wondering though: do you know if library charts can be published on a chart repo? And if yes, is the GitLab one published by any chance?

olevski:
>One thing I was wondering though: do you know if library charts can be published on a chart repo? And if yes, is the GitLab one published by any chance?
@ableuler the gitlab stuff that inserts the certificates is not a library chart: It is just a template part of the top-level gitlab chart: https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/templates/_certificates.tpl

And then it is just used in all children charts like so:
https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/charts/gitlab/charts/gitaly/templates/_statefulset_spec.yaml#L67

certificates/Dockerfile
olevski: @ableuler for `alpine:3.14` that is all there is:
https://pkgs.alpinelinux.org/packages?name=ca-certificates&branch=v3.14

ableuler: Hmm, ok - could you quickly / did you check if a container with root certs injected by this image actually works with letsencrypt certs after their recent change?

ableuler: ie that from inside such a container you can for example resolve renkulab.io?

ableuler:
> I dont see why you could not publish a library chart though. But I think you were hinting at using the gitlab library charts for certificates directly. But the certificate templates are not library charts.

yes, that's why I was asking.

ableuler:
Ah ok, I see [here](https://github.com/SwissDataScienceCenter/renku-notebooks/blob/7a26c5057eb0a2946b937a7761702990f4152d48/helm-chart/renku-notebooks/requirements.yaml#L5) that you actually rely on this library chart being published. Is the publishing of this library chart already part of some github workflow?

ableuler:
> Ah ok, I see here that you actually rely on this library chart being published. Is the publishing of this library chart already part of some github workflow?

Realised that this is part of the subsequent PR, let's merge this.

ableuler (approved):

olevski:
@ableuler this is ready for review.

ableuler (approved):
Looks nice!

PR #2331, opened 2021-10-29
docs: initial draft for updated teaching documentation

gavin-k-lee:
Updating teaching docs, encouraging users to announce via the Google/Microsoft forms.


ksanao (approved):

gavin-k-lee:
@ksanao should we make the repos public then?

PR #2346, opened 2021-11-10
fix urls

sebastianbertoli:


ciyer (approved):
Great, thanks a lot for these corrections!

PR #2362, opened 2021-11-23
Update with JotForm instead of Google or MS

gavin-k-lee:
Use one form instead of two (Google/MS).


ksanao (approved):
The link is functional.
The form works smoothly.

PR #2365, opened 2021-11-28
fix(ci): update github token to org secret

olevski:
I am not sure what is the difference between the `RENKU_CI_TOKEN` and `RENKUBOT_GITHUB_TOKEN` secrets.

But it seems that some of the permissions for `RENKU_CI_TOKEN` have been removed or the token has been fully disabled. Either way it is much nicer to use an org-level secret.

This should probably fix this problem:
https://github.com/SwissDataScienceCenter/renku/runs/4336151543?check_suite_focus=true


ableuler (approved):
Oops, I missed the fact that we hadn't switched this repo yet 🙈 
Thanks @olevski for the fix.

PR #2372, opened 2021-12-01
(fix) Use the same labels for post-install jobs as other resources.

aledegano:
/deploy 
This also fixes the problem where a chart name containing a "+" sign (e.g. when deploying a version containing build metadata) would fail to create the post-install jobs.


ableuler (approved):
Thanks for this fix!

PR #2382, opened 2021-12-06
(feat) Publish Helm chart on all merges to master.

aledegano:
When a PR is merged to the master branch a chart is also published, the version is set to `x.y.z+1-n+1` where `x.y.z` is the latest stable release of Renku and `n` is the latest version created by the job introduced in this PR. If an earlier pre-release version does not exists then it is set to 0.

The workflow also pushes the tag of the just created pre-release, this is necessary for the following workflows to assign the correct, increasing, pre-release number.

The new workflow duplicates somewhat `publish-deploy-staging`, however, at a later time I will take care of consolidating the two, when Staging will also be deployed through Flux.

This supercede #2376.

### DIFF
--- a/.github/workflows/publish-master-merges.yaml
+++ b/.github/workflows/publish-master-merges.yaml
@@ -1,0 +1,56 @@
+name: Publish all merges to master
+
+on:
+  push:
+    branches: ["master"]
+    paths:
+      - helm-chart/**
+      - scripts/**
+      - acceptance-tests/**
+      - .github/**
+  workflow_dispatch:
+
+jobs:
+  publish-chart:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions-ecosystem/action-get-latest-tag@v1
+        id: get-latest-tag
+      - uses: actions-ecosystem/action-bump-semver@v1
+        id: bump-semver
+        with:
+          current_version: ${{ steps.get-latest-tag.outputs.tag }}
+          level: prepatch
+      - id: set-version
+        run: echo "publish_version=${{ steps.bump-semver.outputs.new_version }}.$(echo ${{ github.sha }} | cut -c 1-7)" >> $GITHUB_ENV
+      - uses: SwissDataScienceCenter/renku-actions/publish-chart@v0.3.0
+        env:
+          CHART_DIR: helm-chart/
+          CHART_TAG: "--tag ${{env.publish_version}}"
+          CHART_NAME: renku
+          GIT_USER: renku-bot
+          GIT_EMAIL: renku@datascience.ch
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+          GITHUB_TOKEN: ${{ secrets.RENKUBOT_GITHUB_TOKEN }}
+      - uses: 8398a7/action-slack@v3
+        with:
+          status: custom
+          fields: job,ref
+          custom_payload: |
+            {
+              attachments: [{
+                color: '${{ job.status }}' === 'success' ? 'good' : '${{ job.status }}' === 'failure' ? 'danger' : 'warning',
+                text: `${process.env.AS_JOB} version ${{ env.publish_version }}: ${{ job.status }}.`,
+              }]
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        if: always()
+      - uses: actions-ecosystem/action-push-tag@v1
+        with:
+          tag: ${{ env.publish_version }}
+          message: "${{ env.publish_version }}: PR #${{ github.event.pull_request.number }} ${{ github.event.pull_request.title }}"


.github/workflows/publish-master-merges.yaml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          current_version: ${{ steps.get-latest-tag.outputs.tag }}
+          level: prepatch
+      - uses: SwissDataScienceCenter/renku-actions/publish-chart@v0.3.0
+        env:
+          CHART_DIR: helm-chart/
+          CHART_TAG: "--tag ${{ steps.bump-semver.outputs.new_version }}"
+          CHART_NAME: renku
+          GIT_USER: chart-bot
+          GIT_EMAIL: chart-bot@example.com
```
rokroskar, line 33: why not use the renkubot account?

.github/workflows/publish-master-merges.yaml
```diff
@@ -0,0 +1,54 @@
...
+        id: get-latest-tag
+      - uses: actions-ecosystem/action-bump-semver@v1
+        id: bump-semver
+        with:
+          current_version: ${{ steps.get-latest-tag.outputs.tag }}
+          level: prepatch
+      - uses: SwissDataScienceCenter/renku-actions/publish-chart@v0.3.0
+        env:
+          CHART_DIR: helm-chart/
+          CHART_TAG: "--tag ${{ steps.bump-semver.outputs.new_version }}"
```
rokroskar, line 30: I think it's better to just let chartpress set the default tag - that way it's always identifiable by the commit hash. Our charts used to have dev versions using the `-<commit-sha-7>` suffix - that can also work here I think and it would be consistent with the old charts (not that it's a big requirement in my mind)

aledegano: There are two reasons to not let chartpress do its things:

1. Chartpress does not understand Semver: it uses the **latest** stable tag as base for a new pre-release (e.g. in this case would call it `0.10.3-xxx`, this is wrong because it would tag a pre-release for something that has been already released. This is not only a problem of being pedantic in the use of Semver: since Flux implements correctly the versioning it will never deploy `0.10.3-xxx` because `0.10.3` **is more recent** and has [precedence](https://semver.org/#spec-item-11)
2. By using the commit hash as pre-release version we don't know anymore what is the correct ordering of pre-releases and thus which one to upgrade. It is possible to maintain the hash in the pre-release version by [appending](https://semver.org/#spec-item-10) it with a `+`, e.g. `0.10.4-2+a6a1f62`.

Because of point 2 we also have to push the tag back in this repository so that the workflow knows which one was the latest pre-release published and can bump the new one correctly.

.github/workflows/publish-master-merges.yaml
aledegano: Ah, sure we can do that.

.github/workflows/publish-master-merges.yaml
rokroskar: yeah I see what you mean - the behavior of chartpress is incorrect in this regard, or rather, `git describe --tags` doesn't produce semver-compliant output. 

If I run `git describe --tags` on current master, I get 

```
0.10.3-3-gb696ff7d
```

which means I am three commits after `0.10.3` at commit `b696ff7d` - can't we do something like this, but use proper semver, making this simply `0.10.4-3+gb696ff7d`? That way automation which requires semver for ordering can still use it and we don't need to push this tag back to the repo since it's easy to recreate.

aledegano: yes, I've mentioned it in the reply above :)

rokroskar: right, I was agreeing with you - but I don't think we need to tag every commit to master since it's easy to build that part of the version just from git history.

aledegano: FYI the PR is now ready to review, the final implementation turned out to be `x.y.z-n.<short_commit_hash>` to not have to fiddle too much with Semver manipulation.

.github/workflows/publish-master-merges.yaml
```diff
@@ -0,0 +1,58 @@
...
+        id: get-latest-tag
+      - id: commit
+        uses: pr-mpt/actions-commit-hash@v1
+      - uses: actions-ecosystem/action-bump-semver@v1
+        id: bump-semver
+        with:
+          current_version: ${{ steps.get-latest-tag.outputs.tag }}
+          level: prepatch
+      - id: set-version
+        run: echo "publish_version=${{ steps.bump-semver.outputs.new_version }}.${{steps.commit.outputs.short}}" >> $GITHUB_ENV
```
rokroskar, line 30: you are a github actions search ninja 💪 Though I'm pretty sure something like 

```
echo "publish_version=${{ steps.bump-semver.outputs.new_version }}.$(${{ github.sha }} | cut -c 1-7)" >> $GITHUB_ENV
```

also works, like we do [here](https://github.com/SwissDataScienceCenter/renkulab-docker/blob/master/.github/workflows/build_and_push_to_docker.yml#L24) - this would reduce the number of dependencies

.github/workflows/publish-master-merges.yaml
rokroskar: ping :)

rokroskar (changes requested):
thanks for setting this up! Just one minor change requested to reduce the number of dependencies

.github/workflows/publish-master-merges.yaml
aledegano: done

rokroskar:
Thanks @aledegano - can you make a commit that changes the branch to this one so we can verify that the workflow runs? Then revert that commit when it's done.

aledegano:
> Thanks @aledegano - can you make a commit that changes the trigger branch to this one so we can verify that the workflow runs? Then revert that commit when it's done.

https://github.com/SwissDataScienceCenter/renku/runs/4468084248?check_suite_focus=true

rokroskar:
I'm confused... the docker build for the publish action errors out, but the action works: https://github.com/SwissDataScienceCenter/renku/runs/4468084248?check_suite_focus=true#step:3:115

aledegano:
> I'm confused... the docker build for the publish action errors out, but the action works: https://github.com/SwissDataScienceCenter/renku/runs/4468084248?check_suite_focus=true#step:3:115

It seems to me that, as it failed building ruamel 0.2.6 it fell back on 0.2.2:
```
    Downloading ruamel.yaml.clib-0.2.6.tar.gz (180 kB)
      ERROR: Command errored out with exit status 1:
       command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-7glshc0u/ruamel-yaml-clib_9e0dcf4141fb43da90a0b3d6271196ce/setup.py'"'"'; __file__='"'"'/tmp/pip-install-7glshc0u/ruamel-yaml-clib_9e0dcf4141fb43da90a0b3d6271196ce/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-_03pd2cz
           cwd: /tmp/pip-install-7glshc0u/ruamel-yaml-clib_9e0dcf4141fb43da90a0b3d6271196ce/
      Complete output (3 lines):
      sys.argv ['/tmp/pip-install-7glshc0u/ruamel-yaml-clib_9e0dcf4141fb43da90a0b3d6271196ce/setup.py', 'egg_info', '--egg-base', '/tmp/pip-pip-egg-info-_03pd2cz']
      test compiling /tmp/tmp_ruamel_p26b_vh1/test_ruamel_yaml.c -> test_ruamel_yaml compile error: /tmp/tmp_ruamel_p26b_vh1/test_ruamel_yaml.c
      Exception: command 'gcc' failed with exit status 1
      ----------------------------------------
  WARNING: Discarding https://files.pythonhosted.org/packages/8b/25/08e5ad2431a028d0723ca5540b3af6a32f58f25e83c6dda4d0fcef7288a3/ruamel.yaml.clib-0.2.6.tar.gz#sha256=4ff604ce439abb20794f05613c374759ce10e3595d1867764dd1ae675b85acbd (from https://pypi.org/simple/ruamel-yaml-clib/) (requires-python:>=3.5). Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
    Downloading ruamel.yaml.clib-0.2.2.tar.gz (179 kB)
```

rokroskar:
good point, I missed that detail

.github/workflows/publish-master-merges.yaml
```diff
@@ -0,0 +1,56 @@
...
+                color: '${{ job.status }}' === 'success' ? 'good' : '${{ job.status }}' === 'failure' ? 'danger' : 'warning',
+                text: `${process.env.AS_JOB} version ${{ env.publish_version }}: ${{ job.status }}.`,
+              }]
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        if: always()
+      - uses: actions-ecosystem/action-push-tag@v1
+        with:
+          tag: ${{ steps.bump-semver.outputs.new_version }}
```
rokroskar, line 55: this tag is wrong (see [here](https://github.com/SwissDataScienceCenter/renku/runs/4468084248?check_suite_focus=true#step:11:9)) - should be `${{env.publish_version))`?

rokroskar (changes requested):

.github/workflows/publish-master-merges.yaml
aledegano: oh, damn, yes, thanks!

rokroskar (approved):

rokroskar (approved):

PR #2383, opened 2021-12-06
test: ignore missing iframe where not embeddable

lorenzo-cavazzi:
fix for SwissDataScienceCenter/renku-ui#1521
/deploy

### DIFF
--- a/acceptance-tests/src/test/scala/ch/renku/acceptancetests/MergeRequestsAndIssuesSpec.scala
+++ b/acceptance-tests/src/test/scala/ch/renku/acceptancetests/MergeRequestsAndIssuesSpec.scala
@@ -22,6 +22,7 @@ import ch.renku.acceptancetests.tooling.{AcceptanceSpec, KnowledgeGraphApi}
 import ch.renku.acceptancetests.workflows._
 
 import scala.concurrent.duration._
+import scala.util.Try
 
 class MergeRequestsAndIssuesSpec
     extends AcceptanceSpec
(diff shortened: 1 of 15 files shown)


acceptance-tests/src/test/scala/ch/renku/acceptancetests/MergeRequestsAndIssuesSpec.scala
```diff
@@ -38,7 +38,11 @@ class MergeRequestsAndIssuesSpec
 
     `navigate to the issues tab`
 
-    `verify that the GitLab issues iFrame is visible`
+    // Availability of the iFrame depends on the specifc deployment
+    try `verify that the GitLab issues iFrame is visible`
```
joke1196, line 42: This works but if we wanted to go a bit more scala like a better solution would be 
```
Try(`verify that the GitLab issues iFrame is visible`).fold (
      _ =>  `verify that the GitLab issues iFrame cannot be shown`, // the error path
      identity // the successful path
    )
```

acceptance-tests/src/test/scala/ch/renku/acceptancetests/MergeRequestsAndIssuesSpec.scala
```diff
@@ -54,7 +58,11 @@ class MergeRequestsAndIssuesSpec
 
     `navigate to the merge requests tab`
 
-    `verify that the GitLab MR iFrame is visible`
+    // Availability of the iFrame depends on the specifc deployment
+    try `verify that the GitLab MR iFrame is visible`
```
joke1196, line 62: Same as above

joke1196 (approved):

PR #2392, opened 2021-12-13
(fix) Correct bumping of pre-release.

aledegano:
When bumping the pre-release version ignore what comes after the '.', the GitHub action, otherwise, is confused and bumps the patch number instead of the pre-release one.

Replace the 'level' to be increased in the action that bumps the semver to correctly be `prerelease` instead of `prepatch`.


aledegano:
I have already tested this and works as intended.
Once approved I will remove the incorrect tags and charts that have been published in the meanwhile (0.10.5-0 etc...) just before merging.

rokroskar (approved):

PR #2399, opened 2021-12-15
chore: upgrade to latest Keycloak chart

aledegano:
While the current version of Keycloak is not explicitly affected by Log4shell, it is however safer to ship a more recent version.

/deploy

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,7 +6,7 @@ name = "pypi"
 [packages]
 "urllib3" = ">=1.24.2"
 "jinja2" = ">=2.10.1"
-chartpress = "==0.7.0"
+chartpress = "==1.0.0"
 kubernetes = "*"
 
 [dev-packages]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1e0663169e28964cd872aa3619b304d0e5d9badd17450ffaa6262edc89ee5ac9"
+            "sha256": "09ec27122b35fbbd56897e7eeb556b78c9115ce3aaf568c09fda6b96479f1e35"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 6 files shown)


rokroskar:
did you test to see what happens when you upgrade from the current version to this new one? Any db migrations to be aware of?

aledegano:
> did you test to see what happens when you upgrade from the current version to this new one? Any db migrations to be aware of?

No, I didn't do any tests, I thought it was a trivial upgrade, but as the deployment shows it isn't really. I'll close this PR to avoid confusion. I believe @ableuler will pick this up when his pipeline clears up.

rokroskar:
What was the problem with the deployment?

rokroskar:
It did deploy, actually: https://github.com/SwissDataScienceCenter/renku/runs/4725939662?check_suite_focus=true

> I believe @ableuler will pick this up when his pipeline clears up.

😆 hope dies last

aledegano:
> It did deploy, actually: https://github.com/SwissDataScienceCenter/renku/runs/4725939662?check_suite_focus=true
> 
> > I believe @ableuler will pick this up when his pipeline clears up.
> 
> laughing hope dies last

it deploys, but the last time I've checked all the tests were failing to authenticate...

rokroskar:
@aledegano @ableuler I tested this in rok.dev.renku.ch by first installing the current `master` and then upgrading with this PR. Seems to work normally so I think the PR is gtg.

aledegano:
Wow, great, thanks for picking this up!

Pipfile
```diff
@@ -6,7 +6,7 @@ name = "pypi"
 [packages]
 "urllib3" = ">=1.24.2"
 "jinja2" = ">=2.10.1"
-chartpress = "==0.7.0"
+chartpress = "==1.0.0"
```
ableuler, line 9: Wasn't there a reason for pinning this to 0.7.0 @aledegano ?

rokroskar: CI is using 1.0... I think that was part of the reason for the weird build discrepancy we were looking at last week @ableuler

aledegano: IIRC the 0.7.0 was the latest version that still supported having the chartpress file in a different folder than that it expected.
However this was ~1 year ago and things might have changed(?)

aledegano: https://github.com/jupyterhub/chartpress/issues/118

rokroskar: afaict all of our chart building actions work fine with this change. But maybe I'm not hitting all the places that could break?

aledegano: :shrug:

ableuler: alright then, let's merge this and remind ourselves that this might be a problem if pipelines suddenly started to fail :)

PR #2407, opened 2021-12-20
chore(ci): disable staging workflow

olevski:
Flux is managing this now. So this is not needed. But the helm chart still needs to be published so that part of the workflow is kept and renamed accordingly.


rokroskar (approved):
Thanks @olevski - though I wonder if we should have the various ways of pushing charts in the same file so it's easier to keep track of. We can do that in a future refactor

PR #2410, opened 2021-12-22
test: corrected fetch of current renku version in acceptance test

joke1196:
During the JupyterNotebookSpec the project renku version is fetch from the status page and used to update the project when needed.  A change in the UI made this process simpler and more robust. See: https://github.com/SwissDataScienceCenter/renku-ui/pull/1618

When performing a free text search for a dataset the order of the dataset was random. This PR fixes it by sorting dataset by Title

/deploy renku-ui=1.0-next  renku-graph=cli-v1 renku-core=2507-lock-status-endpoint


lorenzo-cavazzi (approved):
Thanks for your effort on this!

PR #2411, opened 2021-12-22
chore(test): add new test and cleanup actions

olevski:
/deploy #persist

Implements the new test and cleanup actions from https://github.com/SwissDataScienceCenter/renku-actions/pull/16

And also adds continuous mirroring of the acceptance tests artefacts to S3. This way even if the tests timeout or crash or whatever there will always be some artifacts that can be uploaded/added to the github workflow.

In the case when the test artefacts are added to S3 only once the tests are complete there were issues because if the tests time out or crash in some weird way then the artefacts are not uploaded and cannot be found on the github workflow. This should avoid a lot of those scenarios.

### DIFF
--- a/.github/workflows/publish-helm-chart.yml
+++ b/.github/workflows/publish-helm-chart.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set version
         id: vars
         run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
-      - uses: SwissDataScienceCenter/renku-actions/publish-chart@v0.3.2
+      - uses: SwissDataScienceCenter/renku-actions/publish-chart@v0.4.0
         env:
           CHART_DIR: helm-chart/
           CHART_NAME: renku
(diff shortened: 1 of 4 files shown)


rokroskar:
Thanks @olevski ! There seem to be a few issues still:

* the deployment was not actually removed after the test completed (I just pushed a commit to use the new version of the check-pr-description action)
* there was a failure in the test but the test is shown as passing 
* the artifacts seem to be missing (not uploaded to github?)

olevski:
REMINDER: This should be merged after a new version of the renku actions is published, and the actions here are updated to said version.

rokroskar:
@olevski the other PR is in so this one should be updated and merged - once we do that, we should write a quick message on the dev channel to tell people about the new behavior.

olevski:
@rokroskar I will prep a release on renku actions. Once that release is done then automated PRs would go out to every repo that uses the actions. I will explain in slack how to deal with this. I may go and check all the PRs. I think that the github actions in all the different repos that run these tests will have to be edited a bit.

rokroskar:
yeah they do because the test is now run as an external action. Thanks!

olevski:
@rokroskar the acceptance tests are failing but not because of this action. I think we can merge this even though we do not have passing acceptance tests. The most important thing is that the action is properly handling the failing acceptance tests.

And because I recently turned on the `#persist` flag the deployment is not deleted as expected when the tests complete.

rokroskar (approved):

PR #2416, opened 2022-01-11
docs: remove helm2 init command

wesjdj:
Helm init is not used anymore in Helm 3. Helm 2 is no longer supported. Therefore the Helm init command should be removed from the documentation.


ableuler (approved):
Well spotted 👍

PR #2418, opened 2022-01-12
test: corrected css selector for dataset not in kg alert

joke1196:
Fixes a css selector in the acceptance test for dataset search

/deploy renku-ui=0000-not-in-kg renku-graph=cli-v1 renku-core=0000-disable-migration-tmp


lorenzo-cavazzi (approved):
Great to see all green again :star_struck: 
Thanks for the effort! I'll merge the UI counterpart to the `1.0-next` branch

PR #2421, opened 2022-01-14
chore: make Keycloak theme compatible with current Keycloak version

ciyer:
Use updated Keycloak theme

/deploy

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -201,7 +201,7 @@ keycloak:
 
   extraInitContainers: |
     - name: theme-provider
-      image: renku/keycloak-theme:2.0.1
+      image: renku/keycloak-theme:3.0.1
       imagePullPolicy: IfNotPresent
       command:
         - sh


ableuler:
The ci deployment does not have an identity provider as login option configured. I guess you tested this on a personal dev deployment with an identity provider set up?

rokroskar:
What is this bit at the bottom? I looked at it with Firefox and Chrome, looks the same 
image

lorenzo-cavazzi:
Same thing here

ciyer:
That's the background that is included with the new Keycloak theme. Sorry didn't catch that since my screen isn't large enough to see that, but I will make a fix.

ciyer:
Ok, fixed in https://github.com/SwissDataScienceCenter/keycloak-theme/pull/10

You can see it in action if you try to login on this instance: https://sekhar.dev.renku.ch/

helm-chart/renku/values.yaml
```diff
@@ -201,7 +201,7 @@ keycloak:
 
   extraInitContainers: |
     - name: theme-provider
-      image: renku/keycloak-theme:2.0.1
+      image: renku/keycloak-theme:3.0.0
```
rokroskar, line 204: Lets use a new tag, `3.0.1`?

rokroskar (changes requested):

ciyer:
I have activated two dummy identity providers on the PR deployment. You will not be able to log into RenkuLab using them, but you can see what the buttons look like.

ableuler (approved):
LGTM!

rokroskar (approved):

PR #2425, opened 2022-01-18
chore: prep 0.11.2 and update release notes

rokroskar:
Prep 0.11.2 and add images to release notes for 0.11.1


ableuler (changes requested):
I think the gifs are super useful and linking to them from the CHANGELOG is a good idea, but I would not aut0-expand those links.

rokroskar:
@ableuler you mean to have fewer blinky things in the file per default?

ableuler:
> @ableuler you mean to have fewer blinky things in the file per default?

Yes. I think the CHANGELOG file should be something that can be skimmed through quickly, without large images or even animations.

rokroskar:
So this is becoming comically annoying/difficult to resolve... but it's hard to have the images show up/resolve in both the CHANGELOG file itself on github _and_ in the rendered docs. The reason is that for the docs, the images need to be in the `_static` directory... so I moved them there and made a symlink in the root directory. However, when you open the changelog in github and click on the link, it doesn't work. I don't know how much this actually matters or if anyone has an idea how to resolve it easily.

rokroskar:
Ok so to resolve this I did something slightly dirty - I used a direct URL to the images in the 0.11.2 release. It's now inconsistent since the images refer to the 0.11.1 release but I think we can live with that 😄

ableuler (approved):

PR #2429, opened 2022-01-19
Update teaching_with_renkulab.rst

gavin-k-lee:
iFrame embeddings will be phased out in the next Renku release so we will point those who want to announce teachings directly to the JotForm.


docs/topic-guides/teaching_with_renkulab.rst
```diff
@@ -31,8 +31,8 @@ We want to give your teachings the best Renku experience, whether they are one-o
 workshops or weekly lab sessions. This means ensuring adequate computational
 resources for you and your sessions, with a Renku admin on hand if there are
 problems with the platform. As such, we would very much appreciate if you could
-*announce* your teaching using the following form:
-https://renkulab.io/projects/team-renku/renku-teaching-announcements/files/blob/Form.ipynb.
+*announce* your teaching using the following form: `Announce your Renku teaching
```
rokroskar, line 34: just wondering about the wording here... "announce your teaching" sounds a bit strange, what about "announce your course"?

rokroskar: (I realize that's not what this PR is about but I didn't notice before) :)

gavin-k-lee: Sure, we can do that.

rokroskar (approved):
Thanks!

PR #2433, opened 2022-01-24
fix: custom ca certs dockerfile

olevski:
It seems that the package we used before was just removed fully from the package registry on alpine. 

I pinned the packages that are now available on the package registry. The build now succeeds.

/deploy


olevski:
I will open another PR to fix this in master.

olevski:
Also please note that the target branch here is `1.0-next` and not master.

lorenzo-cavazzi (approved):
Thanks for the quick fix! Let's merge it as soon as the deployment is ready -- the acceptance tests won't pass on this branch

PR #2434, opened 2022-01-24
fix: custom ca certificates image build

olevski:
It seems that the packages we previously used and we had pinned are no longer available for alpine.

/deploy


ableuler (approved):
Thanks for fixing this!

PR #2436, opened 2022-01-26
fix(ci): use latest release tag on master merges

olevski:
So the current state of this workflow is such that it grabs the latest tag and it uses that to publish a chart. But the latest tag could be anything.

That is why we got into a situation where a feature tag like `0.0.1-certificates` that I added to a feature branch was picked up and is now used for every merge to master.

This change makes it so that only the latest official release tag is used. See example:
https://github.com/SwissDataScienceCenter/renku/runs/4951186244?check_suite_focus=true

In the example above I commented out the chart publishing and made it run on any push. I will remove these temp changes. But it shows this works as intended.

Here is the weird workflow that got published because of a feature branch tag that should be ignored.
https://github.com/SwissDataScienceCenter/renku/runs/4950384788

### DIFF
--- a/.github/workflows/publish-master-merges.yaml
+++ b/.github/workflows/publish-master-merges.yaml
@@ -2,12 +2,12 @@ name: Publish all merges to master
 
 on:
   push:
-    # branches: ["master"]
-    # paths:
-    #   - helm-chart/**
-    #   - scripts/**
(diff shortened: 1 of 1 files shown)


ableuler (approved):
I think it makes sense to make this based on releases rather than tags. 👍

PR #2444, opened 2022-02-01
chore: create ci ns outside of rancher projects

ableuler:
/deploy


rokroskar:
shouldn't this PR be titled "bump to actions 0.5.0"? It's not making the change it says it's making in the title.

ableuler:
I thought describing the effect of bumping the action is more helpful here.

rokroskar:
yes, except when the reviewer goes through the PR line-by-line multiple times looking for where the namespace is actually set and concluding it must be a change coming in from somewhere else 🤷

rokroskar (approved):

PR #2454, opened 2022-02-07
fix(ci): undo recent changes to tag/release numbering

ableuler:
This reverts commit 85618992425258f1d0b1dfb96a9da0a385b91ae3.

Undoing it because after this changes the first digit of the pre-release numbering isn't increased anymore. We have to otherwise fix the problems with wrong tags being used as basis for figuring out the right tag after tagging a non-renku chart from this repo.
image


rokroskar:
I don't understand why that change meant the pre-release wasn't bumped anymore - it happens in a step lower down...?

rokroskar:
oh because it's not looking at pre-releases... 👍

olevski (approved):

olevski:
Keep in mind that we should not tag anything manually on renku though. Because any new tags added will be used as the basis.

ableuler:
> Keep in mind that we should not tag anything manually on renku though. Because any new tags added will be used as the basis.

Absolutely agree, this is just temporary

PR #2456, opened 2022-02-07
fix(ci): properly set helm chart version

olevski:
This should fix the previous mess I made.

### DIFF
--- a/.github/workflows/publish-master-merges.yaml
+++ b/.github/workflows/publish-master-merges.yaml
@@ -2,12 +2,12 @@ name: Publish all merges to master
 
 on:
   push:
-    branches: ["master"]
-    paths:
-      - helm-chart/**
-      - scripts/**
-      - acceptance-tests/**
-      - .github/**
+    # branches: ["master"]
+    # paths:
+    #   - helm-chart/**
+    #   - scripts/**
+    #   - acceptance-tests/**
+    #   - .github/**
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


olevski:
Example of working workflow. I commented out the part that actually does the publishing. But you can see all the other parts that figure out the tag - and it comes up with the right tag.

https://github.com/SwissDataScienceCenter/renku/runs/5093125960?check_suite_focus=true

.github/workflows/publish-master-merges.yaml
```diff
@@ -17,19 +17,28 @@ jobs:
...
+        with:
+          version: '3.8.0'
+        id: setup-helm
+      - id: get-latest-helm-charts
+        run: |
+          helm repo add renku https://swissdatasciencecenter.github.io/helm-charts/
+          helm repo update
+          helm search repo renku --devel -o yaml > .latest-charts.yaml
+      - id: get-latest-version
+        uses: mikefarah/yq@v4.19.1
```
rokroskar, line 30: I wonder if `jq` is present in the actions builders already by default - if that's the case I would skip this extra dependency and output to json in the step above.

olevski: Ok I will test for that.

olevski: yeah that works, see here: https://github.com/SwissDataScienceCenter/renku/runs/5093411722?check_suite_focus=true

One sec let me just remove the stuff I added to test

ableuler (approved):
LGTM

.github/workflows/publish-master-merges.yaml
rokroskar: looks good, thanks!

PR #2466, opened 2022-02-11
fix: create Project button clicking problem

jachro:
It looks like sometimes there's a problem with a template tooltip overlaying the Create Project button. This PR tries to mitigate that by scrolling the page to the bottom.

/deploy


rokroskar:
lets run this with the community templates turned off - I'll try to make a change to the deploy command

jachro:
:+1

rokroskar:
hmm I don't think it's actually possible to do that right now (the regex extracting the extra-values matches all non-whitespace characters). I'll try to modify it manually in the deployment once it's up and we can run the test manually.

rokroskar:
seems like scrolling to get rid of the tooltip is a bit hacky - is there no way to actually make the tooltip go away specifically or to just not have it show up in the first place?

rokroskar:
I don't think this solves the problem actually (running the tests now) - it could be solved easily by inverting the order in which the operations happen, i.e. first click on the template button, then go fill in the rest of the form and then click "create project" - wdyt @jachro ?

jachro:
> 

Yes, why not. Let me try...

ciyer (commented):
It looks like this fixes the problem (judging from the green mark by the test), which is great, but I have a question about a code change you made.

rokroskar (changes requested):
The change does not yet fix the problem (the acceptance tests failed in the same way as before after changing the templates that are displayed)

rokroskar:
Please don't update the branch until the current test is done!

acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/NewProjectPage.scala
```diff
@@ -46,10 +47,19 @@ case object NewProjectPage
     visibilityField.click() sleep (1 second)
     visibilityOption(project.visibility).click() sleep (1 second)
 
+    scrollDown
```
rokroskar, line 50: do we need `scrollDown` now? It seems that we don't.

ciyer: I think it could be necessary if there are many templates, and it does not hurt, so I would leave it.

rokroskar: good point

rokroskar (approved):

rokroskar:
The tests passed (with the modification to the template list that would normally trigger the bug) so I think it's gtg!

PR #2482, opened 2022-02-16
chore: remove ipython as dev dependency

ableuler:
ipython should not really be a dev dependency.


rokroskar (approved):
Thanks, this has been on my to-do for a few weeks 🙈

PR #2483, opened 2022-02-16
fix(chart): global redis values

olevski:
This should simplify the setup for a global redis instance.

Previously we were trying to extract template values from one sibling chart into another (in the components helm charts) - which does not really work.

With this change everything the components need is specified in the global redis section. So if someone brings their own redis then they do not install the redis chart bundled with renku but they fill out all stuff in `global.redis`. If people use the included redis then they can stick with the defaults and everything should work. If they change the defaults then the `global.redis` section and regular `redis` section need to be compatible and some values between the two will be duplicated.

**THIS IS A BREAKING CHANGE FOR THE HELM CHART**

Need to publish a new version of renku cli 0.16.x to get renku-core to work with redis with a sentinel.


helm-chart/renku/values.yaml
```diff
@@ -83,11 +83,10 @@ global:
     # Specify which renku component should use which DB index
     dbIndex:
       gateway: "0"
-      # coreService: "1"
-      # uiServer: "2"
+      coreService: "1"
```
ableuler, line 86: that's something I have been wondering - do the different versions of the core service handle this on their own, for example by prefixing the keys with their version?

olevski: Yes. They use: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/helm-chart/renku-core/templates/deployment.yaml#L82-L83

olevski:
The PRs that need this to be done are:
- https://github.com/SwissDataScienceCenter/renku-python/pull/2696
- https://github.com/SwissDataScienceCenter/renku-gateway/pull/552
- https://github.com/SwissDataScienceCenter/renku-ui/pull/1699

helm-chart/renku/values.yaml
ableuler: perfect.

ableuler (dismissed):
This PR seems fine, but I guess if we merge this before the corresponding gw release is out things will break, no?

olevski:
@ableuler I think the easiest this is to apply the bump in version for the gateway here rather than in the bot PR. I will close the bot PR if you can re-approve this? That way we dont have a commit on master that is broken.

ableuler (approved):

ableuler:
> @ableuler I think the easiest this is to apply the bump in version for the gateway here rather than in the bot PR. I will close the bot PR if you can re-approve this? That way we dont have a commit on master that is broken.

yeah, totally agree, the two changes need to be merged together in one PR

ableuler (approved):

Issue #2490, opened 2022-03-01
Enable OAuth 2 device authorization grant in Keycloak


mohammad-sdsc:
**Is your feature request related to a problem? Please describe.**
To support device auth grant for CLI login (https://github.com/SwissDataScienceCenter/renku-python/issues/2636) we need to enable it for the `renku-cli` client in the Keycloak.

**Describe the solution you'd like**
- Enable Device Auth Grant for `renku-cli`
- Make `renku-cli` a public client so that it can be authenticated without a client secret

**Additional context**
- Make sure that changes are applied to existing deployments
- `renku-cli` tokens should have a longer expiration date so that CLI users don't have to login frequently

PR #2491, opened 2022-03-01
feat(chart): enable device auth grant

mohammad-sdsc:
Fixes: #2490

### DIFF
--- a/helm-chart/renku/templates/post-install-job-keycloak.yaml
+++ b/helm-chart/renku/templates/post-install-job-keycloak.yaml
@@ -33,7 +33,8 @@ spec:
             "--admin-password=$(KEYCLOAK_ADMIN_PASSWORD)",
             "--keycloak-url=$(KEYCLOAK_URL)",
             "--users-file=/app/data/users",
-            "--clients-file=/app/data/clients"
+            "--clients-file=/app/data/clients",
+            "--apply-changes"
           ]
 
           volumeMounts:
--- a/helm-chart/renku/templates/post-install-job-keycloak.yaml
+++ b/helm-chart/renku/templates/post-install-job-keycloak.yaml
@@ -34,7 +34,7 @@ spec:
             "--keycloak-url=$(KEYCLOAK_URL)",
             "--users-file=/app/data/users",
             "--clients-file=/app/data/clients",
-            "--apply-changes"
+            "--force"
           ]
 
           volumeMounts:
--- a/helm-chart/renku/templates/_keycloak-clients-users.tpl
+++ b/helm-chart/renku/templates/_keycloak-clients-users.tpl
@@ -32,10 +32,9 @@ Define clients and users for Keycloak
     "baseUrl": "{{ template "http" . }}://{{ .Values.global.renku.domain }}",
     "secret": "{{ required "Fill in .Values.global.gateway.cliClientSecret with `uuidgen -r`" .Values.global.gateway.cliClientSecret }}",
     "publicClient": true,
-    "standardFlowEnabled": false,
     "attributes": {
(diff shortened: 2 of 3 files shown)


ableuler (commented):
How do you want to handle the transition period? I guess one option would be for the server side to support both, old and new CLI login for a certain period.

mohammad-sdsc:
Yes, that would work. This PR does not remove anything so, it's safe to merge it. We should keep the custom auth flow in the gateway for a while (i.e. not merging https://github.com/SwissDataScienceCenter/renku-gateway/pull/558 for now) to support old CLI clients.

ableuler:
Sounds good. But for this to work we'd have to still allow the standard flow on the renku-cli for the time being, no?

mohammad-sdsc:
I've checked the `init-realm.py` script and it doesn't update clients automatically. It only warns about mismatches between existing and new clients. I see that this scripts hasn't been updated in ages so I'm not sure what's the policy regarding it: Should we update it or should we apply changes to the deployments manually. If former, then I can add a section to the script to copy new client attributes to an existing one.

ableuler:
Thanks for checking. I think the post-install job should indeed delete and recreate clients which do not match the expected configuration.

rokroskar:
Is this waiting for a change to the post-install job?

rokroskar:
@mohammad-sdsc thanks for looking into this! Can't we just nuke the old configuration and apply the new one in the case that there is a config change? That's effectively what we do by hand now and it works fine...

mohammad-sdsc:
> Can't we just completely remove the old configuration and apply the new one in the case that there is a config change? That's effectively what we do by hand now and it works fine...

I didn't realize this. Thanks for mentioning it! It's more robust and simpler to implement.

rokroskar:
@ableuler please chime in if I misrepresented how this is currently handled :)

ableuler:
> Can't we just completely remove the old configuration
What do you mean by existing configuration? Do you mean _all_ clients in the Renku realm?

rokroskar:
I meant the client that has the configuration conflict. Though in practice we actually removed all the clients and let the post-install script recreate them iirc, but I wouldn't do that in general.

mohammad-sdsc:
> I never delete all clients by hand and even more so I think that we should not do this programmatically.

It doesn't delete all clients; it only deletes/recreates clients that have changed. The difference with the previous approach is in the configuration that is used to recreate the client: Before it was using the existing configuration but now it's using the new configuration. Basically, this should be fine if all the required configuration is already in the value files/templates. If there are manual changes to the clients (i.e. changes via Keycloak UI) then those are gone.

ableuler:
Yeah, the `recreate client instead of update` aspect makes perfect sense to me. I just wasn't sure if Rok was talking about the same thing. Anyway, I think we all agree.

helm-chart/renku/templates/post-install-job-keycloak.yaml
```diff
@@ -33,7 +33,8 @@ spec:
             "--admin-password=$(KEYCLOAK_ADMIN_PASSWORD)",
             "--keycloak-url=$(KEYCLOAK_URL)",
             "--users-file=/app/data/users",
-            "--clients-file=/app/data/clients"
+            "--clients-file=/app/data/clients",
+            "--apply-changes"
```
ableuler, line 37: maybe after your most recent edit calling this just `--force` might be more descriptive?

mohammad-sdsc:
Follow up story: https://github.com/SwissDataScienceCenter/renku/issues/2504

ableuler (approved):
Thanks @mohammad-sdsc, let's get this merged!

PR #2493, opened 2022-03-03
tests: fix code for finding the hide advanced settings link

ciyer:
The text for the "hide advanced settings" link in sessions changed (was previously "hide branches").

### DIFF
--- a/acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
+++ b/acceptance-tests/src/test/scala/ch/renku/acceptancetests/pages/ProjectPage.scala
@@ -331,7 +331,7 @@ class ProjectPage(val projectSlug: String, val namespace: String)
     }
 
     def maybeButtonHideBranch(implicit webDriver: WebDriver): Option[WebElement] = eventually {
-      findAll(cssSelector("div.mb-3 > button")).find(_.text.startsWith("Hide branch"))
+      findAll(cssSelector("div.mb-3 > button")).find(_.text.startsWith("Hide advanced settings"))
     }
 
     private def maybeImageReadyBadge(implicit webDriver: WebDriver): Option[WebBrowser.Element] =


lorenzo-cavazzi (dismissed):

olevski:
@andre-code you can use this PR to bump the UI version when a UI release is made. Since this PR already contains the change to fix the acceptance tests. This way once this is merge we will have completely functioning commit on Renku. Rather than intermediate commits where only part of the required changes are present.

lorenzo-cavazzi (approved):
Looks good! Let's see if the tests pass here before merging

olevski:
Btw the acceptance tests seem to fail multiple times with this message:

```
The code passed to eventually never returned normally. Attempted 551 times over 2.0031752506333333 minutes. Last failure message: Expanding selection to branch, commit, image not found..
```

So I think this has to do something with a UI element not properly located in the tests. I did a few manual tests on the CI deployments and I can launch sessions without an issue.

olevski:
If this is the button that the error and the commit in this PR referes to:

image


Then shouldn't we have:

```
def maybeButtonHideBranch(implicit webDriver: WebDriver): Option[WebElement] = eventually {
      findAll(cssSelector("div.mb-3 > button")).find(_.text.startsWith("Do you want to select the branch"))
    }
```

Instead of

```
def maybeButtonHideBranch(implicit webDriver: WebDriver): Option[WebElement] = eventually {
      findAll(cssSelector("div.mb-3 > button")).find(_.text.startsWith("Hide advanced settings"))
    }
```

ciyer:
> If this is the button that the error and the commit in this PR referes to:
> Then shouldn't we have:
> 
> ```
> def maybeButtonHideBranch(implicit webDriver: WebDriver): Option[WebElement] = eventually {
>       findAll(cssSelector("div.mb-3 > button")).find(_.text.startsWith("Do you want to select the branch"))
>     }
> ```
> 
> Instead of
> 
> ```
> def maybeButtonHideBranch(implicit webDriver: WebDriver): Option[WebElement] = eventually {
>       findAll(cssSelector("div.mb-3 > button")).find(_.text.startsWith("Hide advanced settings"))
>     }
> ```

The logic is a bit convoluted, but correct. That bit of UI clicking is there to reveal the content that says whether the image is available or not. The test first clicks the "Do you want to select the branch" link, then to checks for the hideBranch link to see if it will be able to find the image ready information.

```
    def verifyImageReady(implicit webDriver: WebDriver): Unit = eventually {
      sleep(1 second)
      maybeButtonHideBranch getOrElse {
        buttonShowBranch.click()
        sleep(5 seconds)
      }
      maybeImageReadyBadge getOrElse {
        sleep(10 seconds)
        webDriver.navigate.refresh()
        fail("Image not ready")
      }
    }
```

The strange thing is, the pod that is running the test is not using the code that is in this branch. Any idea why that might be the case?

image

That should be `Hide advanced settings`, not `Hide branch` on the second line there.

olevski:
So that it the same problem we saw before. Where somehow the wrong image (or even completely wrong helm chart version) is being used by the CI pipeline. I will do some digging tomorrow.

olevski:
Yeah so it seems that the image for the acceptance tests with the same name (but wrong contents) is already present in dockerhub. In this case chartpress just decides (rightfully so) to skip the image build and it reuses the existing image.

So the mystery is how such an image exists on dockerhub in the first place. Chartpress comes up with unique names for each image to prevent this case. So it could be that chartpress messed up. Or someone manually replaced this image for another image on dockerhub for some reason.

I think chartpress uses part of the commit sha in the image name. So even if you changed something and force pushed to rewrite an old commit then the image will still be properly updated.

olevski:
This is the same content as this PR. Just with some new lines added in the acceptance tests Dockerfile so that the image is rebuilt. This should cause the acceptance tests to pass.

https://github.com/SwissDataScienceCenter/renku/pull/2503

olevski:
Ok so I deleted the suspect image on Dockerhub and restarted the pipeline. This should clear things up. Will try to see if I can reproduce this problem in the other test PR.

olevski:
So now the correct image is used for the tests. The failures are just flukes. The PR I opened with the exact same contents just to test things out passed. Hopefully this restart takes care of things.

olevski:
These tests will fail again. Somehow now because of caching or some other issues the wrong image for the tests is pulled again. 

I will open a brand new PR and make the changes from this PR there. That should resolve things hopefully.

PR #2495, opened 2022-03-08
chore: adapt for use of traefik helm chart

ableuler:


rokroskar:
I don't see why this depends on https://github.com/SwissDataScienceCenter/renku-gateway/pull/566? It's needed to make current gateway master work (without that PR). cc @mohammad-sdsc

PR #2507, opened 2022-03-16
chore: docker credentials updated in publish-helm-chart.yml

jachro:
It looks we've got some variables' names mismatched.


ableuler (approved):
Thanks 🙏 !

PR #2514, opened 2022-03-17
chore: notebooks 1.5.0, UI 2.1.0 release

olevski:
So chartpress has a bug where it labels the image wrong when you update a branch in a PR.

To avoid this bug I had to reopen the PR so that it does not need to be updated against master.

/deploy #persist


jachro (approved):
It looks good to me but I'm not sure if my approval is good enough here ;)

PR #2526, opened 2022-03-23
chore: prep 0.12.4

rokroskar:
Changelog for the release


lorenzo-cavazzi (approved):
If this should not be bumped to `version: 0.12.4` (I'm not sure about the outcome from yesterday's discussion), it's ok. 
https://github.com/SwissDataScienceCenter/renku/blob/prep-0.12.4/helm-chart/renku/Chart.yaml

rokroskar:
@lorenzo-cavazzi we normally don't worry about updating that since it gets set automatically on chart release by chartpress, but it's nicer if it's consistent. Thanks for pointing it out!

lorenzo-cavazzi (approved):

Issue #2524, opened 2022-03-23
deprecate workflow-cleanup-action


rokroskar:
It appears that github actions natively supports cancelling previously running workflow jobs since... erm last year. 🤷‍♂️ We should stop using the hacky https://github.com/rokroskar/workflow-run-cleanup-action and use [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) settings instead. 

- [x] renku (https://github.com/SwissDataScienceCenter/renku/pull/2528)
- [x] renku-python
- [x] renku-graph
- [x] renku-notebooks
- [x] renku-gateway
- [x] renkulab-docker

PR #2528, opened 2022-03-23
chore: use concurrency in actions

rokroskar:
addresses #2524

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -9,7 +9,7 @@ on:
       - edited
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -8,14 +8,11 @@ on:
       - closed
       - edited
 
+concurrency:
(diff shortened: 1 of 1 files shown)


rokroskar:
This seems to work: https://github.com/SwissDataScienceCenter/renku/actions/runs/2028896758

ableuler (dismissed):

.github/workflows/pull-request-test.yml
```diff
@@ -8,14 +8,11 @@ on:
       - closed
       - edited
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
Panaetius, line 12:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
maybe? so it cancels everything on the same PR if it runs on a PR?

rokroskar: ref is the branch, no?

Panaetius: hmm I guess it should. But I see a lot of repos using the above in the wild.

Panaetius: I think if you had multiple PRs to merge the same branch into different branches (like a hotfix branch into master and develop if you follow proper git-flow) and the trigger to run actions is on PR, one PR would cancel the other PR?

rokroskar: gitflow is beyond my ability of comprehension. If we have the PR number though do we need the ref in there at all? This workflow only triggers on PRs

ableuler (approved):

PR #2535, opened 2022-03-29
chore(docs): add how to guide for s3 buckets

olevski:
Adds a how-to-article on mounting s3 buckets.

### DIFF
--- a/docs/how-to-guides/index.rst
+++ b/docs/how-to-guides/index.rst
@@ -16,6 +16,7 @@ aimed at active users of Renku and target specific use-cases or common issues.
    migrating-projects-to-renku
    collaboration
    git-tips
+   s3
    troubleshooting
    admin/index
    contributing/index
--- a/docs/how-to-guides/index.rst
+++ b/docs/how-to-guides/index.rst
@@ -16,7 +16,6 @@ aimed at active users of Renku and target specific use-cases or common issues.
    migrating-projects-to-renku
    collaboration
    git-tips
-   s3
    troubleshooting
    admin/index
    contributing/index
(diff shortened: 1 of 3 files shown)


docs/how-to-guides/s3.rst
```diff
@@ -0,0 +1,54 @@
...
+8. Click ``Start session``.
+9. Once you session is ready the bucket will be available inside your session at ``/cloudstorage/<bucket name>``. In this case for the GIAB bucket it will be available at ``/cloudstorage/giab``.
+
+Limitations
+-----------
+
+There are a few limitations of this feature that users should be aware of. Please note that as we 
+further develop this feature some of these limitations will be relaxed or fully lifted.
+
+* **Accessing data in mounted buckets is 3x to 10x slower than accessing data on disk.**
```
rokroskar, line 40: I would just say "potentially much slower" since the details will depend on where the S3 is located. It could be at switch, in which case it's probably the same performance as "local" disk in the sessions.

rokroskar (changes requested):
This is great @olevski, thanks! Just some minor phrasing suggestions

docs/how-to-guides/index.rst
```diff
@@ -16,6 +16,7 @@ aimed at active users of Renku and target specific use-cases or common issues.
    migrating-projects-to-renku
    collaboration
    git-tips
+   s3
```
rokroskar, line 19: sorry, just noticed this - I think it should go into the toc in `sessions.rst`

rokroskar (changes requested):
One more thing... maybe we should put this with the rest of the sessions how-tos?

docs/how-to-guides/index.rst
olevski: No problem at all. That makes perfect sense.

rokroskar (approved):

PR #2546, opened 2022-04-06
chore: Update java 18 cert

travissarles:
chore: update java 18 cert

related to https://github.com/SwissDataScienceCenter/renku-graph/issues/836

### DIFF
--- a/helm-chart/certificates/templates/_volume-mounts.tpl
+++ b/helm-chart/certificates/templates/_volume-mounts.tpl
@@ -11,13 +11,6 @@
   readOnly: true
 {{- end -}}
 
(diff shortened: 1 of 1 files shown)


olevski (approved):

olevski:
openjdk 15+ have a symlink from their installation to the system-level java certs. So we do not need to have a volume mount stanza for each version anymore. openjdk 14 does not have this.
```
 ls -l /opt/openjdk-15/lib/security/cacerts
lrwxrwxrwx    1 root     root            27 Jul 29  2020 /opt/openjdk-15/lib/security/cacerts -> /etc/ssl/certs/java/cacerts
```

PR #2547, opened 2022-04-06
chore: publish and update 0.0.2 certificates chart

olevski:
The new version of jdk used by renku-graph required some changes.

So we need a new version. These new changes are general enough so that subsequent releases of jdk will not require more changes.


olevski:
The image in the values.yaml file for renku stays at 0.0.1 because there no changes required there. So a new image is not needed.

PR #2557, opened 2022-04-12
docs: update deploying renku how-to

### LINKED ISSUE #2541: Update documentation to deploy Renku
@aledegano commented on [Mon Apr 04 2022](https://github.com/SwissDataScienceCenter/terraform-renku/issues/477)

The page: https://renku.readthedocs.io/en/latest/how-to-guides/admin/deploying-renku.html
is outdated and contains instructions that are too specific to Openstack.

Update the documentation with the relevant information and give more generic examples in addition to the ones specific to Openstack.

rokroskar:
This PR refreshes the admin how-to doc about deploying renku for the first time. It also adds a new script for generating renku values using a template. This script is an adaptation of the script used previously for generating values for our dev deployments. It can be run either via a wrapper scripts that sets up the python environment or via a docker container. 

To test the values script, you can run 

```
mkdir renku-values
cd renku-values
wget https://raw.githubusercontent.com/SwissDataScienceCenter/renku/000-admin-deploy-docs/scripts/generate-values/generate-values.sh
sh generate-values.sh
```

I'll remove the commit that references this branch after the review is done. 

closes #2541


rokroskar:
Not really sure how to properly test that this works, but I added a CI workflow that at least runs the script and makes sure it completes without error.

aledegano (dismissed):

rokroskar:
@aledegano this actually has a bug - it won't work properly for deploying gitlab _with_ renku - I'll try to fix the script

rokroskar:
@aledegano I've tested this now also with a bundled gitlab deployment and it works. The values file is still a mess, but that's a different problem imho.

aledegano:
> @aledegano I've tested this now also with a bundled gitlab deployment and it works. The values file is still a mess, but that's a different problem imho. 

Okay, I'll try it out ASAP.

docs/how-to-guides/admin/deploying-renku.rst
```diff
@@ -209,13 +236,13 @@ GitLab deployed as part of Renku
...
 If your RenkuLab deployment includes GitLab you need to follow some additional steps to configure an admin user on GitLab.
 
-#. turn off automatic redirection to GitLab by redeploying with the value ``gitlab.oauth.autoSignIn: false``
-#. log in as the root user using the password from ``gitlab.password``
-#. modify any users you want to modify (e.g. to make them admin)
-#. turn the ``autoSignIn`` setting back on
+#. Turn off automatic redirection to GitLab by setting the value ``gitlab.oauth.autoSignIn=false`` and redeploy with ``helm``.
+#. Log in as the root user using the password from ``gitlab.password``.
+#. Modify any users you want to modify (e.g. to make them admin).
+#. Turn the ``autoSignIn`` setting back on and redeploy.
```
olevski, line 242: We have to redeploy renku just because admin users were added to gitlab? Why is this the case?

docs/how-to-guides/admin/configurations/standalone-keycloak.rst
```diff
@@ -3,10 +3,30 @@
...
+secrets in the values file. The mapping between Keycloak clients and Renku
+values is as follows:
+
+=======================   =====================================
+Keycloak client           Renku values file
+=======================   =====================================
+``renku``                 ``.global.gateway.clientSecret``
+``renku-cli``             ``.global.gateway.cliClientSecret``
+``renku-ui``              ``.global.uiserver.clientSecret``
+``renku-jupyterserver``   ``.notebooks.oidc.clientSecret``
```
olevski, line 28: Should we add (the optional) gitlab client secret here? I.e. it is `global.gitlab.clientSecret` in the values file and the client is called `gitlab` in keycloak.

olevski: Also when I was checking the values.yaml file for renku for the kc clients I see that there is also stuff under `global.gateway.gitlabClientId` and this is set to `renku-ui`? Also there is `global.gateway.gitlabClientSecret`. Not sure where these come in and how they relate to `.global.uiserver.clientSecret`.

olevski (changes requested):

docs/how-to-guides/admin/deploying-renku.rst
rokroskar: This is just to log in as the root user and make a normal user an admin. There is no other way to log in as the root gitlab user because the root user doesn't exist as a renku user.

docs/how-to-guides/admin/configurations/standalone-keycloak.rst
rokroskar: yeah the `global.gateway.gitlabClientId` is unfortunately called `renku-ui` - this is the client id for the renku app in gitlab actually. The `global.uiserver.clientSecret` is the UI keycloak client.

olevski (approved):

Issue #2523, opened 2022-03-22
Replace usage of apt-get inside the postgres container in configmap


shock0572:
The configmap for Renku helm chart is currently using apt-get inside a container:

https://github.com/SwissDataScienceCenter/renku/blob/8e0e9aae65c4af9d28203eeef0c1960292aee01f/helm-chart/renku/templates/configmap.yaml#L18-L19

This works in non-constrained, non-proxied environments for deployment, but fails anywhere else.
Those lines should be removed and a new container should be used in 

https://github.com/SwissDataScienceCenter/renku/blob/8e0e9aae65c4af9d28203eeef0c1960292aee01f/helm-chart/renku/templates/post-install-job-gitlab.yaml#L26

by choosing a proper container like:
https://hub.docker.com/r/stangirard/alpine-powerhouse
https://hub.docker.com/r/arunvelsriram/utils
that already includes all the needed tools by the post-install-gitlab, or build a renku/utils container in house.
The solution above has been tested and used in our current deployment.

rokroskar:
I agree this is very bad practice and should be changed :) btw, maybe update the description to link to current code, e.g. https://github.com/SwissDataScienceCenter/renku/blob/8e0e9aae65c4af9d28203eeef0c1960292aee01f/helm-chart/renku/templates/configmap.yaml#L18-L19

rokroskar:
done in #2563

PR #2563, opened 2022-04-22
chore(app): run post-init jobs as non-root

olevski:
This does a few things:
- run post init jobs with user id 1000
- creates an image that can be used by all the post-init jobs that initialize postgres databases (currently some of them do apt-install X when they start)
- run the certificates init container with much more restrictive root security context (i.e. used drop all linux capabilities and prohibited privilege escalation)

/deploy #persist

closes https://github.com/swissdatasciencecenter/renku-notebooks/issues/1003


rokroskar:
Also closes #2523

lorenzo-cavazzi (approved):
:rocket:

Issue #2565, opened 2022-05-02
update architecture docs


rokroskar:
We should go through our architecture docs and update the diagrams and/or descriptions to make sure they match current reality. 

- [x] main architecture diagram 
- [ ] notebook service
- [ ] core service (cc @SwissDataScienceCenter/renku-python-maintainers)
- [ ] graph service
- [ ] gateway

olevski:
@rokroskar what do you mean by 

> diagram to explain the k8s layer components

Is this more related to how amalthea handles things? Or just the general order of the renku helm deployment in k8s?

rokroskar:
I meant all the supporting k8s infra like cert-manager, ingress, load-balancer etc.

PR #2566, opened 2022-05-02
chore: update arch docs and add mermaid

rokroskar:
Replaces outdated graphviz architecture diagram with a mermaid-based one. Installs & configures mermaid.

addresses #2565


docs/conf.py
```diff
@@ -42,6 +42,7 @@
     "sphinx.ext.todo",
     "sphinx.ext.viewcode",
     "sphinxcontrib.spelling",
+    'sphinxcontrib.mermaid',
```
lorenzo-cavazzi, line 45:
```suggestion
    "sphinxcontrib.mermaid",
```

lorenzo-cavazzi (changes requested):
Oh, I guess it's time for a review... Sorry, this slipped out of my radar :sweat_smile: 

Screenshot_20220607_141303

Looks great! And the automatically generated graph has surprisingly no arrows overlapping. This won't happen again, we should never ever change it!

The one detail I would modify is the arrow from KG to GitLab. KG is actually the only backend service that gets server-side events from GitLab through the webhook. Not sure if you want to make it explicit here.

Apart from that, just a tiny detail to update on the conf file.

rokroskar:
@lorenzo-cavazzi I made the arrow between gitlab and kg be bidirectional - if I add a second arrow going in the other direction it completely explodes the neat graph... that goes agains

> we should never ever change it!

lorenzo-cavazzi (approved):
Lgtm!

PR #2577, opened 2022-05-17
chore: avoid None in the output

rokroskar:
This prevents `None` from getting printed to stdout.


seanrmurphy (approved):
All good.

On `master`...
image

On branch `000-fix-script-typo`
image

PR #2600, opened 2022-06-03
fix(helm): use legacy bitnami repo

wesjdj:
Bitnami have removed 12k+ chart versions from their standard Helm chart repo, including the postgresql and redis chart versions we depend on for the latest renku chart. More details on that [here](https://github.com/bitnami/charts/issues/10539). In order to keep using these chart versions, we need to use Bitnami's pre-2022 versions repo.


jachro (approved):

ciyer (approved):
Thanks for reacting to this so quickly! You saved me a lot of time trying to debug k8s error messages.

PR #2601, opened 2022-06-03
build: update keycloak theme

ciyer:
Update to the new theme that is derived from the v2 theme.

/deploy #persist


ciyer:
Superseded by #2789

PR #2602, opened 2022-06-08
test: fix anon disabled

pameladelgado:
This should fix the acceptance tests for deployments where anonymous sessions are disabled.

/deploy


jachro (approved):

aledegano (approved):
Nice catch!

PR #2604, opened 2022-06-10
prep: release 0.12.13

pameladelgado:
Some guidelines for filling in:

* be specific: avoid making general sentences such as 'improvements to x'.
* think as a user: avoid using back-end terms, starting from the components name. For instance use words like: authentication, user session, file display, datasets (these could be used for both UI and Core). Also, avoid giving details of how the issue or improvement was implemented.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,27 @@
 .. _changelog:
 
+0.12.13
+-------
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (dismissed):
Lgtm!

PR #2634, opened 2022-06-28
fix(chart): set device auth polling interval to 5 seconds

### LINKED ISSUE #2599: Reduce the polling interval for Device Auth Grant in KC
Currently, the polling interval for Device Auth Grant on the `dev` deployment is set to 600 seconds which is too long and makes the CLI login unusable.

It should be reduced to 5 seconds which is the default value in KC.

mohammad-sdsc:
Fixes #2599


mohammad-sdsc:
> This closes https://github.com/SwissDataScienceCenter/renku/issues/2599, right?
> And will the existing client automatically be updated if this gets released?

Right. I've tried it and the current value is overridden with the new value in this PR.

rokroskar (approved):

PR #2638, opened 2022-07-01
fix(networkpolicy) Grant post-install-gitlab job access to Postgres.

aledegano:
This fix an issue introduced by #2613 resulting in the Post-install-gitlab
job not being able to access Postgres directly, which is needed.


olevski (dismissed):

aledegano:
> @aledegano I wonder if we should move the `post-install-gitlab` allow section into the if statement for gitlab?

ah yes! of course! Thanks, I've changed that.

olevski (approved):

PR #2639, opened 2022-07-01
fix(network-policies) Fix the correct key for the Gitlab pod name.

aledegano:
Empty descriptions cause the CI pipeline to fail.


aledegano:
@olevski sorry, one more fix...

olevski (approved):

PR #2652, opened 2022-07-06
chore: prep 0.14.1

pameladelgado:
Prepare new release with new minor version of GitLab to 14.10.5.
Merging this will upgrade dev's renku GitLab.

### DIFF
--- a/helm-chart/renku/Chart.yaml
+++ b/helm-chart/renku/Chart.yaml
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku platform
 name: renku
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.13.0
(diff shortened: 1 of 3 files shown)


helm-chart/renku/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku platform
 name: renku
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.13.0
+version: 0.14.1
```
aledegano, line 6: where is `0.14.0`?

ciyer (approved):
👍

helm-chart/renku/Chart.yaml
rokroskar: this gets set automatically when the chart gets published (by chartpress) so it's not strictly necessary to have the correct value here

PR #2654, opened 2022-07-12
feat: Add URL for KG OpenAPI spec

travissarles:
feat: Add URL for KG OpenAPI spec


travissarles:
> You can access the deployment of this PR at https://ci-renku-2654.dev.renku.ch

It is not. We're in the process of making it public. I suppose we can wait until it is.

rokroskar:
Yes this won't work until it's exposed.

jachro (approved):

rokroskar:
@jachro @travissarles I'd like to deploy this to test before we merge if that's ok

jachro:
> @jachro @travissarles I'd like to deploy this to test before we merge if that's ok

Sure.

rokroskar:
You have `renkulab.io` hard-coded in the api spec - that should not be the case: 

image

We can merge this since the change has to happen elsewhere, but the swagger page is not currently usable.

rokroskar (approved):

jachro:
Yes, we noticed that. Travis is preparing a quick fix for this.

rokroskar:
Another thing I noticed was that even though I provided a token, it was not added to the request.

rokroskar:
This doesn't need acceptance tests to pass, does it?

jachro:
That's true. Let me take the `/deploy`, then.

PR #2656, opened 2022-07-15
fix(helm): add port annotation to gitlab service

seanrmurphy:
/deploy extra-values=gitlab.enabled=true,gitlab.sharedRunnersRegistrationToken=f3acdfc1435a622e80cf97dcefa6b7c0f015d29d036e507c3c784467fcafd762


seanrmurphy:
Deploying this via github actions was not really possible as it was necessary to install gitlab.

I set up a deployment using documentation in SDSC-dev-docs (which took quite a while) but eventually all services came up.

The annotation on the port was set up correctly as you can see from the screenshot.

image

@aledegano - can you have a look at this to see if you can see any issues. It would be really good to close this point in this sprint so that we will have cleaner alerts in the next cycle.

aledegano (approved):
LGTM!

rokroskar:
When you merge this, note that some gymnastics are required when a new gitlab chart version is created. So when it comes time to do the release, the gitlab reference in `requirements.yaml` should be made that refers to the _future_ tag (presumably this should be done in the PR that sets up the release). Once the tag + release are made, the renku chart packaging will fail because the gitlab chart with that tag won't exist yet. The gitlab chart _with the just released tag_ should be pushed manually. After that the renku chart packaging and deployment should work.

PR #2659, opened 2022-07-20
fix(netpol): modify postgres network policy to add prometheus access

seanrmurphy:
Currently, our `postgres-ingress` network policy does not permit prometheus to access postgres metrics which are served on port 9187. This generates alerts within our monitoring infrastructure; we need to modify the netpol to enable this access and reduce alert noise.

This fix has been applied in our production environments and as such has been tested there.

I have made a deployment to my own namespace in the dev cluster (`sean`) which contains this change and the netpol was set correctly -  I will keep this deployment up so folks can review.

It's not super cool to somehow hard code the monitoring namespace into the helm chart; I guess specifying the app as prometheus should not be problematic. In any case, I would like this to be reviewed quickly so we can merge into the next release so that I don't have to do hotfixes to our production environments to minimize noise on our alerts; even if the solution is suboptimal, we can devise a better solution in the subsequent sprint.

@aledegano - if you could check it out, given that you're prob most familiar with netpol issues, it would be great - thanks!


helm-chart/renku/templates/network-policies.yaml
```diff
@@ -56,6 +56,14 @@ spec:
           namespaceSelector:
             matchLabels:
               kubernetes.io/metadata.name: {{ .Release.Namespace }}
+        - podSelector:
```
aledegano, line 59: In this case Prometheus is not a component of Renku, thus we cannot guarantee that it will be installed alongside Renku, which labels its pod would have and in which namespace it would run.
Luckily there is an easy fix: since the port 9187 is only used for metrics we can allow any pod to connect to it.

seanrmurphy: I modified the policy as suggested.

I tested it in the deployment in the `sean` namespace on the dev cluster. This is what the netpol looks like there:

image

aledegano (approved):

PR #2643, opened 2022-07-04
build: unify ui and ui-server charts

ciyer:
Combine the charts into one chart for the UI.

/deploy renku-ui=master #persist

PR #2665, opened 2022-07-25
feat: modify charts and tests for new UI release

lorenzo-cavazzi:
/deploy #persist

### DIFF
--- a/scripts/generate-values/base-renku-values.yaml.template
+++ b/scripts/generate-values/base-renku-values.yaml.template
@@ -88,3 +88,16 @@ postgresql:
   postgresqlPassword: <use `openssl rand -hex 32`>
 redis:
   password: <use `openssl rand -hex 32`>
+ui:
+  baseUrl: https://{renku_domain}
+  gatewayUrl: https://{renku_domain}/api
+  client:
+    uiserverUrl: https://{renku_domain}/ui-server
+    welcomePage:
+      text: |
+        ## Welcome to Renku!
+        Renku is software for collaborative data science.
+        With Renku you can share code and data, discuss problems and solutions, and coordinate data-science projects.
+        ## Template
+        I am templateable, so deployment specific information can be put here by
+        changing ui.client.welcomePèage.text in the values file.
--- a/scripts/generate-values/base-renku-values.yaml.template
+++ b/scripts/generate-values/base-renku-values.yaml.template
@@ -89,10 +89,7 @@ postgresql:
 redis:
   password: <use `openssl rand -hex 32`>
 ui:
-  baseUrl: https://{renku_domain}
-  gatewayUrl: https://{renku_domain}/api
   client:
-    uiserverUrl: https://{renku_domain}/ui-server
     welcomePage:
       text: |
         ## Welcome to Renku!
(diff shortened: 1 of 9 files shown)


olevski:
@lorenzo-cavazzi can you isolate the fix to the acceptance tests from the ui chart update? Because it seems to me that this PR tries to do both in one. But perhaps I am confused.

lorenzo-cavazzi:
Sorry for the mess, this includes the first 3 commits from here https://github.com/SwissDataScienceCenter/renku/pull/2643 and the others are the fixes -- but it works only with the latest (yet unreleased) UI version.
I'll try to clean up tomorrow if all the tests pass (both here and in the UI PRs :crossed_fingers: )

olevski:
@lorenzo-cavazzi no problem. I just wanted to make sure it wasn't accidental. You dont have to bother cleaning up. We can just merge this after the other PR you mentioned gets merged.

olevski (dismissed):

olevski (approved):

scripts/generate-values/base-renku-values.yaml.template
```diff
@@ -88,3 +88,16 @@ postgresql:
   postgresqlPassword: <use `openssl rand -hex 32`>
 redis:
   password: <use `openssl rand -hex 32`>
+ui:
+  baseUrl: https://{renku_domain}
+  gatewayUrl: https://{renku_domain}/api
+  client:
+    uiserverUrl: https://{renku_domain}/ui-server
```
rokroskar, line 95: I don't see this set in the default values in the renku chart - why is it set here? Seems like this default should either be set on the component chart or at least on the renku chart.

rokroskar: But more importantly, can't this always be built from the information contained in the `ui.server.serverData` values? Since the two charts are now unified, you shouldn't need to declare this manually.

lorenzo-cavazzi: Do you mean we should add it (commented) to `helm-chart/renku/values.yaml` ? Or we should drop it since it's set as a default in the UI values? If it's the latter, then we should probably drop  `baseUrl` and `gatewayUrl` too since the default is already derived from `ui.protocol` and `.Values.global.renku.domain`.

lorenzo-cavazzi: > can't this always be built from the information contained in the ui.server.serverData values?
Yes, we can unify some of the values from `client` and `server` but that requires a bit more work since the URLs are built in different ways and it would make sense to unify the logic -- should I make an issue in the UI to take care of that?

rokroskar: yes please do, there is certainly an opportunity here to reduce redundancy

pameladelgado (dismissed):
🚀

scripts/generate-values/base-renku-values.yaml.template
rokroskar: drop as much as possible from this template (but make sure it still yields a usable values file) because it's meant as a first exposure for someone deploying renku

lorenzo-cavazzi: This is the issue in the UI https://github.com/SwissDataScienceCenter/renku-ui/issues/1976

I removed all the values except the `welcomePage` since they are not strictly needed.

rokroskar (approved):

PR #2668, opened 2022-07-26
chore(gitlab): update gitlab chart - bump container image version, change redis chart repo, bump chart version

seanrmurphy:
There was an issue with v0.6.0 of this helm chart which impacted how prometheus was scraping data from gitlab; that is the primary driver for the new release of this chart (#2656). 

It was also noticed that the dependent redis chart must be changed due to the issues with bitnami changing how they publish artifacts and also the gitlab container image was a bit out of date.

This has been deployed within a dedicated namespace, with the renku gitlab dep pointing at this local version 0.6.1 of the gitlab helm chart.

### DIFF
--- a/chartpress-gitlab.yaml
+++ b/chartpress-gitlab.yaml
@@ -1,0 +1,25 @@
+# this file should be used as and when necessary to build the renku gitlab helm
+# chart. As renku moves to use of the standard gitlab cloud native chart, this
+# will become redundnat but for now it is still in use.
+#
+# To use this, you will need to remove the chartpress.yaml for building renku and
+# replace it with this file (it seems chartpress cannot work with a file other than
+# one called chartpress.yaml).
+#
+# To build and push a release candidate chart do this:
+# 
+# chartpress --publish --tag x.y.z-rc.Q
+#
+# To publish a chart which is not a release candidate do this, making sure that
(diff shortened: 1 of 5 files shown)


chartpress-gitlab.yaml
```diff
@@ -0,0 +1,25 @@
+# this file should be used as and when necessary to build the renku gitlab helm
+# chart. As renku moves to use of the standard gitlab cloud native chart, this
+# will become redundnat but for now it is still in use.
+#
+# To use this, you will need to remove the chartpress.yaml for building renku and
```
rokroskar, line 5: this seems sub-optimal - can we contribute a `--file` option to the upstream?

seanrmurphy: Sounds v doable - prob just adding an arg with a default and prob a test.

I can look into it.

pameladelgado (commented):
It looks good to me, but I kind of prefer also to have the commented section in the chartpress file.

chartpress-gitlab.yaml
rokroskar: I made https://github.com/jupyterhub/chartpress/issues/170

rokroskar:
what's the status here @seanrmurphy ?

seanrmurphy:
Closing - superceded by #2820.

PR #2696, opened 2022-09-15
chore: release 0.17.1

ciyer:
This creates a version of 0.17 that can be deployed (0.17.0 contains a bug that was found after tagging but before deploying).

/deploy #persist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,11 @@
 .. _changelog:
 
+0.17.1
+------
+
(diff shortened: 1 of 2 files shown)


jachro:
I guess you'd need to bump the version in the `Chart.yaml` so the machinery issuing PRs for the terraform repo can do things correctly.

jachro:
@ciyer I think you need to bump the version in the `Chart.yaml` to make the machinery issuing PRs for the terraform repo work fine.

jachro (approved):
👍

PR #2706, opened 2022-09-23
Release 0.18.0

ciyer:
Updates to changlog for 0.18.0

/deploy #persist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,23 +1,15 @@
 .. _changelog:
 
-0.18.0 (pre-release)
---------------------
+0.18.0
+------
 
-This release brings improvements and bug fixes to Renku UI.
+This release brings improvements and bug fixes to renku-ui and renku-notebooks.
 
 Features
 ~~~~~~~~
 
 * **UI** reduce unnecessary 'project locked' notifications (`#1982 <https://github.com/SwissDataScienceCenter/renku-ui/issues/1982>`_, `#2025 <https://github.com/SwissDataScienceCenter/renku-ui/issues/2025>`_)
 * **UI** expose project metadata as JSON-LD (`#1867 <https://github.com/SwissDataScienceCenter/renku-ui/issues/1867>`_, `#2022 <https://github.com/SwissDataScienceCenter/renku-ui/issues/2022>`_)
-
-Bug fixes
-~~~~~~~~~~
-[fill out or delete before release]
-
-Misc
-~~~~
-[fill out or delete before release]
 
 
 Individual components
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -1,5 +1,41 @@
...
+
+Individual components
+~~~~~~~~~~~~~~~~~~~~~~
+
+- `renku-ui 2.9.0 <https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/2.9.0>`_
+
+Upgrading from 0.17.x
+~~~~~~~~~~~~~~~~~~~~~~
+
+BREAKING CHANGES!
```
olevski, line 31: We need to mention that after upgrading, this makes any currently running anonymous sessions unavailable. This occurs when going from 0.16.x to 0.17.x as things stand right now.

This happens because the ui-server is now setting the anonymous user id and the old user ids set by the gateway are ignored. And any existing anonymous user id cookies are ignored.

I tested this by deploying 0.16.0 (same as on renkulab.io), starting anonymous sessions and upgrading to the latest on renku's master. After I did this I could not see the anonymous sessions anymore at all in the browser.

olevski (approved):

CHANGELOG.rst
olevski: ...This was supposed to be request changes. I misclicked.

rokroskar: what is the default time limit we have on anonymous sessions? Maybe we can just disable them for that amount of time before the upgrade.

jachro: So actually, that's also a problem on the 0.17.2 we are about to roll out. Is that a showstopper?

ciyer: Yes, this will also occur on 0.17.2, but we decided that this is not a problem as long as we do not roll the upgrade out during a workshop of live tutorial session.

We should probably check that there are only at most a few anonymous sessions when make the upgrade.

helm-chart/renku/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku platform
 name: renku
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.17.2
+version: 0.18.0
```
olevski, line 6:
```suggestion
version: 0.0.1-set.by.chartpress
```

This has been updated manually for a while now. But we should leave it to a fixed value like this and not touch it. When we release the helm chart chartpress will assign the right tag.

olevski (changes requested):
Sorry @ciyer one small thing. Everything else looks good.

Thanks for adding the notebooks to the changelog also.

olevski (approved):

PR #2709, opened 2022-10-04
feat(docs): add CLI plugins section to reference

rokroskar:
adds CLI plugins to main docs


.github/workflows/pull-request-test.yml
```diff
@@ -33,8 +33,8 @@ jobs:
           SIDEBAR: 1
         run: |
           cd docs
-          sphinx-build -nNW -b spelling -d _build/doctrees . _build/spelling
-          sphinx-build -qnNW . _build/html
+          sphinx-build -NW -b spelling -d _build/doctrees . _build/spelling
+          sphinx-build -nNW . _build/html
```
Panaetius, line 37: Was this just for debugging?

docs/spelling_wordlist.txt
```diff
@@ -1,9 +1,10 @@
-acyclic
-admin
+onacyclic
```
Panaetius, line 1: is this a word?

docs/spelling_wordlist.txt
```diff
@@ -88,6 +89,7 @@ interoperability
 io
 ipynb
 iteratively
+IWorkflowProvider
```
Panaetius, line 92: It's weird that you had to add this (and some others, like `str` and `cls`), since we don't need them in the rp docs.

docs/spelling_wordlist.txt
```diff
@@ -233,3 +239,56 @@ yaml
 yml
 Zenodo
 連句
+cls
```
Panaetius, line 242: Probably should sort this file alphabetically.

Panaetius (changes requested):

docs/spelling_wordlist.txt
rokroskar: I will make a follow-up PR to fix all of this in the renku-python repo

docs/spelling_wordlist.txt
rokroskar: maybe this was an issue with `nitpick_ignore` not working... I'll check

docs/spelling_wordlist.txt
rokroskar: I left these at the bottom on purpose so i know where to begin for fixing them in renku-python

rokroskar: ok I went ahead and sorted it anyway in case something unpredictable happens with this follow-up PR :)

docs/spelling_wordlist.txt
rokroskar: indeed those were not needed. Probably triggered before I setup the google docstring style.

docs/spelling_wordlist.txt
rokroskar: oops that was indeed a weird one.

Panaetius (approved):
Thank you for looking into this!

PR #2711, opened 2022-10-05
chore(ci): Update renku actions to v1.2.0

RenkuBot:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #2712, opened 2022-10-05
chore(gh-actions): Upgrade rollout-renku-version action.

aledegano:
Upgrade rollout-renku-version to v1.2.0.


aledegano:
let's go with the automated PR #2711

PR #2713, opened 2022-10-07
fix(tests): increase sleep for finding the terminal

olevski:
The tests are timing out because they are trying to click the terminal icon and eventually succeed. But this takes ~ 45 mins.

I think the problem is that when the terminal is not found then whole page is reloaded and we look for the terminal again. But the sleeps in this process are not long enough to fully download assets and render stuff. But if we do not reload the page and keep retrying the page should just load eventually and the step should succeed.

### DIFF
--- a/acceptance-tests/src/test/scala/ch/renku/acceptancetests/workflows/JupyterNotebook.scala
+++ b/acceptance-tests/src/test/scala/ch/renku/acceptancetests/workflows/JupyterNotebook.scala
@@ -60,7 +60,7 @@ trait JupyterNotebook extends Datasets with Project with KnowledgeGraphApi {
 
     `try few times before giving up` { _ =>
       When("the user clicks on the Terminal icon")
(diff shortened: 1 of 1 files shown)


olevski:
@ciyer reported that sometimes the sessions needs to be reloaded to find css elements like the terminal because the rendering or loading of resources gets stuck. 

So I increased the sleep in between retries for finding the terminal. I am too scared to mess with the sleep in the `try few times before giving up` function because it is used in many different places but only this step currently has problems.

jachro (approved):
Yes, I think it's a good way of satisfying both the needs that were mentioned in the PR. Waiting longer here cannot break anything (except we won't hit the global timeout set on the tests) and only can make the test run a bit longer.

olevski:
Here is an example of a passing test with this change: https://github.com/SwissDataScienceCenter/renku-notebooks/actions/runs/3204508502/jobs/5236505066

PR #2741, opened 2022-10-18
docs: refactor gitlab documentation

wesjdj:
Refactoring documentation on GitLab to provide a main section with GitLab information


olevski (approved):
Thanks!

olevski (changes requested):
@wesjdj the change in the helm chart is good. Just please update the changelog and the docs explaining this change and its implications.

olevski (changes requested):
Thanks Wes this is great. I got a bit picky on some parts.

But most of the things are suggestions and you can simply accept them or accept and modify them.

There are only a few things that require additions or more changes and these are not hard.

docs/how-to-guides/admin/gitlab.rst
```diff
@@ -0,0 +1,107 @@
...
+    read_repository (Allows read-access to the repository)
+    read_registry (Grants permission to read container registry images)
+    openid (Authenticate using OpenID Connect)
+
+Copy the secret and client ID and paste it to the dedicated places in renku-values.yaml (step 3)
+
+
+GitLab deployed as part of Renku
+--------------------------------
+
```
olevski, line 88: There is no explanation at all on how to setup Renku as an identity provider to Gitlab. Please add that.

Explain that when you use the Gitlab deployment that comes in Renku this is done for you. But if you use the helm chart then you have to change some things. If you can avoid too many details here and just link to the relevant gitlab documentation page that is the best.

But still I believe that the setup of this is not complicated. But it is definitely worth pointing out.

docs/how-to-guides/admin/gitlab.rst
```diff
@@ -0,0 +1,107 @@
...
+#. Log in as the root user using the password from ``gitlab.password``.
+#. Modify any users you want to modify (e.g. to make them admin).
+#. Turn the ``autoSignIn`` setting back on and redeploy.
+
+
+Migrating to an external GitLab
+-------------------------------
+
+
+If you migrate from the Renku-bundled GitLab to the cloud-native GitLab Helm chart,
```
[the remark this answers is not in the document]
wesjdj: I have a whole other page on migrating from Omnibus to cloud-native Helm chart I'm working on in terraform-renku, and that I want to refine as I actually do the migrations and get experience doing them. I think I will hold off adding anything more here until that is done and I can confidently provide the steps

docs/how-to-guides/admin/gitlab.rst
```diff
@@ -0,0 +1,107 @@
...
+     * Use Renku's Keycloak as IdP for GitLab
+     * Have separate Renku and GitLab identities
+
+
+Configuring an External GitLab as Identity Provider for Renku
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the Renku deployment is relying on an existing GitLab instance, this instance could be configured as an identity provider for Renku's Keycloak. This way all of the users from the existing GitLab instance can log into Renku with their existing login information.
+
+#. Go to \https://<renku-domain>/auth (or to your stand-alone Keycloak dashboard), login to ``Admin console`` using ``admin`` as username and the decoded password stored in ``keycloak-password-secret``.
```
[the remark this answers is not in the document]
wesjdj: I'm going to leave the last bit out because that is already covered in the Deploying Renku section and not relevant to GitLab specifically

wesjdj:
I have moved the changes for the documentation into #2750

docs/how-to-guides/admin/gitlab.rst
```diff
@@ -0,0 +1,240 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the Renku deployment is relying on an external GitLab instance, 
+this GitLab instance could be configured as an identity provider for Renku's Keycloak. 
+This way the Renku login screen will show the option to "Login via GitLab" 
+and existing GitLab users can use Renku without creating a separate Renku account.
+
+#. Login to the admin console of Renku's Keycloak, usually found at \https://<renku-domain>/auth, using the username and password stored in the Kubernetes secret called  ``keycloak-password-secret``. Ensure you ``base64`` decode these values before using them to login. If you use an external Keycloak with Renku, login to the external Keycloak instead.
+#. Add an ``Identity Provider`` of type ``OpenID Connect v1.0``.
+#. Set ``Alias`` to <renku-domain>, ``Authorization URL`` to \https://<gitlab-domain>/oauth/authorize, ``Token URL`` to \https://<gitlab-domain>/oauth/token and ``Client ID`` and ``Client Secret`` to the respective values copied from the Renku GitLab application details.
```
seanrmurphy, line 78: Could these be bullet points? Also, there are 2 things missing which I needed to specify:

image

image

(I'm not sure if I picked the correct values...)

olevski: @seanrmurphy I am not sure if gitlab supports client secret sent as JWT. I have always done it with "client secret sent as post". When you select "client secret sent as post" then you do not have to pick a signature algorithm.

Sending the secret as JWT is definitely safer but I am not sure if this is possible from the Gitlab side. @wesjdj did you see anything about this in the helm chart setup? We should try this and ensure it works if we want to put it in the docs. You can easily add this option to a running dev deployment and check if you can login via gitlab.

olevski: Either way we should be clear which client authentication method should be used. Or at least recommend one as preferred.

docs/how-to-guides/admin/gitlab.rst
```diff
@@ -0,0 +1,240 @@
...
+#. If you have a value set at ``global.gitlab.urlPrefix`` change it from ``/gitlab`` to ``/``
+#. Set ``gitlab.enabled`` to ``false``.
+#. Re-install the Renku Helm chart with the newly modified values.
+
+Troubleshooting
+^^^^^^^^^^^^^^^
+
+It is possible there will be some issues post-upgrade.
+
+If you can sign in normally to GitLab using a Keycloak account, but Renku still shows the ``RenkuLab is down`` message, try deploying a Renku instance with a newly generated values file to a different namespace and see if the issue persists. If the new Renku instance is able to launch correctly, there is probably an issue with your Renku configuration, in particular the connection to GitLab.
```
olevski, line 240: I would remove this whole `Troubleshooting` section. It is short and it may not be feasible for people to deploy Renku again in a different namespace.

olevski (changes requested):

docs/how-to-guides/admin/gitlab.rst
olevski: @seanrmurphy if you use the `client secret sent as post` method for `Client Authentication` then you do not have to select a signature algorithm. That is the only one I have used when I have set this up for my dev deployments. Sending the secret as JWT is safer but I am not sure if it will work.

If we want to add the JWT option to the docs we should try it first to make sure it works. You can easily try if the JWT method will work on a ci dev deployment.

Anyhow we should definitely specify in the docs, right in this location which `Client Authentication` method should be used.

docs/how-to-guides/admin/gitlab.rst
```diff
@@ -0,0 +1,240 @@
...
+
+#. Run ``gitlab-rails console -e production`` (this might take a while).
+#. Find the user you would like to grant the admin role, for example by running ``user = User.find_by(email: 'renku@renkulab.io')`` or ``user = User.find_by(username: 'renku')``.
+#. Grant the user the administrator role by running ``user.admin = true``.
+#. Save the user's profile by running ``user.save!``.
+#. Leave the console by running ``exit``.
+
+Migrate from Renku-bundled Omnibus GitLab to cloud-native Gitlab Helm chart
+---------------------------------------------------------------------------
+
```
olevski, line 100:
```suggestion

Migrating from the Renku-bundled Gitlab to the cloud the cloud-native Helm chart should be done only after the existing Gitlab deployment is fully and properly backed up. Doing the migration without having a fresh backup of the existing Gitlab has a very high risk of irrecoverable loss of user data and code.

```

wesjdj: I'm not sure we need to add this, as the backup is part of the migration and is covered on lines 165-187

PR #2743, opened 2022-10-19
chore: add global debug flag

Panaetius:
Mainly used to enable/disable automounting the serviceaccount token in components, can be used for other debug features in the future.

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -127,6 +127,10 @@ global:
     customCAs: []
       # - secret:
 
+  ## Set to true to enable debug mode
+  ## Individual components can handle debug mode as makes sense for them
+  debug: false
+
 ## Ingress configuration
 ## See: https://kubernetes.io/docs/concepts/services-networking/ingress/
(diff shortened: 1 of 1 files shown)


rokroskar:
I'm not sure about this tbh. "debug" doesn't scream "mount the service token" to me. Why not make it a flag that says what it does?

lorenzo-cavazzi:
Every repository could use this `debug` flag to make it debug-friendly -- whatever that means for the specific repo (caveat: I guess this is more for development than debugging tho, not sure if we should pick another name).
That would simplify deploying a dev-friendly environment without the need to remember a bunch of separate values that make no sense to activate alone.

rokroskar:
Sure that's fine, but mounting the service token seems very specific and not obviously a debug mode.

rokroskar:
I would expect, at the very least, this flag to set e.g. logs to debug. But it doesn't do that...

rokroskar:
closing this since it's been open for ~1 year and seemingly not a priority to resolve.

PR #2769, opened 2022-11-03
fix(tests): remove gitlab string from gitlab urls

andre-code:
PR to update gitlab url used in acceptance tests.

/deploy #persist


jachro:
Fixed by #2770

PR #2776, opened 2022-11-06
chore(tests): add cypress tests for rstudio

olevski:
Currently we do not test anything about rstudio.

This fixes that.

I want to keep this in the renku repo only for a bit to make sure the tests are not flaky and perform as expected. Then I can make a dedicated action for this in the renku actions repo.

/deploy #persist


olevski:
Additional features:
- screenshots of error uploaded as github artifacts when the test fails
- video of the test is uploaded as github artifact regardless of the test fails or passes
- artifacts retained for 7 days for screenshots and 3 days for videos (can be easily changed)
- total runtime = 2-3 minutes (most of this is waiting for an image to build or the sessions to start)
- tests the session in the iframe view - not in a separate tab (although testing with a separate tab is possible too)

olevski:
The tests already are catching a problem that I saw when developing them with rstudio. Namely sometimes when I open a session for rstudio in an iframe some of the elements from the rstudio are gone. Everything is functional but certain panels (the one for looking at files and variables) are just not there.

This is what things should look like:
image


But sometimes after you open (or re-open) the session with the iframe view things look like this:
image

This does not seem to happen consistently. Also it seems to not happen at all if you open the session in a dedicated tab not in the iframe view.

cypress-tests/cypress/fixtures/example.json
```diff
@@ -0,0 +1 @@
+{}
```
lorenzo-cavazzi, line 1: Is this a placeholder or a leftover?
I would either remove it or add a comment about the purpose

cypress-tests/cypress/e2e/rstudioSession.cy.ts
```diff
@@ -0,0 +1,112 @@
...
+import { registerCustomCommands } from "../support";
+
+const email = Cypress.env("TEST_EMAIL")
+const username = Cypress.env("TEST_USERNAME")
+const password = Cypress.env("TEST_PASSWORD")
+const firstName = Cypress.env("TEST_FIRST_NAME")
+const lastName = Cypress.env("TEST_LAST_NAME")
+const projectName = `test-project-${uuidv4()}`
+
+describe('Basic rstudio functionality', () => {
```
lorenzo-cavazzi, line 12: This a minor detail, but would you mind switching all the `'` to `"` just for coherence? Some in a few other place on this file.
At some point we could add a linter (eslint?) for this and other details, but it's definitely not urgent nor necessary in this PR.

cypress-tests/cypress/e2e/rstudioSession.cy.ts
```diff
@@ -0,0 +1,112 @@
...
+  })
+  beforeEach(() => {
+    // Prevents the ui from automatically refreshing the page, it does this when the local storage is wiped
+    // and the local storage is wiped by default after an "it" section ends in cypress
+    cy.restoreLocalStorage()
+  })
+  it('Creates a project', { defaultCommandTimeout: 30000 }, () => {
+    const templateName = "Basic R"
+    cy.visit("/")
+    cy.get('[data-cy="username-home"]').should("include.text", username)
```
lorenzo-cavazzi, line 62:
```suggestion
    cy.get(`[data-cy="username-home"]`).should("include.text", username)
```
In cases where `'` and `"` are nested, we could use the ` char

cypress-tests/cypress/e2e/rstudioSession.cy.ts
```diff
@@ -0,0 +1,112 @@
...
+  })
+  it('Launches a session', () => {
+    cy.startSession(username, projectName)
+  })
+  it('Waits for the session to fully start', { defaultCommandTimeout: 300000 }, () => {
+    cy.get('.time-caption').should("contain.text", "Running")
+  })
+  it('Opens the session in an iframe', () => {
+    cy.contains("Open").click()
+    cy.get('div.details-progress-box', { timeout: 300000 }).should("not.exist")
```
lorenzo-cavazzi, line 76: This `300000` ms timeout is repeated a few times. Would it makes sense to have a couple of variables so that future extensions will use similar timeouts fo long/unstable operations? Something like:
```
const TIMEOUTS = {
    "standard": 10000,
    "long": 30000,
    "vlong": 300000
}
```

lorenzo-cavazzi (changes requested):
Adding a set of Cypress-based tests is great, thanks for making an effort to start this!
I added a few minor comments inline.

A generic note: is it necessary to run them on both `renku-dev-test.yaml` and `pull-request-test.yml`?

P.S: I also feel some of the `it` sections in the "rstudioSession.cy.ts" file might be merged together, but that's a minor thing and I guess it's just personal taste.

cypress-tests/README.md
```diff
@@ -0,0 +1,51 @@
...
+  through any but the Renku login screens.
+- The tests expect to use the one set of provided credentials to log in
+  to up to 2 different Renku deployments. In the case where the Gitlab used by
+  Renku is part of another deployment the tests will try to log in twice.
+- Using `https` if possible is reccomended in the `BASE_URL` environment
+  variable. Completing the registration and logging in fails if the `http`
+  is used on the standard dev and/or CI Renku deployments.
+- Using the electron browser from cypress will not work because rstudio does not load at
+  all in the electron browser.
+- The tests need slight adjustments to the setup to work in Firefox so
```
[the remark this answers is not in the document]
olevski: It does not work. I have to spend some time playing with it to see how (and if we can fix it).

For now it is best to say it does not work. It is not clear what (if anything) should be done to fix this.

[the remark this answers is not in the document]
olevski: I adjusted the instructions

olevski:
> A generic note: is it necessary to run them on both renku-dev-test.yaml and pull-request-test.yml?

This was there already. I just looked and added the cypress tests wherever the acceptance tests are. The `renku-dev-test.yaml` workflow has a `repository_dispatch` trigger. This is a webhook trigger that is used when we publish a new renku helm chart. And now we publish a new helm chart on every merge to master. So this webhook is called then. This allows us to run the tests on every helm chart release. I think this is from YAT. I pinged them to ask whether it is still needed.

olevski:
Also as @ciyer mentioned I updated to cypress v11.

lorenzo-cavazzi (approved):
Lgtm!

PR #2786, opened 2022-11-08
test: GH action to verify tests compile

jachro:
This PR introduces a new GH action that checks if the Acceptance Tests code still compiles. The reason for the change is to have a smoke test that verifies if the tests are still OK after a dependencies update.


eikek (approved):
🙏🏼

PR #2789, opened 2022-11-09
build: update keycloak theme

ciyer:
Update key cloak theme to unify appearance with the RenkuLab UI.

/deploy #persist

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -233,7 +233,7 @@ keycloak:
 
   extraInitContainers: |
     - name: theme-provider
-      image: renku/keycloak-theme:3.0.2
+      image: renku/keycloak-theme:3.0.6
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi:
This works well!
Just one detail: I've noticed some errors in the registration page show a white text instead of red. Here a couple of examples

image
```
<span id="input-error-email" class="pf-c-form__helper-text pf-m-error required kc-feedback-text" aria-live="polite">
    Email already exists.
</span>
```
image

It would be nice to have them styled as the following:
image

ciyer:
You are right! It looks like that problem has actually been there for a while. I have a fix for it here: https://github.com/SwissDataScienceCenter/keycloak-theme/pull/19

lorenzo-cavazzi (approved):
This works well! :tada:

lorenzo-cavazzi (approved):

PR #2804, opened 2022-11-16
fix(ingress): Fix frame-ancestors CSP on the Renku ingress.

aledegano:
Also allow `*.<domain>` to frame pages so it can work with CN Gitlab,
the upcoming user-sessions domains and other special deployments (e.g. dev.renku.ch).

Fix SwissDataScienceCenter/terraform-renku#845.

/deploy #persist


aledegano:
@Panaetius FYI

Panaetius (approved):
looks good, thanks!

PR #2812, opened 2022-11-18
tests: update selector for notebook cells

ciyer:
Update the selector used to find notebook cells to be compatible with the new notebook renderer.


lorenzo-cavazzi:
Included in #2798

PR #2818, opened 2022-11-22
chore(gitlab): remove redis from gitlab chart

seanrmurphy:
/deploy #persist

### DIFF
--- a/helm-chart/gitlab/Chart.yaml
+++ b/helm-chart/gitlab/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: A Helm chart for the Renku Gitlab server
 name: gitlab
-version: 0.6.0
+version: 0.7.0-rc.2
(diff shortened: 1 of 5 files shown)


seanrmurphy:
This PR contains two specific changes to the renku-gitlab helm chart:
- the external redis is removed and the configuration is modified to use the redis service which is deployed by default inside the gitlab container
- the version of gitlab is modified from 12.9.5 to 14.10.5 which is deployed on all the clusters we manage

The main rationale for this is that it is difficult to update the redis version on the main renku helm if there is a different redis version embedded in the gitlab helm chart. Further, as we are moving to the Cloud Native version of gitlab, maintaining a production grade embedded renku-gitlab becomes less relevant.

A renku-gitlab 0.7.0 chart has already been pushed to our helm repo. If changes are required to this PR, then that will need to be updated. An upgrade has been performed on a personal cluster from renku 0.19.1 to a new renku chart which includes the gitlab 0.7.0 subchart and the redis version was updated as expected.

A subsequent PR will be created to modify the renku helm chart to incorporate the new renku-gitlab 0.7.0 helm chart

olevski (approved):

PR #2820, opened 2022-11-23
chore(gitlab): update version of renku-gitlab helm chart

seanrmurphy:
For a couple of reasons, the renku-gitlab helm chart has been updated (see [here](https://github.com/SwissDataScienceCenter/renku/pull/2818) for more info). This chart has been pressed and is served via the renku helm repo. It is now necessary to modify the renku helm chart such that it picks up this new renku-gitlab helm chart. That is the focus of this PR.

Some notes:
- I have already tested the upgrade process on a personal cluster and it works fine
- I would like someone else to go through these tests before this gets approved - see testing approach below
- Merging and rolling this out will do the following (in the case gitlab is enabled in the values):
  - remove gitlab redis
    - this may result in some data loss, mostly related to sidekiq job management
    - we have more or less done this before and we did not perceive any data loss
    - salvaging this data properly would be non-trivial and our current configuration seems to be v designed around the basic notion that all data in redis is disposable
  - restart gitlab with a new configuration in which the redis inside gitlab is used rather than an external one
    - this will require a few minutes of gitlab down time
  
How to test:
- deploy a personal cluster with renku chart `0.19.1`
  - wait until everything has come up
- register a new user on the system
  - upgrade the user to an admin user using the approach described [here](https://github.com/SwissDataScienceCenter/renku/blob/redirect-gitlab-path/docs/how-to-guides/admin/gitlab.rst#gitlab-deployed-as-part-of-renku)
- Log in as this user and go to the gitlab admin functions
- Check the version of redis in use by the gitlab instance
  - This is in Monitoring | Background Jobs
  - It should be `6.0.5` which is the same as the gl-redis container version
- Change the version of the renku chart deployed to be `0.19.2-10.30909de.git.1821.ha8299a2d` (this is one that I generated from a branch and contains the `0.7.0` gitlab chart)
- Make update
- You should see that the gl-redis container is removed
- Once gitlab comes up again, check the redis version: it should now be `6.2.6` which indicates that it is using the internal version
- Run a simple job to test that everything is working properly


wesjdj (dismissed):
tested as described, LGTM

helm-chart/README.rst
```diff
@@ -62,6 +62,29 @@ Most information related to upgrading from one chart version to another is cover
 in the `values changelog file <https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md>`_.
 For upgrades that require some steps other than modifying the values files to be executed, we add some instructions here.
 
+Upgrading to 0.22.0
+*******************
+This version changes the embedded `renku-gitlab` helm chart from `0.6.0` to
+`0.7.0`. This removes the dependency on an external `redis` in the `renku-gitlab`
+`helm` chart and makes `redis` upgradable in the main `renku` `helm` chart. It does
+mean that the embedded `renku-gitlab` `helm` chart now uses `redis` which is run
+within the `gitlab` omnibus container and this is not recommended for production
```
rokroskar, line 71: > It does mean that the embedded `renku-gitlab` `helm` chart now uses `redis` which is run
within the `gitlab` omnibus container

Presumably this is the _default_ behavior but can still be overridden to use an external redis?

lokijuhy (approved):

wesjdj (approved):

rokroskar (approved):

PR #2821, opened 2022-11-23
docs: Add Workflow File Documentation

lokijuhy:
Add content about Workflow Files to the Tutorial and Topic Guide sections.

The built docs can be found [here](https://renku.readthedocs.io/en/workflow-file-documentation/).


rokroskar (changes requested):
Thanks @lokijuhy !! It's looking pretty good! Two main things

1. make the example in the topic guide consistent between the CLI-only and workflow file part
2. reformat yaml to use 2-space indents :)

docs/topic-guides/workflows.rst
```diff
@@ -154,8 +154,174 @@ tells Renku to automatically infer dependencies between steps for you. The newly
 created ``my-workflow`` can also be executed with :meth:`renku workflow execute <renku.ui.cli.workflow>`.
 
 
-Inspecting Workflows
---------------------
+The Renku Worfklow File
```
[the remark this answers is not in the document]
lokijuhy: With the goal of making our docs more navigable, I have split this general workflows topic guide into separate pages per sub-topic. And I listed Workflow File above Workflows on the CLI.

docs/topic-guides/workflows.rst
```diff
@@ -85,8 +85,8 @@ code with different parameters and compare the results, or send it to different
 execution backends.
 
 
-Working with Workflows
-----------------------
+Working with Workflows on the Command Line
```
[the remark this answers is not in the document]
lokijuhy: I changed it to "Tracking Workflows Interactively"

docs/topic-guides/workflows/workflow-file.rst
```diff
@@ -0,0 +1,362 @@
...
+-------------------------
+
+Running :meth:`renku run workflow.yml <renku.ui.cli.run>` will execute all steps
+in the workflow file. Executing the workflow will commit all workflow inputs and
+outputs, too, including the workflow file itself.
+
+.. code-block:: console
+
+    $ renku run workflow.yml
+    Executing step 'workflows/workflow-topic-guide-last.yml::filter': 'python src/filter_flights.py data/flight-data/2019-01-flights.csv.zip data/output/flights-filtered.csv' ...
```
mohammad-alisafaee, line 173: As we discussed today, the name will change to `flights-processing-pipeline::filter`. This is not final yet since I need to decide about the separator to be `::` or `.`.

docs/topic-guides/workflows/workflow-file.rst
```diff
@@ -0,0 +1,362 @@
...
+
+..         # runs every step between 'filter' and 'count', including 'filter' and 'count' themselves
+..         $ renku run workflow.yml filter:count
+
+
+Workflow Step Execution Order
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you execute a workflow file, Renku builds an execution graph. This means
+that Renku determines how the steps in the workflow are related. For example,
```
mohammad-alisafaee, line 217: I would change this to something similar to:

`When you execute a workflow file, Renku builds an execution graph to determines how the steps in the workflow are related. Renku then executes steps in that order. This means that the order of steps in the workflow file has no effect on the execution order.`

I believe we should clearly say the last sentence and also say that the execution graph is used to determine the execution order.

docs/topic-guides/workflows/workflow-file.rst
```diff
@@ -0,0 +1,362 @@
...
+          - data/output/flights-filtered.csv
+
+.. note:: Note that the script being run is always listed in the ``inputs`` section (here, ``src/filter_flights.py``).
+
+This workflow file defines the workflow's name and a sequence of steps. This
+file only includes one step, which is named ``filter``. Within the ``filter``
+step, list the ``command`` to run, and then we tell Renku which parts of this
+command are ``inputs`` and ``outputs`` by copying those paths into the relevant
+sections.
+
```
mohammad-alisafaee, line 55: It's good to have a note or warning to say that the workflow file itself is a YAML file but with basic format. Something like the following:

```
Renku uses basic YAML syntax for workflow definition files. 
Users shouldn't use advanced YAML syntax like anchors, aliases, schema, etc. since the behavior is undefined. 
Moreover, in future we will implement a customized YAML parser that won't allow these features.
```

mohammad-alisafaee (changes requested):
Thanks Laura! This looks great! I've found some typos and added some suggestions.
The only topic which is not covered is running workflow files with a different executor (provider) by using `--provider` flag in `renku run`. This allows parallel execution of workflow steps when using `toil` for example. I also need to add this to the CLI documentation.

docs/topic-guides/workflows/workflow-cli.rst
```diff
@@ -0,0 +1,119 @@
...
+To track your code execution as a Renku workflow, simply prepend :meth:`renku run <renku.ui.cli.run>`
+in front of your command. Assigning a name makes it easier to re-run this step
+later.
+
+.. code-block:: console
+
+    $ renku run --name run-analysis python run_analysis.py input_file.csv output_file.csv
+
+You may also identify the workflow step's inputs and outputs via the ``-i`` and
+``-o`` flags, as shown here:
```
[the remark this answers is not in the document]
lokijuhy: Ha! This is part of the documentation currently- I did not write this. I will follow your suggestion of replacing it with a note about `--`.

docs/topic-guides/workflows/workflow-file.rst
lokijuhy: Good point. I added this as a note at the bottom of the Templating section, since that's where I'd guess users might look for info about syntax shortcuts.

lokijuhy:
@mohammad-alisafaee Thank you for the thorough review!

> The only topic which is not covered is running workflow files with a different executor (provider) by using `--provider` flag in `renku run`. This allows parallel execution of workflow steps when using `toil` for example. I also need to add this to the CLI documentation.

Ok, let me know when you've added `provider` to the CLI documentation, and then I'll copy over the relevant portions into the Topic Guide.

docs/topic-guides/workflows/workflow-concepts.rst
```diff
@@ -0,0 +1,85 @@
...
+data``, ``code``, and ``results``:
+
+.. graphviz::
+    :align: center
+
+    graph foo {
+        rankdir="LR"
+        nodesep=0.01
+        node [fontname="Raleway"]
+        _results [color="white", label="", image="../_static/icons/scatter_plot.svg"]
```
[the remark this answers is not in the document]
lokijuhy: I added the extra level up, and the images went from blank spots ...to blue question mark icons 😕  Do you have any ideas what I'm doing wrong?

docs/topic-guides/workflows/workflow-file.rst
```diff
@@ -0,0 +1,364 @@
...
+In the simplest version of the Renku workflow file, the command, inputs and
+outputs are simply listed, as in the example below:
+
+.. code-block:: yaml
+
+    name: flights-processing-pipeline
+    steps:
+      filter:
+        command: python src/filter_flights.py data/flight-data/2019-01-flights.csv.zip data/output/flights-filtered.csv
```
rokroskar, line 41: I believe this is too verbose of an example for the topic guide. In the next section we also include a parameter that does not actually make sense in this context (`-n 10`) so it's confusing if the user has already done the tutorial. Can we simplify?

docs/topic-guides/workflows/workflow-file.rst
lokijuhy: Updated to `.`

Panaetius (dismissed):

rokroskar (approved):

rokroskar (approved):

PR #2835, opened 2022-12-01
fix: acceptance tests not to fail when getting event processing status

olevski:
This PR brings some fixes to the acceptance tests around finding the status of events processing on the KG.

/deploy #persist


eikek (approved):
👍🏼

Issue #2797, opened 2022-11-14
Cover missing UI features with Cypress tests


lorenzo-cavazzi:
As a follow-up to https://github.com/SwissDataScienceCenter/renku/pull/2776, we could use the Cypress framework to quickly make new tests to cover the latest UI features still not covered by the Selenium acceptance tests.

If the effort is reasonable, we could try to covert and refresh some of the older tests to the new framework.

ℹ️ Please timebox the effort in this issue: for the `2022-12-15` sprint that's up to 3 days. Feel free to suggest new cases we should cover as new comments. They will be later added to the list

-----

More in details, we should add the following test cases (in order of importance):

1. login, create a public project, logout, go to home, search the project, login again. We should somehow preserve the name for the following actions, and this action should happen before all the others.
2. interact with the UI through a new project:
    - [x] check the readme is visible, check the `Welcome to your new Renku project!` banner, move to status and check the project is up-to-date
    - [x] go to "Files", open `.renku/metadata/project` file, open `README.md` file.
    - [x] go to Datasets, add a dataset, modify the dataset, search the dataset, open it again
    - [ ] go to Settings, check the clone command includes `<project_name>.git`, go to submenu Settings, change number of CPUs, go to files, open `.renku/renku.ini`, check the CPUs setting is there

Follow ups:
- [#2852](https://github.com/SwissDataScienceCenter/renku/issues/2852)
- [#2853](https://github.com/SwissDataScienceCenter/renku/issues/2853)

PR #2837, opened 2022-12-02
Additional cypress-based tests for the UI

ciyer:
Covers most of #2797, except changing the session settings with will be handled in https://github.com/SwissDataScienceCenter/renku-ui/issues/2215


lorenzo-cavazzi (changes requested):
Thank you for the effort! The tests look good!
There are a couple of unused functions, but it makes sense to keep them around until we handle the project settings page properly.

Just a minor note: could you refresh the `README.md` file? It currently says that the tests only play around with an R project.
It would also be great to include how to use the `/deploy` string to invoke the cypress tests (and maybe also a link to the updated action and point to the UI as an example of how to integrate these tests in other repos)

PR #2841, opened 2022-12-07
chore(tests): use requirements.txt for python

olevski:
A few changes to make the acceptance tests a bit easier to use/run.

1. Updated the readme to indicate how to run the tests locally (and especially how to run only a specific test)
2. Moved the python packages used in the acceptance tests in a requirements.txt file

/deploy #persist

### DIFF
--- a/acceptance-tests/requirements.txt
+++ b/acceptance-tests/requirements.txt
@@ -1,0 +1,9 @@
+papermill==2.1.3
+requests>=2.20.0
+jupyterhub==1.1.0
+nbresuse==0.3.3
+jupyterlab-git==0.20.0
+pipx>=0.15.0.0
+pandas==1.3.0
+seaborn==0.11.1
+toil==5.7.1
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,8 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 5
+  - package-ecosystem: "pip"
+    directory: "acceptance-tests/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
--- a/acceptance-tests/Dockerfile
+++ b/acceptance-tests/Dockerfile
@@ -37,16 +37,7 @@ USER attests
 ENV PATH=/home/attests/.local/bin:/home/attests/.local/lib/python3.8/site-packages:$PATH
 
 RUN python3 -m pip install --upgrade 'pip==22.3.1' && \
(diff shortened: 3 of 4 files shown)


acceptance-tests/requirements.txt
```diff
@@ -0,0 +1,9 @@
+papermill==2.1.3
```
rokroskar, line 1: This seems like a weird collection of dependencies. If we keep them here should we add them to the dependabot config?

olevski: Sure.

acceptance-tests/requirements.txt
```diff
@@ -0,0 +1,9 @@
...
+requests>=2.20.0
+jupyterhub==1.1.0
+nbresuse==0.3.3
+jupyterlab-git==0.20.0
```
rokroskar, line 5: but why do we need these for running the tests? presumably `nbresuse`, `jupyterhub`, `jupyterlab-git` are all there for jupyterlab sessions and this is just running the CLI is it not?

olevski: @rokroskar I am not sure. These were just hardcoded in the image for the tests. I did not add these to the image for the tests. Just moved them to a separate file.

rokroskar: sure I just thought we could remove them and see if anything breaks :) but lets leave it for now.

rokroskar (approved):

acceptance-tests/requirements.txt
rokroskar: ok sure if you want to do that here or in another PR. I thought since we added dependabot it's pointless to be bombarded with PRs for upgrades that we don't actually care about

rokroskar: I _think_ all it needs is pandas and papermill: https://renkulab.io/gitlab/renku-tutorial/renku-tutorial-flights/-/blob/master/.tutorial/meta/templates/00-FilterFlights.ipynb

olevski: Ok so I think that the only thing the acceptance tests need is `papermill`. Now the tests should pass. There was only one test that failed after I removed all the packages and it failed because it could not run `papermill`

olevski:
@rokroskar I tried but papermill and the hands on project require several python libraries. So it wasn't just papermill that is needed. Running just with papermill failed because the python kernel is missing. And I think the hands on tutorial also needs pandas and seaborn. So I will just put back all the packages as they were.

rokroskar (approved):

PR #2842, opened 2022-12-07
chore: add nix configuration

rokroskar:
This PR adds a nix configuration that simplifies dependency management for building charts and documentation from the main renku repository.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,51 @@
+{ system ? builtins.currentSystem, version ? null }:
+let
+  pkgs = import <nixpkgs> { inherit system; };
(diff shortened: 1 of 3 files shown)


default.nix
```diff
@@ -0,0 +1,55 @@
+{ system ? builtins.currentSystem, version ? null }:
```
eikek, line 1: Is `version` used?

eikek (dismissed):
Nice!

default.nix
rokroskar: ah no, this is a relic from an old setup I was using somewhere else...

eikek (approved):

PR #2863, opened 2022-12-16
fix: sqlite3 vulnerability CVE-2022-46908

olevski:
Fixes https://app.snyk.io/org/renku/project/4cee7bca-7cb5-481c-bf36-5455095123de#issue-SNYK-DEBIAN11-SQLITE3-3167788

/deploy #persist


olevski:
Fixing this in the acceptance tests image is really hard. Because we need to install a bunch of python packages and scala in alpine. All debian images have the vulnerable sqlite3 library.

I will see if I can get this in another PR. The acceptance tests also run only on demand so the risk from these is less than an image that runs every time or constantly.

rokroskar:
afaik we don't use the SQLite CLI anywhere?

rokroskar (approved):

PR #2873, opened 2022-12-21
chore: tests docker image to extend python:3.9

jachro:
/deploy #persist

### DIFF
--- a/acceptance-tests/Dockerfile
+++ b/acceptance-tests/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8-slim
+FROM python:3.9-slim
 
 RUN apt-get update && \
     apt-get install -y \


seanrmurphy (approved):
LGTM

seanrmurphy:
Thanks for the heads up here @jachro - since it's just for acceptance tests, I assume it's fine. If it does break things in acceptance tests, I guess that will show up quickly and in the worst case we can revert.

jachro:
> Thanks for the heads up here @jachro - since it's just for acceptance tests, I assume it's fine. If it does break things in acceptance tests, I guess that will show up quickly and in the worst case we can revert.

Yes, we'll see how it goes. I wanted to bump it event to a newer image but I'm not sure if we support Python > 3.9

PR #2889, opened 2023-01-06
Update from auto-update/renku-graph-2.26.0

RenkuBot:
This is an automated pull request but it's waiting for an arrival of a new UI that matches some tiny breaking change on one of our APIs.

/deploy


lorenzo-cavazzi (changes requested):
I believe version 0.22 has not been released yet -- that info should probably go under 0.22 instead of a new section 0.23

lokijuhy:
> I believe version 0.22 has not been released yet -- that info should probably go under 0.22 instead of a new section 0.23

This is correct

lorenzo-cavazzi:
I moved the changelogs to version 0.22, re-phrased the a bit, and added the changes from the UI since they must be merged together for the tests to pass

lokijuhy (approved):

aledegano (approved):

PR #2893, opened 2023-01-10
chore(helm): update redis to chart version 17.4.2

wesjdj:
Update Redis from chart version 10.7.11 to 17.4.2

todo:

- point at specific bitnami helm repo url
- test deploying with gitlab enabled

The bitnami redis helm chart used in renku is quite old (v10.7.11, published 9/7/2020). It contains Redis version 6.0.5 (released 6/9/2020).

Latest versions of the bitnami helm chart are 17.4.x and they contain redis version 7.x. Redis releases have moved from the 6.x series to the 7.x series, with the last 6.x release being in April 2022 and hence it makes sense to make the transition.

The newer redis helm charts operate somewhat differently from the older ones - the notion of slave has been replaced with replica and there is now a clear distinction between operating in a mode which has an explicit master or operating in sentinel mode - these are mutually exclusive. This means that the values we use to control redis are not backwards compatible and it is necessary to simultaneously perform a redis chart upgrade and change the values. Clearly this means that care should be taken with this upgrade.

In our previous redis configuration, the standard configuration was to have a single pod named master and two slave pods. In the new configuration, all three pods have essentially the same names (node-0|1|2). All three nodes contain 3 containers: redis, sentinel and a metrics container for prometheus data collection.

A service is defined called renku-redis and all three pods sit behind this; the redis ports and the sentinel ports are accessible via this service.

Currently, our services query sentinel on renku-redis to obtain the current master - this works. It is (arguably) not the standard sentinel configuration (in which a list of sentinels would be defined); however, it provides an upgrade path which is no less robust than our current solution and requires no changes to our services - it could even be more robust.

/deploy #persist


wesjdj:
The plan is to merge this not in the next renku release, but the one afterwards. This is due to the change to redis in the renku-gitlab Helm chart version 0.7.0, which will be included in the next release. The change in this PR should be in the release following the release containing renku-gitlab Helm chart version 0.8.0

wesjdj:
added changes from @seanrmurphy's previous work done on this topic in #2813

seanrmurphy (approved):
LGTM - nice work @wesjdj !

helm-chart/renku/requirements.yaml
```diff
@@ -36,6 +36,6 @@ dependencies:
   version: 0.0.3
   repository: "https://swissdatasciencecenter.github.io/helm-charts/"
 - name: redis
-  version: 10.7.11
-  repository: "https://raw.githubusercontent.com/bitnami/charts/eb5f9a9513d987b519f0ecd732e7031241c50328/bitnami"
+  version: 17.4.2
+  repository: "https://charts.bitnami.com/bitnami"
```
rokroskar, line 40: the reason for the explicit link before was that the bitnami repo broke, right? Should we not continue to follow the same pattern?

helm-chart/renku/values.yaml
```diff
@@ -326,67 +326,101 @@ redis:
...
-    #       matchLabels:
-    #         app: redis
-    persistence:
-      enabled: false
+
+  # default values in this section
+  image:
+    registry: docker.io
+    repository: bitnami/redis
+    tag: 7.0.7-debian-11-r7
```
rokroskar, line 341: why pin the tag here? Should it not come from the chart?

helm-chart/renku/values.yaml
```diff
@@ -326,67 +326,101 @@ redis:
...
+    pullPolicy: IfNotPresent
+
+  # replication is the default; standalone is also an option but this results in
+  # no read only replicas
+  architecture: replication
```
rokroskar, line 346: shouldn't the default values be a minimal set that will get the thing up and running? Replication, sentinels, etc is more suitable for production deployments and brings additional complexity/requirements

rokroskar (commented):
Looks like a nice improvement! Just a question about what the expected use-case for the default values should be...

seanrmurphy:
> Looks like a nice improvement! Just a question about what the expected use-case for the default values should be...

Not sure what you're asking here; if you're asking why are some defaults left in there, it's just to make it clear what the config is so someone who does not know the specifics of the values file can look at this and have a pretty good idea of the config without need to reference the values file to understand defaults. Also, someone can more quickly grok what is possible from this again without needing to look at the reference values.

Obviously, it's a judgement call regarding what is left in (and hence is debatable) - I guess the first q is whether you think this approach reasonable at all, or whether we should go with an approach in which defaults are removed and just picked up from the default values. If we're ok with the approach, then we can discuss if certain (default) values should be removed or other defaults added to this file.

helm-chart/renku/values.yaml
seanrmurphy: I think we don't have complete consistency here in the chart: the specific image is configured for postgres [here](https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/values.yaml#L293). For redis, I think it's sufficient that when we do a chart upgrade we should add a comment in this values file indicating what the default redis version is for that chart version. I will modify.

rokroskar:
No no, defaults are fine - I guess the main thing that jumped out at me was the sentinel config, which I assumed was _not_ a default. I do like it when defaults (that are expected to be changed reasonably regularly) are clearly made accessible in the top-level values file, makes everyone's life much easier.

helm-chart/renku/values.yaml
rokroskar: I don't doubt we have inconsistencies ;)  I suppose ideally we would take what comes from the chart, otherwise it's easy to get into unexpected situations.

helm-chart/renku/requirements.yaml
rokroskar: so was the issue that the version we had was simply too old to be found somehow?

seanrmurphy: Yes - iirc, bitnami had too many charts and it was taking ages to download indexes (indices?) - they decided to remove older charts from the index at `charts.bitnami.com` index and this caused problems in our reconciliations as the specified chart version could not be found in the `repository` - do I have that right @wesjdj ?

I think they articulated a policy of a 6 month window, but I might have that wrong...

wesjdj: You are correct @seanrmurphy, 6 month retention of charts in the Bitnami repo index.

wesjdj: What'd be ideal is if we pointed to a specific tarball for the chart version, rather than the repo and the version.

helm-chart/renku/values.yaml
seanrmurphy: Iiuc, sentinels were enabled in the older stuff by default. The redis helm chart defaults to 3 replicas and the resource requirements are typically small (cpu: 250m, memory: 256Mi are examples in the helm chart). Given that the aggregate requirements of redis overall are modest, I think we should just go with this. I can add a comment to this effect if we wanna go with this approach.

rokroskar: ok sounds good!

seanrmurphy:
@rok - I've addressed your comments from yday; I've also added changelog content - imo, we're pretty close to merging here - lemme know what you think.

There have been random test failures; I think they should pass if I just run them again.

rokroskar (approved):
LGTM thanks!

olevski (approved):

PR #2894, opened 2023-01-10
chore(helm): update postgres to chart version 12.1.8

wesjdj:
Update Postgres from chart version 9.1.1 to 12.1.8

todo:
- point at specific bitnami helm repo url
- ~~test deploying with gitlab enabled done~~ successful


aledegano:
This is a big jump in versions!
We need to evaluate how the migration will go for larger databases (e.g. Renkulab.io).
`infra-staging` could be a starting point, perhaps.

PR #2895, opened 2023-01-10
docs: add C4 system diagram

Panaetius:
Using https://github.com/plantuml-stdlib/C4-PlantUML

Following the [C4 model approach](https://c4model.com/#ContainerDiagram)

Preview [here](https://renku.readthedocs.io/en/add-architecture-diagram/reference/services/services-architecture.html#system-context)


seanrmurphy:
Some basic feedback from my side:
- for the system level view, I don't really think it's necessary to highlight postgres, redis and jena - arguably, these are components which are 'inside' Renku - I guess most folks would understand that any reasonably complex system requires its own storage, so I don't think it's necessary at this higher level view to highlight that and to highlight what technology choices have been made for db, keystore and jena stuff
- I also think from an optics pov, that the layout here should be different - I'd expect Renku to be somewhat larger and in the centre - currently, it looks like a small component in a land of beasts - personally, I'd also prefer a left to right flow but that's completely subjective
- the container diagram is quite helpful imo and gives rise to rather obvious q's about why it works like this (and whether it should be changed); that said, I understand that the purpose of this PR is to reflect current state in a sensible way and not open a fundamental architecture discussion
- I do find the labeling of `component` and `system` and the technology used for implementation a bit unnecessary, but again this is probably personal - I'm looking at it from the perspective of an architecture diagram where I want to understand the key components, what they do and what they interact with and I don't really care what technology was used for implementation
- the difference between Amalthea and renku-notebooks is quite non-obvious - I basically understand, but to any third party (or new team member), I think it would be confusing; one is labeled `interactive session management` and the other `k8s operator for scheduling sessions` - would it be clearer to change these to 'interactive user session management` and `mapping user sessions to kubernetes resources`?
- `core-service` and `renku-graph` seem to be performing the same functionality (backend service for project information) - I guess this is an oversight
- maybe `UI` component should not just be the home page, but perhaps serves static content (if that's correct)
- I find the functionality of the `UI-server` (Backend for frontend) quite confusing - I guess it's some kind of adaptor which we have for some legacy reasons - can we think of a different way to phrase this?

I did not go through the different arrows etc yet - I still need to grok what the main pieces are and only then can I start to think about how they interact.

Hope this is useful!

rokroskar:
Why is `kubernetes` off to the side and only accessed by `Amalthea`? The notebook service and the CRC service both talk to the k8s API as well afaik.

olevski (approved):

olevski:
> I find the functionality of the UI-server (Backend for frontend) quite confusing - I guess it's some kind of adaptor which we have for some legacy reasons - can we think of a different way to phrase this?

@seanrmurphy the ui-server will be moved fully behind the gateway. Currently it is occupying a very strange place and is acting very much like a 2nd gateway. But with the gateway refactoring this will change and the ui-server will just be another backend service that the gateway routes to.

Panaetius:
@seanrmurphy also BFF is a pretty established pattern (though as Tasko mentioned, it shouldn't take on gateway functionality): https://learn.microsoft.com/en-us/azure/architecture/patterns/backends-for-frontends

PR #2897, opened 2023-01-11
ci: prevent uploading cypress videos on success

lorenzo-cavazzi:
Prevent uploading videos when tests run successfully. They aren't very useful and this should reduce the runtime.


lorenzo-cavazzi:
Not working as expected.

PR #2904, opened 2023-01-16
chore: upgrade from keycloak to keycloakx chart

rokroskar:
This PR makes the change from the current `codecentric/keycloak` chart to the new `codecentric/keycloakx` chart. 

### To deploy this as an upgrade of Renku 0.21.0

If the secrets for the deployment are managed by helm, make these changes _before_ you deploy this upgrade:

In the `keycloak-password-secret` secret, update keys to be: 

- `KEYCLOAK_USER` --> `KEYCLOAK_ADMIN`
- `KEYCLOAK_PASSWORD` --> `KEYCLOAK_ADMIN_PASSWORD`

In the `renku-keycloak-postgres` secret, update the keys to be:

- `DB_ADDR` --> `KC_DB_URL_HOST`
- `DB_DATABASE` --> `KC_DB_URL_DATABASE`
- `DB_USER` --> `KC_DB_USERNAME`
- `DB_PASSWORD` --> `KC_DB_PASSWORD`

If there are any custom keycloak values passed to helm, you need to change the `keycloak` namespace to `keycloakx`.

/deploy

### DIFF
--- a/chartpress.yaml
+++ b/chartpress.yaml
@@ -31,7 +31,7 @@ charts:
       init-realm:
         contextPath: scripts/init-realm
         dockerfilePath: scripts/init-realm/Dockerfile
-        valuesPath: keycloak.initRealm.image
+        valuesPath: keycloakx.initRealm.image
         paths:
           - scripts/init-realm
           - scripts/init-realm/init-realm.py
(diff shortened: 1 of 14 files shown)


aledegano:
I have tested multiple times this upgrade and I have also added a commit that allows for the seamless upgrade from the current version of Keycloak to the one introduced in this PR.

As far as I can tell everything works.

rokroskar:
@aledegano I think we should merge this so that it does not get rolled out quietly with other application changes, just in case there are issues. We should merge + tag a release after a regular release is made and then do a separate roll out. wdyt?

aledegano:
In general I agree with the idea of releasing this on its own. I would like to have some more opinions on the way I have changed the secrets handling though because, afaik, this is the first time we try this strategy.

PR #2914, opened 2023-01-20
Update from auto-update/renku-graph-2.27.0

RenkuBot:
This is an automated pull request.

/deploy #persist

PR #2924, opened 2023-01-26
chore: bump template version

rokroskar:
Bumps the template version.

/deploy #persist

### DIFF
--- a/acceptance-tests/Dockerfile
+++ b/acceptance-tests/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get update && \
     wget \
     curl
 
-# This 'mkdir' is just to make the openjdk-11-jdk installation to pass
+# This 'mkdir' is just to make the openjdk installation to pass
 RUN mkdir -p /usr/share/man/man1 && \
     apt-get update && \
     apt-get install -y openjdk-17-jdk
@@ -36,7 +36,7 @@ USER attests
 
 ENV PATH=/home/attests/.local/bin:/home/attests/.local/lib/python3.8/site-packages:$PATH
 
-RUN python3 -m pip install --upgrade 'pip==22.3.1' && \
+RUN python3 -m pip install --upgrade 'pip==23.0' && \
     python3 -m pip install -r /tests/requirements.txt
 
 ENV DOCKER="1"
--- a/acceptance-tests/requirements.txt
+++ b/acceptance-tests/requirements.txt
@@ -1,4 +1,4 @@
 papermill==2.4.0
-toil==5.7.1
-ipykernel==6.20.2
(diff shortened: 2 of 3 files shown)


lorenzo-cavazzi (approved):
Lgtm :rocket: 
I tested the template with the changes coming here https://github.com/SwissDataScienceCenter/renku/pull/2914

I suggest having a sample workflow file that works out-of-the-box when invoking it through `renku run`, but this should not a blocker for this PR (since it includes other necessary fixes)

rokroskar:
It's a good point - @lokijuhy @mohammad-alisafaee any reason why you didn't make the templated workflow file something that can run immediately?

lokijuhy:
> It's a good point - @lokijuhy @mohammad-alisafaee any reason why you didn't make the templated workflow file something that can run immediately?

I'm confused - What would the workflow run? There are no code or data files in an empty project template

rokroskar:
It can run any shell command e.g. `head -n 10 README.md` - I agree it's contrived, but my guess is that many users will try to run it and be annoyed it doesn't do anything

lorenzo-cavazzi:
It actually prints an error, and that's a bit confusing. A non-sense workflow that doesn't fail would be better.

image

rokroskar:
ok thanks - lets fix the template and retag, it's not "out" yet anyway

lokijuhy:
I see it quite differently. I think a user who is totally new to workflows would be confused coming into a workflow template file and seeing a bunch of bash commands that they don't recognize. I think that would hamper their understanding of how to use the workflow file. I think it's more relatable to see how the workflow file is intended to be used with python and R scripts- the materials the user is accustomed to.

On the other hand, I am having trouble imagining what a user would expect a workflow template file would do when run un-edited...?

lokijuhy:
> It actually prints an error, and that's a bit confusing. A non-sense workflow that doesn't fail would be better.
> 
> image

That doesn't look like a workflow-specific error- that looks like the same cwl error that's affecting all renku commands.

But yes, (once that general renku error is fixed) running the workflow template without editing it will fail. I could see perhaps adding a step to the workflow template that runs `cat "I'm a template- edit me to use me!"` ?

rokroskar:
it's a yaml indentation error, should be 

```
diff --git a/workflows/my-workflow.yaml b/workflows/my-workflow.yaml
index 1e89851..9d154f3 100644
--- a/workflows/my-workflow.yaml
+++ b/workflows/my-workflow.yaml
@@ -25,8 +25,8 @@ steps:
           path: data/output/results.csv
     parameters:
       - n:
-        prefix: -n
-        value: 10
+          prefix: -n
+          value: 10
 ```

PR #2929, opened 2023-02-02
chore: decrease redis sentinel getMasterTimeout

rokroskar:
Lowers the `redis.sentinel.getMasterTimeout` to avoid having redis start very slowly. 

/deploy


helm-chart/renku/values.yaml
```diff
@@ -374,6 +374,7 @@ redis:
       appendonly no
       # Disable RDB persistence.
       save ""
+
```
aledegano, line 377: not that is super important, but was this new line intended?

rokroskar: no...

rokroskar: haha now I'm confused... I removed it but somehow it's still there 🤷

aledegano: there's also 2 white spaces?...

rokroskar: 🤦

seanrmurphy (approved):

Panaetius (approved):

PR #2992, opened 2023-03-08
test(cypress): verify infrastructure availability before testing

lorenzo-cavazzi:
This PR adds a test to verify that all the backend services are available and working properly _before_ starting the browser-based tests.
It also removes to (ineffective) 1-minute waiting time.

### DIFF
--- a/cypress-tests/cypress.config.ts
+++ b/cypress-tests/cypress.config.ts
@@ -5,6 +5,7 @@ export default defineConfig({
   e2e: {
     baseUrl: process.env.BASE_URL || "https://dev.renku.ch",
     specPattern: [
+      // Verify infrastructure before running any other tests
       "cypress/e2e/verifyInfrastructure.cy.ts",
       "cypress/e2e/*.{js,jsx,ts,tsx}"
     ],


lorenzo-cavazzi:
The issue reference is wrong, but I'm leaving it there to prevent interrupting and re-triggering deployments and tests :scream:

cypress-tests/cypress.config.ts
```diff
@@ -3,7 +3,11 @@ import { TIMEOUTS } from "./config";
 
 export default defineConfig({
   e2e: {
-    baseUrl: process.env.BASE_URL || "https://dev.renku.ch"
+    baseUrl: process.env.BASE_URL || "https://dev.renku.ch",
+    specPattern: [
+      "cypress/e2e/verifyInfrastructure.cy.ts",
```
ciyer, line 8: I think a comment here would be helpful to ensure that no one accidentally reorders this.

```suggestion
      // Verify infrastructure before running any other tests
      "cypress/e2e/verifyInfrastructure.cy.ts",
```

ciyer (approved):
This looks good! 🚀 

I added a comment suggestion for clarification, but I think it's optional. If you agree that it would be helpful, you can merge the change and I will re-approve. If you don't think it is necessary go ahead and merge.

cypress-tests/cypress.config.ts
lorenzo-cavazzi: Good idea! :+1:

ciyer (approved):
👍

lorenzo-cavazzi:
I accidentally merged #2993 before this one. Since the additional commit is only adding one comment, I added it to #2994 and will close this PR -- sorry for the snafu!

Issue #2852, opened 2022-12-13
Cypress tests for Workflows


ciyer:
- [x] Basic workflow test
- [x] Prepare a project with complex workflows and add to the `renku-ui-test` group (depends on #2853)
- [x] Advanced workflow test

## Basic workflow test

- start a session
- connect
- run very simple renku commands to create a workflow
- check it in the workflows tab

## Advanced workflow test

- start a session
- re-run a workflow
- verify the workflows tab show the change
- go to the file view, confirm there is new content

lorenzo-cavazzi:
Proposal:
- Basic workflow: as suggested, but add it to the current `useSession` spec since most of the time would be spent in starting the session. Furthermore, the current `useSession` lacks typing code in the console, which would be a great addition.
- Advanced workflow: since we cannot run workflows directly from the UI, for now we could just create a project with complex workflows and verify we can see them in the UI. We can expand this new `checkWorkflows` spec in the future once we could interact with workflows. This should be enough to catch breaking changes from the renku-python side

PR #2993, opened 2023-03-08
test(cypress): add session-specific test

lorenzo-cavazzi:
This PR extends the Cypress acceptance tests with a new test to verify that starting and terminating a session goes through the expected steps. There is also minimal interaction inside a JupyterLab environment.
This addresses the first part of #2852 but it doesn't tests workflows yet -- to be expanded.

re #2852


ciyer (approved):
Very nice!

PR #2994, opened 2023-03-08
test(cypress): use v9 and v10 projects for testing updates

lorenzo-cavazzi:
This PR refreshes the "Update Projects" tests by using v9 projects and trying an actual update.


ciyer (approved):
LGTM!

PR #3007, opened 2023-03-20
chore: add load tests helm chart

olevski:
This combines pre-existing load tests that existed mostly in the renku-python repo, puts them all in one place (here in this repo), updates the tests to use the chai-like describe/expect syntax and adds a helm chart for the load tests.

In addition to tests can be setup to publish their metrics to prometheus.

See existing metrics at https://grafana.dev.renku.ch. All metrics from k6 start with the name `k6`.

The idea is that the cronjob would run nightly on the dev cluster.

Once this is merged and the helm chart is published I will make a PR in the dev ops repo to deploy this. The existing results are only from running the tests a few times.

### DIFF
--- a/helm-chart/load-tests/templates/cronjob.yaml
+++ b/helm-chart/load-tests/templates/cronjob.yaml
@@ -1,0 +1,68 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "load-tests.fullname" . }}
+  labels:
+    {{- include "load-tests.labels" . | nindent 4 }}
+spec:
+  schedule: {{ .Values.schedule }}
+  concurrencyPolicy: Replace
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          {{- with .Values.podAnnotations }}
+          annotations:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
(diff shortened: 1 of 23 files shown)


olevski:
Example plot of response times:
image

helm-chart/load-tests/templates/cronjob.yaml
```diff
@@ -0,0 +1,68 @@
...
+          containers:
+            - name: {{ .Chart.Name }}
+              securityContext:
+                {{- toYaml .Values.securityContext | nindent 16 }}
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
+              resources:
+                {{- toYaml .Values.resources | nindent 16 }}
+              env:
+                {{- if gt (len .Values.extraEnvVars) 0 }}
```
Panaetius, line 38: Shouldn't the config like `BASE_URL` be here and in the values.yaml?

Panaetius: Maybe add them commented out in the values file as well? just so it's clear what values could be set

Panaetius (approved):

PR #3016, opened 2023-03-27
chore: upgrade Keycloak

aledegano:
/deploy #persist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,17 @@
 .. _changelog:
 
+0.27.0
+------
+
+Renku ``0.27.0`` upgrades the Keycloak version that is shipped with the project.
+
+Internal Changes
+~~~~~~~~~~~~~~~~
+
+**Improvements**
+
+- **Infrastructure**: (BREAKING CHANGE) Use a new Keycloak Helm chart (codecentric/keycloakx) and upgrade Keycloak to 20.0.3. This requires modifying your current values file to work with the new Keycloak Helm chart, see (`the helm chart values changelog <https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md>`_) for instructions.
+
 0.26.2
 ------
 
--- a/helm-chart/renku/Chart.yaml
+++ b/helm-chart/renku/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: '1.0'
+appVersion: "1.0"
 description: A Helm chart for the Renku platform
 name: renku
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.18.1-2.01781e6
+version: 0.0.1-set.by.chartpress
--- a/helm-chart/renku/templates/_helpers.tpl
+++ b/helm-chart/renku/templates/_helpers.tpl
@@ -54,7 +54,7 @@ Define subcharts full names
 {{- end -}}
 
 {{- define "keycloak.fullname" -}}
-{{- printf "%s-%s" .Release.Name "keycloak" | replace "+" "_" | trunc 20 | trimSuffix "-" -}}
+{{- printf "%s-%s" .Release.Name "keycloakx" | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{- define "gitlab.fullname" -}}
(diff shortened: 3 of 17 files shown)


seanrmurphy (approved):

helm-chart/renku/values.yaml
```diff
@@ -242,15 +255,15 @@ keycloak:
   initRealm:
     image:
       repository: renku/init-realm
-      tag: "latest"
+      tag: '0.23.1-5.5b17747.n045.h3cb67ea5'
```
olevski, line 258: Reset the tags to what they were before. Chartpress will set them to the right thing when we publish the helm chart.

helm-chart/renku/values.yaml
```diff
@@ -794,7 +807,7 @@ tests:
   enabled: false
   image:
     repository: renku/tests
-    tag: "latest"
+    tag: '0.23.1-5.5b17747.n045.h3cb67ea5'
```
olevski, line 810:
```suggestion
    tag: "latest"
```

helm-chart/renku/values.yaml
```diff
@@ -885,4 +898,4 @@ swagger:
 initDb:
   image:
     repository: renku/init-db
-    tag: "latest"
+    tag: '0.23.1-5.5b17747.n045.h3cb67ea5'
```
olevski, line 901:
```suggestion
    tag: "latest"
```

helm-chart/renku/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku platform
 name: renku
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.18.1-2.01781e6
+version: 0.23.1-5.5b17747.n048.h90d0df90
```
olevski, line 6: I thought we never usually set this to anything in particular and let chartpress set it when we publish or install. For the longest time it used to be set to `0.0.1-set.by.chartpress`. I think we should revert to something like that or to `latest` or to the latest published version.

helm-chart/renku/templates/_helpers.tpl
```diff
@@ -103,3 +103,47 @@ fail "External PostgreSQL and Renku-bundled PostgreSQL cannot both be disabled.
 {{- if .Values.global.externalServices.postgresql.enabled and .Values.global.externalServices.postgresql.password and .Values.global.externalServices.postgresql.existingSecret -}}
 fail "External PostgreSQL password and existing Secret fields cannot both be populated."
 {{- end -}}
+
+{{- define "keycloak.admin-secret" -}}
+{{- $secretAdmin := lookup "v1" "Secret" .Release.Namespace "keycloak-password-secret" -}}
+{{- if $secretAdmin -}}
+{{- if $secretAdmin.data.KEYCLOAK_ADMIN -}}
```
olevski, line 110: I think this will crash if there is no `KEYCLOAK_ADMIN` in the secret? I think we should use `get` or something similar to aboid an error if the field is not set. I assume the purpose of the if statement is to avoid crashing if the field is not present in the secret.

helm-chart/renku/templates/_helpers.tpl
```diff
@@ -103,3 +103,47 @@ fail "External PostgreSQL and Renku-bundled PostgreSQL cannot both be disabled.
...
+{{- else -}}
+KEYCLOAK_ADMIN: {{ .Values.global.keycloak.user | b64enc }}
+KEYCLOAK_ADMIN_PASSWORD: {{ default (randAlphaNum 64) .Values.global.keycloak.password.value | b64enc }}
+{{- end -}}
+{{- end -}}
+
+{{- define "keycloak.postgres-secret" -}}
+{{- $secretPostgres := lookup "v1" "Secret" .Release.Namespace "renku-keycloak-postgres" -}}
+{{- if $secretPostgres -}}
+{{- if $secretPostgres.data.KC_DB_URL_HOST -}}
```
olevski, line 129: Same as the other helper section. I suspect this may crash the helm chart if the key is not present and the else statement will never be reached.

helm-chart/renku/values.yaml
```diff
@@ -171,21 +171,34 @@ ingress:
   #      - example.local
 
 ## Keycloak configuration
-keycloak:
+keycloakx:
```
olevski, line 174: There are already a few checks in `_helpers.tpl` that fail on specific conditions with an explanation why. We should add a similar statement that will fail the helm upgrade/install if we detect that the values file does not have a `keycloakx` field but has a `keycloak` field. I feel like people may not read the release notes and not upgrade the values file but simply try to install the helm chart and then wonder why everything is broken. What do you think?

olevski (changes requested):

olevski:
@aledegano one more thing. I do not see any documentation in the `CHANGELOG.rst` or even more appropriately to `helm-chart/README.rst` where we discuss the steps to upgrade/migrate to different Renku helm chart versions. IMO we should add this as part of this PR.

aledegano:
Do you mean steps to upgrade to this version?
Because the migration is fully automated.

olevski:
> Do you mean steps to upgrade to this version? Because the migration is fully automated.

So if I have a values file with a custom `keycloak` section and I do not rename it to `keycloakx` everything will work? And I will also still get the new Keycloak helm chart? I think in this case where someone has a non-default `keycloak` section and if they do not rename this to keycloakx they will still get the new keycloak but without their custom settings. So we need to make it clear that the `keycloak` section should be updated to `keycloakx` right?

Also regardless of how simple or complicated the migration is we need to add a new section with new version and release notes for this change in `CHANGELOG.rst`. This is missing.

Lastly since we change the values file it would be nice to mention this in `helm-chart/values.yaml.changelog.md`. I think we kind of have things about helm migrations scattered between `helm-chart/README.rst` and `helm-chart/values.yaml.changelog.md`. But `helm-chart/values.yaml.changelog.md` is more complete and up-to-date. So even if the migration is fully automated mentioned this in the values.yaml.changelog.md is a good idea IMO.

helm-chart/renku/templates/_helpers.tpl
aledegano: No, it doesn't crash, it return `false`. The purpose of the `if` statement on line 110 is to decide whether the secret exist with the var name expected from Keycloakx `KEYCLOAK_ADMIN`, which means that the migration already happened and the secret just needs to remain the same or (this is the else) if the secret has the var name used with Keycloak `KEYCLOAK_USER`.

If the secret does not exist at all (outer `if` nest) then the secrets are generated.

aledegano:
@olevski I have addressed all your comments in the latest commit. When you have time could you check again please?

CHANGELOG.rst
```diff
@@ -1,5 +1,17 @@
...
+------
+
+Renku ``0.27.0`` upgrades the Keycloak version that is shipped with the project.
+
+Internal Changes
+~~~~~~~~~~~~~~~~
+
+**Improvements**
+
+- **YAT**: Upgrade Keycloak to 20.0.3, by using the chart codecentric/keycloakx
```
olevski, line 13:
```suggestion
- **YAT**: (BREAKING CHANGE) Use a new Keycloak Helm chart (codecentric/keycloakx) and upgrade Keycloak to 20.0.3. This requires modifying your current values file to work with the new Keycloak Helm chart, see (`the helm chart values changelog <https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md>`_) for instructions.
```

helm-chart/renku/values.yaml
```diff
@@ -599,7 +609,7 @@ ui:
           ref: 0.6.1
           name: Renku
         - url: https://github.com/SwissDataScienceCenter/contributed-project-templates
-          ref: 0.5.0
+          ref: 0.4.1
```
olevski, line 612:
```suggestion
          ref: 0.5.0
```
This should not change at all. Unless there is something specific form this PR that requires for the community project templates to change. I suspect this is the result of resolving conflicts in the PR.

helm-chart/renku/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: '1.0'
+appVersion: "1.0"
 description: A Helm chart for the Renku platform
 name: renku
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.18.1-2.01781e6
+version: 0.25.6
```
olevski, line 6:
```suggestion
version: 0.0.1-set.by.chartpress
```

If we set this to a different specific value then we have to keep updating it on every commit. It is easier to have it like this. We used to have it set to this a few months ago. I suspect a specific version got accidentally committed and merged.

olevski (changes requested):
Just a few minor edits and further reverts of version that have been accidentally committed from previous PRs.

Also please resolve the conflicts.

olevski (dismissed):

wesjdj (approved):

olevski (dismissed):

CHANGELOG.rst
```diff
@@ -1,5 +1,17 @@
...
+------
+
+Renku ``0.27.0`` upgrades the Keycloak version that is shipped with the project.
+
+Internal Changes
+~~~~~~~~~~~~~~~~
+
+**Improvements**
+
+- **YAT**: (BREAKING CHANGE) Use a new Keycloak Helm chart (codecentric/keycloakx) and upgrade Keycloak to 20.0.3. This requires modifying your current values file to work with the new Keycloak Helm chart, see (`the helm chart values changelog <https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md>`_) for instructions.
```
rokroskar, line 13: Don't use "YAT" in public changelogs, it's an internal name

rokroskar (changes requested):

CHANGELOG.rst
bethcg: We usually use  Renku Core Service

rokroskar: In this case it should just be something like "supporting services" or "3rd party services" or "Infrastructure" - "core service" is actually a service we write

aledegano: done

bethcg (approved):

rokroskar (approved):

olevski (approved):

olevski:
@aledegano the acceptance tests here will never pass because the values file contains the old keycloak section. And we made the helm chart fail if this is the case. We currently have no way to remove a section from the values file for a CI deployment. I know you tested this in a separate deployment. So it is probably easiest to merge this and then change the secret that contains the values files for all CI deployments.

aledegano:
> @aledegano the acceptance tests here will never pass because the values file contains the old keycloak section. And we made the helm chart fail if this is the case. We currently have no way to remove a section from the values file for a CI deployment. I know you tested this in a separate deployment. So it is probably easiest to merge this and then change the secret that contains the values files for all CI deployments.
> 
> Alternatively we can remove the check that fails the helm chart deployment on purpose and add it in a different PR.

I've already changed the values file for the CI deployments. The deployment of this PR is already passed, now the tests are running.

PR #3027, opened 2023-04-06
chore: docs for renaming templates

mohammad-alisafaee:
Documents for renaming templates in https://github.com/SwissDataScienceCenter/renku-python/pull/3347.

### DIFF
--- a/docs/reference/templates.rst
+++ b/docs/reference/templates.rst
@@ -329,8 +329,10 @@ contains all the specifications needed by the ``renku init`` function to
 create a new project. You can specify multiple templates in the same
 repository. Each of them requires an entry with the following parameters:
 
-* ``id`` (instead of ``folder`` which is deprecated): the target directory inside the repository
-  where the template files are stored. You must use a different id/directory for each template.
+* ``id``: the target directory inside the repository where the template files are stored.
+  You must use a different id/directory for each template. Note that this field was called
+  ``folder`` previously which is deprecated now. You can still use ``folder`` but
+  when validating your templates, you'll get a warning to replace it with ``id``.
 * ``name``: a short user-friendly name.
 * ``description``: a brief description of your template. This will be
   presented to the user when choosing between templates.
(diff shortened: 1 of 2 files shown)


docs/reference/templates.rst
```diff
@@ -471,6 +475,38 @@ include your repository in the RenkuLab template source through the
...
 
 
+.. _rename-templates:
+
+Renaming templates
+^^^^^^^^^^^^^^^^^^
+
+If you are the owner of a template repository, you can change the name of the
+templates. To do this, edit the ``manifest.yaml`` file and put the new name
+in the ``id`` (``folder``) field of the template you want to rename. You then
```
olevski, line 485: I think we should mention that `folder` is deprecated and should not be used here as well. Because this to me looks like you can use either `folder` or `id`.

docs/reference/templates.rst
```diff
@@ -379,13 +382,14 @@ repository. Each of them requires an entry with the following parameters:
...
+* ``allow_template_update``: This field is deprecated and will be ignored.
+  It was used to indicates that the template supports being updated. All
+  templates are updatable now. When the template gets updated, projects
   created from it will get updated with the new template files. Defaults to
   ``false``. Also see ``immutable_template_files``.
 * ``immutable_template_files``: A list of file paths inside the template
-  (relative to the project root) that should not be changed by users for
-  ``allow_template_update`` to work. Users changing any of these files will
+  (relative to the project root) that should not be changed by users.
+  Users changing any of these files will
```
olevski, line 392: I think this warning happens only if you have the cli's commit hooks installed right? Perhaps we should clarify this.

docs/reference/templates.rst
```diff
@@ -329,8 +329,8 @@ contains all the specifications needed by the ``renku init`` function to
 create a new project. You can specify multiple templates in the same
 repository. Each of them requires an entry with the following parameters:
 
-* ``folder``: the target folder inside the repository where the template files
-  are stored. Please use a different folder for each template.
+* ``id`` (instead of ``folder`` which is deprecated): the target directory inside the repository
```
olevski, line 332: What happens to template repositories in the "wild" where they still use `folder`? Is the grace period for switching already done and from now on you have to use `id` and only using `folder` will fail?

olevski (changes requested):
I think we just need to clarify a few things about the deprecation. Otherwise all looks good to me.

docs/reference/templates.rst
mohammad-alisafaee: No. `folder` will be supported forever. We only print a warning to replace it with `id` when validating a template using `renku template validate` command.

olevski (approved):
Thank you!

PR #3032, opened 2023-04-12
chore: update changelog for 0.25.0

rokroskar:
Adds a blurb about why users should upgrade their base images.


elisabet-cg (approved):
Clear :)

bethcg (approved):

PR #3053, opened 2023-05-02
fix(cypress): close all existing running sessions in useSession

### LINKED ISSUE #3009: Make cypress acceptance tests more robust to previous failures
Sometimes a cypress acceptance test runs and leaves a zombie session around that causes problems for later runs.

We should make the logic that connects to a session also consider the project (which should be unique per run), since this will make the test more robust.

leafty:
Close all running sessions in `useSession.cy.ts` before running the tests. This should fix zombie sessions making tests fail.

Fixes #3009.

/deploy #persist #cypress


lorenzo-cavazzi (approved):
Lgtm!

I pushed a commit for linting the code -- not 100% necessary tbh, but it prevents some warnings with basic linting rules turned on.
I also swapped the `intercept` and `visit` even tho in practice visiting the page won't be that quick and the API should be intercepted anyway

leafty:
> Lgtm!
> 
> I pushed a commit for linting the code -- not 100% necessary tbh, but it prevents some warnings with basic linting rules turned on. I also swapped the `intercept` and `visit` even tho in practice visiting the page won't be that quick and the API should be intercepted anyway

Great, thanks! :pray:

PR #3056, opened 2023-05-03
feat: acceptance test not to remove project on failure

jachro:
It was found out that it's pretty inconvenient if a project created as part of a test execution is removed in case of a failure. In some situations, it's very difficult to reason about the failure as all events that may contain information about provisioning errors are removed.

This PR introduces a change that preserves the project created as part of a failing test scenario.

/deploy #persist


eikek (approved):
🙏🏼

PR #3067, opened 2023-05-10
feat: add crac service helm chart

olevski:
/deploy #persist extra-values=crac.image.tag=pr-7

depends on: https://github.com/SwissDataScienceCenter/renku-data-services/pull/7


helm-chart/renku/templates/configmap.yaml
```diff
@@ -146,3 +148,22 @@ data:
...
+
+  # DO NOT USE set -x you will leak the password in the logs
+  init-renku-db.sh: |-
+    #!/bin/bash
+
+    PASSWORD=$(cat /db-common-password/password)
+    DB_NAME=renku
+
+    psql -v ON_ERROR_STOP=1 <<-EOSQL
+      create database "$DB_NAME";
```
seanrmurphy, line 160: We think we've been doing this wrong; we think that we should be creating the db such that the db owner is the new user; hence the user should be created first and the db should be created with this user as owner.

olevski: I changed this as you requested. I looked at the other PR where you fixed this for the kg.

CHANGELOG.rst
```diff
@@ -1,5 +1,40 @@
 .. _changelog:
 
+0.31.0
+------
+
+Renku ``0.31.0`` introduces the compute resource control (CRC) service and supporting or related changes in
```
rokroskar, line 6: Can we have a clear statement here about what this is for and what it enables?

bethcg: I have just proposed one.

CHANGELOG.rst
```diff
@@ -1,5 +1,39 @@
 .. _changelog:
 
+0.31.0
+------
+
+Renku ``0.31.0`` introduces the compute resource control (CRC) service. This feature enables Renku administrators to manage the users' access to specific computing resources. The service exposes an HTTP API for the administrators to interact with. In this way, Renku administrators can create, update or delete resource pools, and can add or remove users to resource pools. Thus, this release includes new features in Renku individual components related to the compute resource service, and supports common R file extensions.
```
rokroskar, line 6: This should also say something about what it means for users, i.e that they will from now on need to select a resource class not individually select compute requirements.

rokroskar, line 6:
```suggestion
Renku ``0.31.0`` introduces the compute resource control (CRC) service, enabling Renku administrators to manage access to specific computing resources. The service exposes an HTTP API for the administrators to interact with. In this way, Renku administrators can create, update or delete resource pools, and can add to or remove users from resource pools. 

Apart from the changes needed to support compute resource access features, support has also been added for common R file extensions.
```

lorenzo-cavazzi (approved):
I manually tested some scenarios and this works well. Very nice job! :rocket:

CHANGELOG.rst
olevski: I expanded the description to mention this.

CHANGELOG.rst
```diff
@@ -1,5 +1,57 @@
 .. _changelog:
 
+0.31.0
+------
+
+Renku ``0.31.0`` introduces the compute resource control (CRC) service, enabling Renku administrators to manage access to 
+specific computing resources. The service exposes an HTTP API for the administrators to interact with. In this way, 
+Renku administrators can create, update or delete resource pools, and can add to or remove users from resource pools. 
+Please note that a user interface for the CRC has not been added yet but it will be added in a future release. Currently,
```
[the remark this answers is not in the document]
olevski: I think that administrators especially the ones deploying or updating to a new version of Renku are most likely to read the changelog in the repo or the docs. At least from my experience helping admins deploy Renku I have heard complaints specifically because the release notes excluded information or were too brief. Currently we have no other place to include this information as far as I know. So I think we should definitely include this here.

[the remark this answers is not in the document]
olevski: Alternatively we can make another changelog of section of the docs for this information specifically tailored for admins. But I am not sure it is worth the effort and it will only scatter information that I think belongs together.

lorenzo-cavazzi (approved):

bethcg (approved):
Good for me!

aledegano (approved):

PR #3068, opened 2023-05-11
feat: tests to remove test projects older than 7 days

jachro:
This PR modifies the existing `BatchProjectRemovalSpec` so it removes all the projects falling into the `test-(\d{4})-(\d+)-(\d+)-(\d+)-(\d+)-(\d+).*` pattern and are older than 7 days. To some extent, it's a complementary change to the one merged recently which started to skip project removal on test failure.

/deploy #persist

### DIFF
--- a/acceptance-tests/.scalafmt.conf
+++ b/acceptance-tests/.scalafmt.conf
@@ -19,7 +19,7 @@ align.openParenCallSite = true
 align.openParenDefnSite = true
 
 rewrite {
-  rules = [RedundantBraces, RedundantParens, AsciiSortImports, SortModifiers, PreferCurlyFors]
+  rules = [RedundantBraces, RedundantParens, SortModifiers, PreferCurlyFors]
   sortModifiers.order = [
     "private", "protected", "final", "implicit", "sealed", "abstract", "override", "lazy"
   ]
--- a/acceptance-tests/src/test/scala/ch/renku/acceptancetests/BatchProjectRemovalSpec.scala
+++ b/acceptance-tests/src/test/scala/ch/renku/acceptancetests/BatchProjectRemovalSpec.scala
@@ -22,8 +22,9 @@ import ch.renku.acceptancetests.tooling.{AcceptanceSpec, KnowledgeGraphApi}
 import ch.renku.acceptancetests.workflows._
 
 import java.lang.System.getProperty
-import java.time.{Duration => JDuration, Instant}
+import java.time.{Instant, Duration => JDuration}
 import Ordering.Implicits._
+import scala.util.matching.Regex
 
 /** Delete old projects.
   *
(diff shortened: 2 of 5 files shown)


eikek (dismissed):
👍🏼

Panaetius:
Should we also remove projects like `test-session-933d2928fc69` ?

jachro:
> Should we also remove projects like `test-session-933d2928fc69` ?

Sure. I hesitated a bit as they are created by some other tests but why not.

Let me add them.

acceptance-tests/.scalafmt.conf
```diff
@@ -19,7 +19,7 @@ align.openParenCallSite = true
 align.openParenDefnSite = true
 
 rewrite {
-  rules = [RedundantBraces, RedundantParens, AsciiSortImports, SortModifiers, PreferCurlyFors]
+  rules = [RedundantBraces, RedundantParens, SortModifiers, PreferCurlyFors]
```
eikek, line 22: There is a scalafix semantic rule [organize imports](https://github.com/liancheng/scalafix-organize-imports) that I found quite nice to use. Perhaps something to look at/discuss for later.

acceptance-tests/src/test/scala/ch/renku/acceptancetests/BatchProjectRemovalSpec.scala
```diff
@@ -18,108 +18,74 @@
...
+        .getOrElse(true)
+
+    val defaultPatterns = List(
+      "test-(\\d{4})-(\\d+)-(\\d+)-(\\d+)-(\\d+)-(\\d+).*",
+      "test-session-.*"
+    )
+
+    val projectNamePatterns = Option(getProperty("remPattern"))
+      .orElse(sys.env.get("RENKU_TEST_REMOVE_PATTERN"))
+      .map(List(_))
```
eikek, line 86: When using properties or env variables, we can't specify multiple patterns, right?

eikek (approved):

acceptance-tests/src/test/scala/ch/renku/acceptancetests/BatchProjectRemovalSpec.scala
jachro: To be honest I think it's a bit dead feature. I believe it was created with an idea in mind that from time to time we run the tests manually in the mode allowing the removal of the projects. In fact, it's barely ever used and when it will be, the feature should work as before.

acceptance-tests/.scalafmt.conf
jachro: Yes, sure. I'm more than happy to talk about that. I simply removed it as I quite often organize imports with Intellij and I couldn't force it to use Ascii Sorting. There are few other options to choose from so maybe we can find a common denominator to satisfy both.

acceptance-tests/src/test/scala/ch/renku/acceptancetests/BatchProjectRemovalSpec.scala
jachro: Thanks Sekhar.
Yes, I thought it was you who created (and run) the test :-) As I mentioned in the PR desc, we don't delete projects created by a test that fails during execution so we thought a more systematic clean-up process was a good thing to do.

PR #3109, opened 2023-06-07
fix(gitlab): Set the authorized params GL accept from OAuth to fix upgrade to KeycloakX.

aledegano:


helm-chart/README.rst
```diff
@@ -62,6 +62,28 @@ Most information related to upgrading from one chart version to another is cover
...
+    'args' => {
+        [...]
+          authorize_params: {
+            scope: "openid profile email"
+          }
+        },
+    [...]
+      }
+    ]
+```
```
rokroskar, line 85: isn't this done automatically in `_gitlab.rb.tpl`?

rokroskar: ah nevermind, I missed the "independently" part 😆

helm-chart/README.rst
```diff
@@ -62,6 +62,28 @@ Most information related to upgrading from one chart version to another is cover
 in the `values changelog file <https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/values.yaml.changelog.md>`_.
 For upgrades that require some steps other than modifying the values files to be executed, we add some instructions here.
 
+Upgrading to 0.27.0
```
olevski, line 65: Can mention in the release notes (i.e. CHANGELOG.rst) that people should look at both the `values.yaml.changelog.md` and `helm-chart/README.rst` for instructions on how to update.

Currently in `CHANGELOG.rst` we only tell users to look at `values.yaml.changelog.md` but not the helm chart readme. But both are relevant and should be checked.

olevski (changes requested):

olevski (approved):

bethcg (approved):
Proceed 👍

PR #3137, opened 2023-07-04
fix(chart): generate db secrets on upgrade

olevski:
I am not sure what is the logic behind the helm hooks for the secret creation.

But IMO it was needlessly complicated and it did not work when you upgrade from the helm chart version where the secret wasn't there.

So the changes I made are the following:
- removed the deletion hook because it was setting a value that is already the default
- made the hooks run before installing, upgrading or rolling back

This way the secret can be created when you are upgrading. Which was not the case before. 

I tested by deploying an older version and then upgrading. With this fix the secret is properly randomly created when you upgrade renku.

Because we have the logic to check if the secret exists (and not update it if it does) we can run the hook every time.

We have had similar discussions about this before. This does not prevent the admin from setting the value for a password in the helm chart values and breaking the deployment. Because the postgres secret is not updated when the helm chart values change. But this is acceptable IMO. I also added warnings about this in the values for the helm chart.

/deploy #persist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,14 @@
 
 Renku ``0.29.0`` introduces a PostgreSQL DB for triples-generator.
 
+Internal Changes
+~~~~~~~~~~~~~~~~
+
+**Bug Fixes**
+
+- **Infrastructure**: properly generate postgres secrets on upgrade (`#3137 <https://github.com/SwissDataScienceCenter/renku/issues/3137>`_).
+
+
 
 0.28.1
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ Internal Changes
 
 **Bug Fixes**
 
-- **Infrastructure**: properly generate postgres secrets on upgrade (`#3137 <https://github.com/SwissDataScienceCenter/renku/issues/3137>`_).
+- **Infrastructure**: properly generate PostgreSQL secrets on upgrade (`#3137 <https://github.com/SwissDataScienceCenter/renku/issues/3137>`_).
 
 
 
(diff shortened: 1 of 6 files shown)


wesjdj:
If an external Postgres is used with credentials automatically generated, the problem was that when the chart is reinstalled, the secret containing the credentials was being overwritten with newly generated values which were incorrect. This meant that the correct credentials were then lost and had to be changed on the database manually to reflect the new credentials, which had overwritten the previous credentials.

lorenzo-cavazzi (dismissed):

olevski:
> If an external Postgres is used with credentials automatically generated, the problem was that when the chart is reinstalled, the secret containing the credentials was being overwritten with newly generated values which were incorrect. This meant that the correct credentials were then lost and had to be changed on the database manually to reflect the new credentials, which had overwritten the previous credentials.

I discussed this with Wes over slack and tested this scenario. It does not occur. I think we had this problem a while ago when we did not have this logic in each secret:

```
{{- if .Values.gitlab.enabled }}
{{- $db_password := default (randAlphaNum 64) .Values.global.gitlab.postgresPassword.value | b64enc | quote }}
{{- $renkuFullname := include "renku.fullname" . -}}
{{- if not .Values.global.gitlab.postgresPassword.value -}}
{{- $secretName := cat $renkuFullname "-gitlab-postgres" | nospace }}
{{- $secret := (lookup "v1" "Secret" .Release.Namespace $secretName) }}
{{- if $secret }}
{{- $db_password = index $secret.data "gitlab-postgres-password" }}
{{- end -}}
{{- end -}}
```

But with the addition of the above the secret is not overwritten by the automated secret generation if the secret already exists.

The secret is overwritten though if the admin sets the value in the helm chart but they had previously had the helm chart generate it for them. And this will lead to problems in postgres. But that is why I added a warning to the values file near all the passwords that are affected by this fix.

lorenzo-cavazzi (approved):

PR #3138, opened 2023-07-04
chore: Create triples-generator db owned by the service user

eikek:
The database is now created with the service user as the owner.
/deploy


seanrmurphy:
This is how the db was created - we can see that owner is triplesgenerator and triplesgenerator and postgres users have access....

image

seanrmurphy (approved):
LGTM

PR #3148, opened 2023-07-07
Improve realm init

sgaist:
This merge request does several small improvements:

- Fixes a warning that will turn into an error when python-keycloak 4 will be out
- Improves the warning message when something goes wrong adding the status code so one can know what might be happening
- Bump the python-keycloak version to the latest release at the time of writing (3.3.0)

The last point goes beyond what is proposed in #3066 and this pull request can be used in place of it.

In any case, the python-keycloak bump is required as the current release of `init-realm` (0.29.0) fails to generate the realms and clients due to an internal error in the python-keycloak version used.

### DIFF
--- a/scripts/init-realm/init-realm.py
+++ b/scripts/init-realm/init-realm.py
@@ -224,7 +224,7 @@ success = False
 
 while not success and n_attempts < 31:
     try:
-        sys.stdout.write("Getting an admin access token for Keycloak...")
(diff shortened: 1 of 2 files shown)


olevski (approved):
We already have a PR open with a subset of these changes. But external contributors are always welcome! So it would be nice to merge this.

olevski:
@sgaist please rebase your branch and I will re-approve.

olevski (approved):

PR #3150, opened 2023-07-07
Separate and parallelize Cypress test files

lorenzo-cavazzi:
Separate and parallelize Cypress test files

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -42,8 +42,9 @@ jobs:
         with:
           name: docs
           path: docs/_build/html/
+
   check-deploy:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     outputs:
       pr-contains-string: ${{ steps.deploy-comment.outputs.pr-contains-string }}
       renku-core: ${{ steps.deploy-comment.outputs.renku-core}}
@@ -62,10 +63,11 @@ jobs:
         with:
           string: /deploy
           pr_ref: ${{ github.event.number }}
+
   deploy-pr:
     if: github.event.action != 'closed'
     needs: [check-deploy]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     environment:
       name: ci-renku-${{ github.event.number }}
     steps:
@@ -109,11 +111,11 @@ jobs:
             You can access the deployment of this PR at https://ci-renku-${{ github.event.number }}.dev.renku.ch
 
   test-pr:
-    runs-on: ubuntu-20.04
     if: ${{ github.event.action != 'closed' &&
       needs.check-deploy.outputs.pr-contains-string == 'true' &&
       needs.check-deploy.outputs.test-enabled == 'true' }}
     needs: [check-deploy, deploy-pr]
+    runs-on: ubuntu-22.04
     steps:
       - uses: SwissDataScienceCenter/renku-actions/test-renku@v1.4.6
         with:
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi:
There is probably some work to be done to separate cookies between different runs. It seems that one early logout creates problems with the `user` endpoint for the other tests:

```
Method: GET
URL: https://ci-renku-3150.dev.renku.ch/ui-server/api/user

The response we got was:

Status: 401 - Unauthorized
Headers: {
  ...
}
Body: {
  "message": "401 Unauthorized"
}
```

olevski:
@lorenzo-cavazzi this is marked as draft is it ready for review or not?

lorenzo-cavazzi:
@olevski not ready yet; tests seem to fail when running in parallel and I see a bunch of 401 errors on the user endpoint. I assume logging out affects all the running tests.

I'll test something else tomorrow and if I don't manage to run them, I'll revert to running them sequentially

lorenzo-cavazzi:
@olevski it's ready now. Sadly, I couldn't find an effective way to prevent login/logout from interfering with other running tests. For now I just limited the parallel tests to 1 so they can be re-run independently. I'll give it another try next sprint.

Relevant changes are:
- Update to cypress 12 and introduce a new mechanism to preserve credentials (the previous one was deprecated).
- Update the tests to be more stable (publicProject and updateProject) and delete projects from the UI (purblicProject).
- Separate the tests: can re-run 1 by 1 and they should be ready to run in parallel after fixing the problem with login/logout.

After the review, I'll port the logic to https://github.com/SwissDataScienceCenter/renku-actions and invoke the action from there instead of locally.

.github/workflows/pull-request-test.yml
```diff
@@ -126,46 +128,72 @@ jobs:
...
           TEST_LAST_NAME: Bot
+          TEST_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
           TEST_USERNAME: renku-test
+        with:
+          browser: chrome
+          install: true
+          spec: ${{ env.E2E_FOLDER }}verifyInfrastructure${{ env.E2E_SUFFIX }}
+          working-directory: cypress-tests
+      - uses: cypress-io/github-action@v5
+        name: Run Cypress acceptance tests in parallel
```
leafty, line 164: Why is this test not using the same structure as above? (with `spec: ${{ env.E2E_FOLDER }}${{ matrix.tests }}${{ env.E2E_SUFFIX }}`)

.github/workflows/pull-request-test.yml
```diff
@@ -126,46 +128,55 @@ jobs:
...
     if: ${{ github.event.action != 'closed' &&
       needs.check-deploy.outputs.pr-contains-string == 'true' &&
       (needs.check-deploy.outputs.test-enabled == 'true' ||
       needs.check-deploy.outputs.test-cypress-enabled == 'true') }}
     needs: [check-deploy, deploy-pr]
+    runs-on: ubuntu-22.04
+
+    strategy:
+      fail-fast: false
+      max-parallel: 1
```
leafty, line 140:
```suggestion
```

It seems that parallel execution works, see: https://github.com/SwissDataScienceCenter/renku/pull/3166.

leafty (changes requested):

.github/workflows/pull-request-test.yml
lorenzo-cavazzi: Oh... I tried that quite a few times and I kept getting 401 errors on the user endpoint.
Can we try to re-run it a few times from scratch? I fear the issue happens when one of the tests fails early or the logout in the publicProject spec runs before the others, but that's not always the case

lorenzo-cavazzi: Here is an example with the `401: Unauthorized` error on the `GET /user` endpoint that we get when running the tests in parallel
https://github.com/SwissDataScienceCenter/renku/actions/runs/5589204312/jobs/10217824415

leafty (approved):

PR #3154, opened 2023-07-11
fix(chart): properly handle postgres secret

olevski:
/deploy #persist


olevski:
I tested that both a fresh installation and an upgrade from an older version works.

Panaetius (approved):

PR #3157, opened 2023-07-12
Improve ingress handling

sgaist:
This merge request has the following aims:

- Make the ingress class selection follow current best practice
- Allow the deployment with and without cert-manager
- Improve the documentation related to certificate handling

The updated documentation provides now an example that developers can use to test-drive Renku using https.

### DIFF
--- a/docs/how-to-guides/admin/prerequisites/certificates.rst
+++ b/docs/how-to-guides/admin/prerequisites/certificates.rst
@@ -5,27 +5,74 @@ Certificates
 
 To enable HTTPS on the Renku website you need valid certificates for the ingress.
 
-This can be achieved by either deploying a certificate issuer like LetsEncrypt or creating certificates manually and inserting them as Kubernetes secrets.
+Certificates can be obtained in different ways:
 
-LetsEncrypt
-------------------
+#. Use a certificate issuer like LetsEncrypt
+#. Use a certificate bought from a certificate authority like SwissSign
+#. Use self-signed certificates
 
-If you have installed Helm you can deploy `LetsEncrypt <https://letsencrypt.org/>`_.
+Options one and three can be automated through the use of `cert-manager <https://cert-manager.io>`_
 
-The following are steps to install LetsEncrypt using Helm.
-To run the following commands you can use the `cert-manager-values.yaml <https://github.com/SwissDataScienceCenter/renku-admin-docs/blob/master/helm-installs/cert-manager-values.yaml>`_ and `cert-manager-issuer.yaml <https://github.com/SwissDataScienceCenter/renku-admin-docs/blob/master/manifests/cert-manager-issuer.yaml>`_ files.
-Make sure to edit the cert-manager-issuer.yaml file and fill in the correct `email` field.
+This document assumes that the default ``renku-tls`` secret name is used to configure the ingress.
+
+
+cert-manager
+------------
+
+If using Helm, you can deploy cert-manager using the following:
 
 .. code-block:: console
 
-   $ helm install cert-manager jetstack/cert-manager -f helm-installs/cert-manager-values.yaml --namespace kube-system
-   $ kubectl apply -f manifests/cert-manager-issuer.yaml
+  $ helm repo add jetstack https://charts.jetstack.io
+  $ helm repo update
+  $ helm install \
+  cert-manager jetstack/cert-manager \
+               --namespace cert-manager \
+               --create-namespace \
+               --version v1.12.0 \
+               --set installCRDs=true
 
-Now you can check that certificates can be issued automatically using a test deployment.
-For more information, please check `LetsEncrypt cert-manager Helm documentation <https://hub.helm.sh/charts/jetstack/cert-manager>`_
+More details on the deployment can be found in the `dedicated section <https://cert-manager.io/docs/installation/helm/>`_ of the project installation documentation.
+
 
 Generate manually
---------------------
+-----------------
+
+For development purposes, self-signed certificates can be used to properly test that all
+components are communicating securely together.
+
+Follow the `cert-manager guide <https://cert-manager.io/docs/configuration/selfsigned/>`_ to create a self-signed certificate issuer.
+
+You can then use it for your deployment by setting annotations such as the following in your
+configuration values:
+
+.. code-block:: yaml
+
+  ingress:
+    enabled: true
+    annotations:
+      cert-manager.io/cluster-issuer: null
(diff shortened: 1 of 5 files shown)


helm-chart/renku/values.yaml
```diff
@@ -178,11 +178,15 @@ ingress:
   ## Enables the creation of an ingress
   enabled: false
 
+  className: nginx
+
+  certManager:
```
aledegano, line 183: This doesn't work. The annotations are for the ingress, they are not necessarily related to `certManager`.
To achieve what you want here you can simply remove the two annotations: `cert-manager.io/cluster-issuer` and `kubernetes.io/ingress.class:`

aledegano (commented):
Thank you for this PR.
However it's unnecessarily complicated: the Helm templates can remain as they are and the use of cert-manager or not controlled by the annotations.

helm-chart/renku/values.yaml
sgaist: It does work. This new field has dedicated handling in the `ingress` template which will add the additional annotations.

The original idea was to clearly separate the handling of cert-manager dedicated annotations from the defaults provided by the chart itself.

aledegano: What I mean is what should one do if they want to add annotation to the ingress regardless of cert-manager?
Also why are you proposing this change?
What problem are you trying to solve?

sgaist:
> Thank you for this PR. However it's unnecessarily complicated: the Helm templates can remain as they are and the use of cert-manager or not controlled by the annotations.

You're welcome. I agree it's a bit more complicated than required. As explained above, the original idea was to keep the default ingress annotations untouched as otherwise, the admin has to copy them over to their custom `renku-values.yaml` just to manage cert-manager. However, if this is what is expected when deploying Renku, then I can simplify this PR and re-work the documentation to be explicit about that.

helm-chart/renku/values.yaml
sgaist: While writing this answer, I found out the issue that I had in mind but that I did not address properly (thanks for making me think a bit further): if one wants to have additional annotations, they currently have to copy the content from the `values.yaml` file and then add their own annotations.

```yaml
ingress:
  enabled: true
  annotations:
    # original values
    nginx.ingress.kubernetes.io/proxy-body-size: "0"
    nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
    nginx.ingress.kubernetes.io/proxy-buffer-size: "8k"
   # user custom values
```
If all these annotations shall be mandatory for Renku to operate properly (at least the last two seems to be), then I would rather expect the chart to give the option of only changing their values (for example, `ingress.proxyBufferSize: "8k"`) so that the ingress always has the minimal set of annotations required. On top of that, you can give the option for additional annotations which could use something like `extraAnnotations`. These `extraAnnotations` would then be a good fit for the `cert-manager` annotations.

I am proposing this change to help administrators and developers have an easier time to deploy and test renku and also to avoid breaking it unintentionally.

sgaist:
I understand your position and sensible defaults are very good. As explained in my [updated comment](https://github.com/SwissDataScienceCenter/renku/pull/3157#discussion_r1261093631), there were some things that I missed.

After some more digging, I think that updating the documentation and comments in `values.yaml` will be enough for what I wanted to achieve except for adding the option to make use of IngressClass which is a non-breaking feature that will be of interest for future deployments.

sgaist:
You're welcome !

I added entries to the changelog however I am unsure in which categories the changes I did fit. I am happy to take suggestions there.

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,26 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+----
+## Upgrading to Renku 0.30.0
```
rokroskar, line 9:
```suggestion
## Upgrading to Renku 0.34.0
```

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,26 @@ For changes that require manual steps other than changing values, please check o
...
+* NEW - *ingress.className* is now available to select a specific IngressClass to
```
rokroskar, line 10: Could you please clarify here if existing deployments which _do not_ need this functionality need to make any changes?

rokroskar (changes requested):
Thanks @sgaist! Just a small comment to clarify changes needed to existing deployments.

helm-chart/values.yaml.changelog.md
sgaist: Sure thing

helm-chart/values.yaml.changelog.md
sgaist: Done

helm-chart/values.yaml.changelog.md
rokroskar: thanks!!

docs/how-to-guides/admin/prerequisites/certificates.rst
```diff
@@ -5,27 +5,74 @@ Certificates
...
+More details on the deployment can be found in the `dedicated section <https://cert-manager.io/docs/installation/helm/>`_ of the project installation documentation.
 
-Now you can check that certificates can be issued automatically using a test deployment.
-For more information, please check `LetsEncrypt cert-manager Helm documentation <https://hub.helm.sh/charts/jetstack/cert-manager>`_
 
 Generate manually
---------------------
+-----------------
+
+For development purpose, self-signed certificates can be used to properly test that all
```
rokroskar, line 41:
```suggestion
For development purposes, self-signed certificates can be used to properly test that all
```

rokroskar:
Thanks @sgaist - sorry just found one tiny typo and then it lgtm :)

docs/how-to-guides/admin/prerequisites/certificates.rst
sgaist: Done

rokroskar:
You mean to clean up the commit history? In that case, we squash-merge all PRs anyway so it's a non-issue :)

rokroskar (dismissed):

aledegano (commented):
I'm suggesting one final, minor change, then we can merge this!

rokroskar (approved):

aledegano (approved):

PR #3193, opened 2023-08-07
feat(chart): move gateway helm chart to renku

olevski:
This makes the gateway helm chart be part of the renku "parent" helm chart. This is one of the first steps in making the renku helm chart and values file simpler.

/deploy #persist  extra-values=keycloakx.initRealm.image.repository=registry.dev.renku.ch/tasko.olevski/renku-images/init-realm,keycloakx.initRealm.image.tag=0.32.1-8.f21ffbb.git.2069.h7695e9f5

The following changes were made:
- all OIDC secrets are stored in one place the renku-gateway-revproxy secret
- removed 5+ `required` helm statements that were forcing users to use `openssl` in the terminal to generate OIDC client secrets for Gitlab, these can still be provided and if so they are honoured, but if they are not provided they will be generated automatically and saved in a k8s secret
- when a the renku-bundled gitlab is used then the client secret for the oauth application in gitlab is not required to be set because renku can generate and set it automatically (if set in the values then it is honoured)
- when an external gitlab is used then the client secret for the oauth application in the external gitlab is required and Renku cannot generate this automatically, the helm upgrade/install will fail in this case
- the OIDC secrets are overwritten if there are any values set in the values file because the keycloak init job will always overwrite and update clients even if they already exist


olevski:
I also tested in my namespace with the renku-bundled gitlab and it works.

olevski:
I am converting this back to a draft because the `required` fields are still `required` because the child helm charts like core service, ui or kg only know about pulling stuff from the values file. Not from a generated secret or anything like that. So the values file cannot be simplified until we merge all helm charts.

rokroskar:
@olevski is this still relevant now that all the charts are in the same place? Or can we close it?

Issue #3181, opened 2023-07-27
Update "Get started" tutorial to match the UI style


lorenzo-cavazzi:
The documentation occasionally shows outdated pictures from the UI and the instructions should be brushed up as well -- here is an example that might confuse users

<kbd>
  <img src="https://github.com/SwissDataScienceCenter/renku/assets/43481553/adc1a5f9-c0f5-499b-9aae-a6524f676041">
</kbd>

.
The problem is particularly evident in the "Get started" section https://renku.readthedocs.io/en/stable/tutorials/01_firststeps.html
It's worth going through the other pages as well.

PR #3196, opened 2023-08-09
docs: Update "Get started" tutorial to match the UI style (#3181)

andre-code:
Closes https://github.com/SwissDataScienceCenter/renku/issues/3181


rokroskar (changes requested):
Thanks a lot @andre-code !!

Two minor things: 

1. There is one page where the formatting is a bit weird - it shows the gitlab and renkulab views on top of each other and it's hard to tell which is which:

image

Adding some text in between the two figures would solve that. 

2. I also find it a little confusing to have the red box around the main action button, when the image shows the drop-down - intuitively if I were following the instructions I would probably click the main button expecting the drop-down, but that's not what happens.

andre-code:
Thanks @rokroskar I have applied the suggestions and also rewrite it a bit the part of creating Issues in Gitlab.

rokroskar:
Thanks @andre-code! I think the same issue with the drop-down can be seen in this image:

image

I'm not sure though if from the text context it's enough to understand... maybe the red box number "2" can also extend to the drop-down? not sure what the best way is

andre-code:
> Thanks @andre-code! I think the same issue with the drop-down can be seen in this image:
> 
> image
> 
> I'm not sure though if from the text context it's enough to understand... maybe the red box number "2" can also extend to the drop-down? not sure what the best way is

sure!
I have changed the image to make the steps clearer for these cases.

rokroskar (approved):
Thanks!!

PR #3222, opened 2023-08-29
chore: add single global Gitlab URL

Panaetius:
this should replace all the places where we use gitlab URL in our services.

- [x] https://github.com/SwissDataScienceCenter/renku-gateway/issues/675
- [x] https://github.com/SwissDataScienceCenter/renku-python/issues/3613
- [x] https://github.com/SwissDataScienceCenter/renku/pull/3335
- gateway
- tests

/deploy extra-values=global.gitlab.url=https://gitlab.dev.renku.ch,global.gitlab.registry.host=registry.dev.renku.ch


rokroskar:
Do you plan on following this up with changes in all the other charts?

Panaetius:
@rokroskar each team (well kg and rp) should have issues this sprint to switch over to using this.

rokroskar:
Got it! Can we add the issues as a todo list at the top of the PR to keep track?

rokroskar:
@Panaetius by the looks of it, this could be merged?

Panaetius:
it would be good if someone from @SwissDataScienceCenter/renku-graph-maintainers could check that I didn't make a mistake in any of their values.

helm-chart/renku/templates/_helpers.tpl
```diff
@@ -137,7 +137,7 @@ KC_DB_PASSWORD: {{ default (randAlphaNum 64) .Values.global.keycloak.postgresPas
 {{- end -}}
 
 {{- define "renku.gitlabUrl" -}}
-{{ .Values.global.gitlab.url | default (printf "%s://%s/gitlab" (include "renku.http" .) .Values.global.renku.domain) }}
+{{ .Values.global.gitlab.url | default (printf "%s://gitlab.%s/" (include "renku.http" .) .Values.global.renku.domain) }}
```
rokroskar, line 140: do we need this? It's weird to have `renku.gitlabUrl` in some places and in others `.Values.global.gitlab.registry.host` for example. It would be cleaner maybe to just require `global.gitlab.url` to be set and not try to guess it.

Panaetius: :man_shrugging: I don't really know, I thought this is still relevant when using the embedded gitlab. @SwissDataScienceCenter/yat do you know?

rokroskar: I believe that will be done fully in https://github.com/SwissDataScienceCenter/renku/pull/3297 unless that's not relevant anymore?

aledegano: ah, I thought it was already merged...
IIRC, however, the internal Gitlab is broken anyway...
In any case, we can easily _still_ remove this value by replacing https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/templates/gateway/deployment.yaml#L50 with `global.gitlab.url` which will work for any deployment regardless.

Panaetius: I've removed it

rokroskar (changes requested):
Great, this change makes a lot of sense now, thanks @Panaetius - just one thing we might want to do: existing deployments will fail because the registry host for the notebooks service is now expected to come from `global.gitlab.registry.host` - could you add a line to the values changelog that clarifies this?

rokroskar (approved):

rokroskar:
@Panaetius I approved this but I'm not sure we should merge it right away - I _think_ it should also require a change to values for CI deployments?

PR #3224, opened 2023-08-30
chore(readme): update copyright year in readme

seanrmurphy:
PR for testing changes to CI values.

/deploy #persist


Panaetius:
In renku-python we switched to only having the year in the LICENSE file and not in all the headers. Maybe we should do that everywhere

seanrmurphy:
@Panaetius - I looked at the `LICENSE` file in renku-python and it has a placeholder for year and copyright owner - is this somehow templated out into the documentation?

https://github.com/SwissDataScienceCenter/renku-python/blob/develop/LICENSE#L190

Panaetius:
> @Panaetius - I looked at the `LICENSE` file in renku-python and it has a placeholder for year and copyright owner - is this somehow templated out into the documentation?
> 
> https://github.com/SwissDataScienceCenter/renku-python/blob/develop/LICENSE#L190

:see_no_evil: no it shouldn't be...

seanrmurphy:
I've modified this to remove the copyright and licensing stuff from the readme and amend the license file. Hope that seems reasonable.

Of course, we can just drop this PR if it does not make sense (the original driver was to make some small change to trigger a CI deployment to test some of that machinery...)

Panaetius (dismissed):

Panaetius (approved):

PR #3216, opened 2023-08-24
test(cypress): add Cypress tests for workflow pages

lorenzo-cavazzi:
Add Cypress tests to cover workflows.

fix #2852
/deploy #persist #notest

PR #3231, opened 2023-09-01
test(cypress): add test for private projects and visibility

lorenzo-cavazzi:
Add Cypress tests to cover private projects and test visibility.

fix https://github.com/SwissDataScienceCenter/renku/pull/3216
/deploy #persist #cypress #notest

### DIFF
--- a/cypress-tests/cypress/e2e/privateProject.cy.ts
+++ b/cypress-tests/cypress/e2e/privateProject.cy.ts
@@ -95,7 +95,7 @@ describe("Basic public project functionality", () => {
     // ? We need to wait before other checks take place.
     // ? This is a workaround until we use the new Project update endpoint.
     // ? Reference: https://github.com/SwissDataScienceCenter/renku-ui/issues/2778
-    cy.wait(10000); // eslint-disable-line cypress/no-unnecessary-waiting
+    cy.wait(TIMEOUTS.standard); // eslint-disable-line cypress/no-unnecessary-waiting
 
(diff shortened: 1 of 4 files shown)


cypress-tests/cypress/e2e/privateProject.cy.ts
```diff
@@ -0,0 +1,157 @@
...
+      .contains("Change visibility to Public")
+      .should("be.visible");
+    cy.dataCy("update-visibility-btn").should("be.visible").click();
+    cy.get(".modal .alert-success", { timeout: TIMEOUTS.long })
+      .contains("The visibility of the project has been modified")
+      .should("be.visible");
+    // ? We need to wait before other checks take place.
+    // ? This is a workaround until we use the new Project update endpoint.
+    // ? Reference: https://github.com/SwissDataScienceCenter/renku-ui/issues/2778
+    cy.wait(10000); // eslint-disable-line cypress/no-unnecessary-waiting
```
ciyer, line 98: I would suggest using our existing constants for this.

```suggestion
    cy.wait(TIMEOUTS.standard); // eslint-disable-line cypress/no-unnecessary-waiting
```

Or, if it is important that we not wait longer than necessary, then 

`TIMEOUTS.standard * 0.5`

ciyer (changes requested):
Very nice! One small suggestion to use a named constant instead of an inline magic value.

ciyer (approved):
👍

PR #3232, opened 2023-09-01
test(cypress): optimize acceptance tests and cover missing UI features

### LINKED ISSUE #3184: Cypress tests to cover missing features
A few missing features we want to be sure we cover:

* Searching for datasets: we can re-use `publicProject` and search for the newly created dataset before deleting it.
* (_optional_) Change settings -- tags, description, avatar. Mind that this might be slightly shaky cause it requires KG to pick up the changes. The logic should check the values were updated, but better avoiding to check the previous empty values. That would cause re-runs to fail if the issue was a slow update on KG side.

lorenzo-cavazzi:
This PR improves the Cypress acceptance tests to cover more features on the UI. The main changes are:

* Creating common Cypress functions wherever code is reusable by multiple specs.
* Create a dedicated dataset spec with new and improved tests. Mind that we might want to include even more dataset-specific tests here.
* Be sure lineage is created -- mind that we plan to overhaul this feature, but still we want to be sure the lineage is created correctly even if that section hasn't received much love lately.
* Cover more scenarios in the session tests, including anonymous sessions and sessions on projects where users have no privileges. I plan to blend in these tests also the R studio test to speed it up a bit and prevent repeating the same checks, but that change will be in a follow-up PR.

fix #3184


cypress-tests/README.md
```diff
@@ -1,108 +1,46 @@
...
-- Stop the session
-- Delete the created project
+- Create a `cypress.env.json` file with the necessary user credentials. You can use the
+  `cypress.env.template.json` file as a template.
+- Install the dependencies with `npm install`.
+- Run the tests with `npm run e2e`. If you prefer not to use the GUI, you can run the
+  tests in headless mode with `npm run e2e:headless`.
 
-## Public Project
+> Tip: you might prefer not to save you password in plain text in the `cypress.env.json`
```
ciyer, line 21: I thought the section on how to run the tests with information passed in environment variables in general was useful and should be kept. Maybe including a table with the different ENV variables that are supported and their purpose.

cypress-tests/cypress/support/commands/general.ts
```diff
@@ -1,50 +1,40 @@
...
-  Cypress.Commands.add("dataCy", dataCy);
+  Cypress.Commands.add("getDataCy", getDataCy);
 }
 
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace Cypress {
     interface Chainable {
-      dataCy(value: string, exist?: boolean);
+      getDataCy(value: string, exist?: boolean);
```
ciyer, line 36: This is a good change! I always got confused by `dataCy` and had to remind myself that it got an element from the DOM.

cypress-tests/cypress/support/commands/login.ts
```diff
@@ -76,34 +109,35 @@ function registerAndVerify(props: RegisterAndVerifyProps) {
...
-    return registerAndVerify(localProps);
-  });
+      return registerAndVerify(localProps);
+    }
+  );
 }
 
 function logout() {
-//  cy.visit("/");
+  //  cy.visit("/");
```
ciyer, line 140: Should we just remove this? Not sure why we have this line.

ciyer (changes requested):
This looks good! 🤩 I need a bit of time to look at some of the details, but I wanted to put a few comments out right away.

ciyer (commented):
Changes look really good! I think the only thing necessary is documentation of the environment variables, then this can be merged.

cypress-tests/README.md
lorenzo-cavazzi: I added that in the last commit. I felt the names were self-explanatory, but having a table listing all the variables sounds useful anyway.

ciyer (approved):
Great! This should hopefully make tests more robust and easier to maintain.

PR #3246, opened 2023-09-08
refactor: removed tests that are covered by Cypress

jachro:
This PR removes all the selenium tests that are done in Cypress acceptance tests.

/deploy #persist


rokroskar:
When this is merged, which tests do we need to make required? cc @lorenzo-cavazzi @ciyer @leafty

lorenzo-cavazzi:
This will stay for testing a few things in the CLI -- it should still be required.
The Cypress tests have a different structure, we might need to make all of them mandatory if we wish to keep the same approach. Mind that skipping the deploy phase still allows merging PRs without tests.

P.S. We are adding the last bits to these tests to cover all the functionalities https://github.com/SwissDataScienceCenter/renku/pull/3232
As soon as it's ready, we can enforce all the Cypress tests

rokroskar:
@lorenzo-cavazzi but if we merge this we should enforce the cypress tests already now, no? Or do we wait with this PR until #3232 is done?

lorenzo-cavazzi:
Let's wait until #3232 is done. I need another couple of tiny changes

acceptance-tests/src/test/scala/ch/renku/acceptancetests/LineageSpec.scala
lorenzo-cavazzi, line 1: I ended up adding a section to test the Lineage page in the Cypress tests ([here](https://github.com/SwissDataScienceCenter/renku/blob/master/cypress-tests/cypress/e2e/useSession.cy.ts#L182-L189)).
We can probably remove the `view the lineage of the outputFile` function; that's the last one testing the UI.

lorenzo-cavazzi (commented):
Lgtm!
I added a comment to the Lineage test since we now cover it in the Cypress tests and it's the last piece testing the UI. We can then go on and merge.
Thank you for the PR :pray:

acceptance-tests/src/test/scala/ch/renku/acceptancetests/LineageSpec.scala
lorenzo-cavazzi: Oh... this comment was meant to be at line 96

jachro: I've just removed our lineage spec

lorenzo-cavazzi (approved):
Let's merge it! :rocket:

PR #3260, opened 2023-09-20
test(cypress): prevent search page failures

lorenzo-cavazzi:
Fix the tests involving some sort of search.

We can go for alternative solutions once we update the search page.
For the time being. The additional waiting time prevents an early components refresh from breaking the tests -- that is very frequent especially when searching datasets since it requires changing the search filters.

/deploy


lorenzo-cavazzi:
Included in #3261

PR #3261, opened 2023-09-20
test(cypress): improve stability

lorenzo-cavazzi:
This should fix a couple of recurring problems with Cypress tests:
- Metadata indexing doesn't necessarily start immediately.
- Using the search page frequently fails.

/deploy


ciyer (approved):
Code looks good!

PR #3278, opened 2023-10-02
test(cypress): fix editing dataset

lorenzo-cavazzi:
This prevents early failures when trying to modify datasets

/deploy


lorenzo-cavazzi:
Will be merged through another PR

PR #3280, opened 2023-10-02
chore: workflow name should be unique in tests

olevski:
/deploy


olevski:
The workflow name and output file should be unique. So that rerunning the test on the same repo does not result in an error.

aledegano (dismissed):

lorenzo-cavazzi:
@olevski I changed the output file name cause we need to keep it short for the current tests and `.substring(4)` resulted in a very long name since it keeps the string _from_ position 4

PR #3283, opened 2023-10-03
chore: update renku-notebooks to 1.19.1

mohammad-alisafaee:


olevski (approved):

aledegano:
Thanks for the fix @mohammad-alisafaee .
It should go as 0.37.1 however as we will not jump from 0.36 to 0.39, plz!

aledegano:
> Thanks for the fix @mohammad-alisafaee .
> It should go as 0.37.1 however as we will not jump from 0.36 to 0.39, plz!

Nevermind: I synced with @olevski and decided together that we'll just go straight to 0.39.2.
So just ignore the previous comment.

lokijuhy (approved):

PR #3284, opened 2023-10-05
fix(chart): session network policies, tolerations and affinities

olevski:
This is the result of the helm chart consolidation.

We did not see this because the problems that this solves do not appear in CI deployments or are not obvious in CI deployments.

/deploy


aledegano (dismissed):

mohammad-alisafaee (dismissed):
LGTM, thanks!

lokijuhy (approved):

mohammad-alisafaee (approved):

aledegano (approved):

PR #3288, opened 2023-10-06
fix(charts): gitlab chart template names

olevski:
The gitlab helm chart is using renkus templates i.e. `renku.http` but before these were not prefixed by `renku.` so we had to fix this there and publish a new version.

The new version of the helm chart has been published. We just have to use it in the renku helm chart now.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,18 @@
 .. _changelog:
 
+0.39.3
+------
+
+Renku ``0.39.3`` fixes various bugs.
+
+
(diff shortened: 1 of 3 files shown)


aledegano (dismissed):
Thank you for the fix!

rokroskar:
I guess this is superseded by #3297?

olevski:
Replaced by https://github.com/SwissDataScienceCenter/renku/pull/3297 which will fully remove the internal gitlab helm chart.

PR #3293, opened 2023-10-09
fix(chart): default value for client ID in Gitlab

olevski:
The gateway helm chart did not have a default value set for this. But the default value was set in the renku helm chart. When I was combining the helm charts I copied over the values from the component charts into Renku. And when I did this I removed the default we had set for this in the renku chart.

This puts it back. 

For example if you go to https://github.com/SwissDataScienceCenter/renku/blob/0.38.0/helm-chart/renku/values.yaml#L41 you will see it. 0.38.0 is the last version before we combined the helm charts.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,18 @@
 .. _changelog:
 
+0.39.3
+------
+
+Renku ``0.39.3`` fixes various bugs.
+
+
+Internal Changes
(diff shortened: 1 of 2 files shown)


wesjdj (approved):

aledegano (approved):
Thanks

lokijuhy (approved):

PR #3298, opened 2023-10-09
chore: modify renku docs and script for values

olevski:
After we combined the values file we should change the docs and script that generates the values file.


docs/how-to-guides/admin/deploying-renku.rst
```diff
@@ -59,45 +59,36 @@ This response basically means that:
...
 
-.. note::
-   These examples are specific to our use of OpenStack and the details of
-   the storage classes and the volume provider will be different depending on the
-   specifics of your cluster.
+1. (Optional) Create and configure Renku PVs and PVCs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+All of Renku information is stored in two volumes needed by Jena and Postgresql. 
+You can leave k8s to dynamically provision these volumes but we advise to create them
```
aledegano, line 90:
```suggestion
You can leave Kubernetes to dynamically provision these volumes but we advise to create them 
```
In the docs I'd rather use full names to avoid confusion from less experienced users/admins.

aledegano (dismissed):
I've left a minor comment, otherwise it's all good, thanks!

I also want to point out that with the official removal of our integrated "omnibus" Gitlab chart, we should probably give this issue more priority: https://github.com/SwissDataScienceCenter/terraform-renku/issues/1218

aledegano (approved):

olevski:
> I also want to point out that with the official removal of our integrated "omnibus" Gitlab chart, we should probably give this issue more priority: https://github.com/SwissDataScienceCenter/terraform-renku/issues/1218

@aledegano I think that this job should not be part of the Renku helm chart.  You can add something that automates this for the deployments we control. But trying to automate this away is hard and will still require some input from the user. I.e. instead of asking for the client ID and secret and for the external gitlab as we do now we will have to ask for a PAT that can make the Oauth application in the gitlab or for a user password/username. For the renku helm chart I would much rather simply ask for the Oauth client credentials.

But you can automate this in the terraform/flux repo where a job is created outside of the Renku helm chart that does this and creates a specific secret that is used by the renku helm chart.

PR #3308, opened 2023-10-17
chore: prepare 0.40.1

jachro:
This PR creates a bugfix 0.40.1 release of renku with graph 2.43.1 that reverts the recent changes to Lucene config.

/deploy

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,19 @@
 .. _changelog:
 
+0.40.1
+------
+
+Renku ``0.40.1`` reverts recent changes to Lucene configuration in the Triples Store preventing users from searching by keywords.
+
+**🐞 Bug Fixes**
+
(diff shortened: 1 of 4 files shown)


olevski:
@jachro there is no point in merging this. Everything looks good - you can make a release from this branch.

I approved the other PR that adds these changes in master.

olevski:
p.s. the CI actions wont run if there is a conflict. But you cannot resolve the conflict because there is already 0.41.0 in the changelog in master. So if you really want them to run then try to merge into a dummy/random branch (probably derived from 0.40.0 where the conflict will not show up).

olevski:
It was not clear to me from the changelog. But this removes the improvements we have made in [this PR](https://github.com/SwissDataScienceCenter/renku-graph/pull/1725). And therefore I dont think we should roll out this bugfix which fixes one thing but breaks another.

jachro:
This was a hotfix release so no need to merge this PR. #3309 incorporates this change into `master`.

PR #3322, opened 2023-10-25
chore: update release notes for 0.40.2

olevski:
A renku bugfix release.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,18 @@
 .. _changelog:
 
+0.40.2
+------
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -1,5 +1,18 @@
 .. _changelog:
 
+0.40.2
```
aledegano, line 3: Thank you @olevski !
I just wonder if this section should be after `0.40.1` and before `0.41.0`, WDYT?

olevski: Ah yeah that is what I mean to do. My bad.

rokroskar (approved):

PR #3327, opened 2023-10-27
feat(helm): generate random secrets when possible

aledegano:
Generates random secrets on deploy whenever possible.


rokroskar:
Doesn't this approach lead to things breaking when you redeploy? For example, the Jena password, redis password etc. Or can those be changed on-the-fly?

aledegano:
> Doesn't this approach lead to things breaking when you redeploy? For example, the Jena password, redis password etc. Or can those be changed on-the-fly?

There might be a problem if you keep the PVs BUT not the secrets, then if you also have not kept track of the secrets generated you are screwed indeed. But this is already the case for many secrets in Renku.

On the other hand if you don't delete the secrets you'll be fine as the chart reads them.

This feature is meant for CI and dev deployments, admins are still supposed to define all the secrets in the values.

rokroskar:
> There might be a problem if you keep the PVs BUT not the secrets, then if you also have not kept track of the secrets generated you are screwed indeed. But this is already the case for many secrets in Renku.
> 
> On the other hand if you don't delete the secrets you'll be fine as the chart reads them.
> 
> This feature is meant for CI and dev deployments, admins are still supposed to define all the secrets in the values.

We had some mechanism with hooks that allowed you to do this and _not_ overwrite the secrets on redeploy. I'm not sure if that is still being used. I believe @Panaetius (and or @olevski ?) implemented that at some point. 

The problem of overwriting secrets for dev/CI deployments makes it worse imho because that's exactly where you are constantly redeploying. So the strategy would need to be to delete the deployment between iterations which is not great.

olevski:
This is on my to-do list to review. I just wont be able to get it today. There is a way to simply never delete the auto-generated secrets. That is what we do already in most cases when we autogenerate stuff.

olevski:
@aledegano this is still a draft though? Are you thinking of pushing more changes? Should I wait for more code/changes before I review?

aledegano:
> @aledegano this is still a draft though? Are you thinking of pushing more changes? Should I wait for more code/changes before I review?

Tasko, yes this is a draft as I need to test it more and I would prefer https://github.com/SwissDataScienceCenter/renku-graph/issues/1768 to be done and integrated.
That said, if you could quickly peek at how the secrets are read through the Helm chart to avoid changing them (which is what I believe we do now for some of the secrets), I would be grateful.

rokroskar:
@aledegano do you plan on picking this up again?

aledegano:
> @aledegano do you plan on picking this up again?

Yes, I still want to finish it and get it merged...

aledegano:
Superceded by #3909

PR #3337, opened 2023-10-31
chore(swagger): add enableServiceLinks swagger pod

seanrmurphy:
The swagger pods has been crashing on renkulab due to a relatively large amount of env variables being defined - initialization scripts which parsed the env vars did not work.

These are not required in the swagger pod and this change means that the large set of env vars are not defined in this pod.

### DIFF
--- a/helm-chart/renku/templates/swagger.yaml
+++ b/helm-chart/renku/templates/swagger.yaml
@@ -51,6 +51,7 @@ spec:
             httpGet:
               path: /swagger
               port: http
+      enableServiceLinks: false
 ---
 apiVersion: v1
 kind: Service


aledegano:
Amazing!
@rokroskar @Panaetius and I had racked our brains trying to figure this one out!
Have you tested it already?

seanrmurphy:
> Amazing! @rokroskar @Panaetius and I had racked our brains trying to figure this one out! Have you tested it already?

I made the change on infra-staging and it resulted in a smaller set of env vars, so it worked in terms of reducing the number of env vars. I did not test in situ on renkulab which is where the real problem is due to the large number of services defined.

olevski (approved):

PR #3340, opened 2023-10-31
feat: Add authorization service

olevski:
This adds Authzed's [SpiceDB](https://authzed.com/spicedb) as an authorization service.


rokroskar:
A general question - should this target a feature branch?

olevski:
> A general question - should this target a feature branch?
I am not sure what you mean here @rokroskar ?

helm-chart/renku/templates/authz/deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+  labels:
+    app: renku-authz
+    chart: {{ template "renku.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  {{- if not .Values.authz.autoscaling.enabled }}
+  replicas: {{ .Values.authz.replicaCount }}
+  {{- end }}
+  selector:
```
wesjdj, line 14: Is there a reason no update strategy has been included?

helm-chart/renku/templates/authz/service.yaml
```diff
@@ -0,0 +1,22 @@
...
+  type: ClusterIP
+  ports:
+    - port: 8443
+      targetPort: http
+      protocol: TCP
+      name: http
+    - port: 50051
+      targetPort: grpc
+      protocol: TCP
+      name: grpc
```
wesjdj, line 20: We need a port for the metrics too, as well as the prometheus annotations

helm-chart/renku/templates/authz-secrets.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+{{- $renkuFullname := include "renku.fullname" . -}}
+
+{{- $secretName := cat $renkuFullname "-authz" | nospace }}
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace $secretName) }}
+{{- if $secret }}
+{{- $db_password_enc = index $secret.data "SPICEDB_POSTGRES_PASSWORD" }}
+{{- $db_conn_uri = index $secret.data "SPICEDB_DATASTORE_CONN_URI" }}
+{{- $api_key = index $secret.data "SPICEDB_GRPC_PRESHARED_KEY" }}
+{{- end -}}
```
wesjdj, line 15: This doesn't seem to allow for existing postgres credentials to be provided in the values file

helm-chart/renku/templates/network-policies.yaml
```diff
@@ -75,6 +75,13 @@ spec:
           namespaceSelector:
             matchLabels:
               kubernetes.io/metadata.name: {{ .Release.Namespace }}
+        - podSelector:
+            matchLabels:
+              app: renku-authz
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{ .Release.Namespace }}
+
```
wesjdj, line 84: Remove empty line

helm-chart/renku/templates/setup-job-authz-db.yaml
```diff
@@ -0,0 +1,63 @@
...
+            - name: DB_ADMIN_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.global.externalServices.postgresql.existingSecret }}
+                  key: postgres-password
+            {{- end }}
+            - name: AUTHZ_DB_USERNAME
+              value: authz
+            - name: AUTHZ_DB_NAME
+              value: authz
```
wesjdj, line 58: We should allow for these values to be set in the values instead of hard coding them. See `setup-job-gitlab.yaml`

helm-chart/renku/values.yaml
```diff
@@ -1613,6 +1613,21 @@ dataService:
...
+  replicaCount: 1
+  image:
+    repository: authzed/spicedb
+    pullPolicy: IfNotPresent
+    tag: "v1.30.0"
+  resources: {}
+  autoscaling:
+    enabled: false
+    minReplicas: 1
+    maxReplicas: 10
```
wesjdj, line 1627: This seems quite high. Has this been necessary during testing?

wesjdj (changes requested):

helm-chart/renku/templates/setup-job-authz-db.yaml
olevski: I dont want anyone to change this. The deployment/app manages this. We do not allow people to change any of the database names we create for the data service as well.

If this is in the values file and it is changed then the deployment will break because no one will be able to access anything since there will be a new blank database without any permissions.

If you look at the `setup-job-renku-dbs.yaml` file you will see that the DB names for the KG and the data services (we call the db renku) are also all hardcoded.

`setup-job-gitlab.yaml` is only used for the internal gitlab only. So it is actually never used and it will be removed soon.

helm-chart/renku/values.yaml
olevski: I have not been able to do any load testing. It is still too early for this because the authorization service is not hooked up to anything in the backend yet. So I cannot do any meaningful load tests. I changed this to 3 and we can revise it further.

But this is also very dependent on each deployment. If the deployment serves 10 people it will need less replicas than something like renkulab.io.

Either way I can do some tests when I implement stuff on the backend and revise this. The autoscaling is also off by default.

helm-chart/renku/templates/authz-secrets.yaml
olevski: This is similar to the database name. I do not want to allow people to change this. What we accept is a single postgres admin password that can be passed through the values file. Then using this password renku makes different postgres accounts with their own passwords and gives them access to specific hardcoded databases. If I accept this in the values file I also have to add logic and jobs to compare it with a previous value and to then if needed update the database so that everything works.

helm-chart/renku/templates/authz/service.yaml
olevski: Ok after the discussion in the weekly sync meeting it is clear to me that I will be doing this. So I will do this in a separate followup PR.

wesjdj (approved):

PR #3342, opened 2023-11-01
chore: update data service - prep 0.42.1

rokroskar:
This deploys the fix for the following bugs found in 0.42.0:

- [x] listing users in the data-services API
- [x] creating and persisting node affinities and tolerations
- [x] enforcing the flag that prevents access to the default pool
- [x] accidentally removing the repo directory from hibernated sessions

/deploy

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,13 +3,20 @@
 0.42.1
 ------
 
-Renku ``0.42.1`` is a bugfix release to patch a bug found in the data service which prevented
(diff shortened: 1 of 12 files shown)


rokroskar:
Just making sure we didn't forget anything - I tested all the things we found wrong with 0.42.0:

* preventing a user to use the default pool
* listing users in the API
* setting tolerations/affinities on resource pool `POST` and class `PATCH`

PR #3346, opened 2023-11-02
do not merge: testing values for the showcase

ciyer:
This is to test values for the showcase. Attempt to deploy Renku outside a PR failed, so using this as an easy way to get a deployment

/deploy


ciyer:
Not needed anymore. See https://github.com/SwissDataScienceCenter/terraform-renku/pull/1984

PR #3351, opened 2023-11-06
docs: configures pipx path according to documentation

champost:
Following the recommended way to configure the PATH variable for pipx : https://github.com/pypa/pipx#install-pipx


rokroskar (approved):
👍 thanks @champost!

PR #3356, opened 2023-11-08
chore: set keycloak max db connections to 10

olevski:
@leafty found that the default value for this in the Keycloak helm chart is 100.\

/deploy


aledegano (approved):
That's great!

olevski:
FYI this only saves us a 4-5 connections. Because KC is usually at a very low number and it bursts very briefly to 10-15 and goes back down.

PR #3370, opened 2023-11-15
Update from auto-update/renku-graph-2.47.1

RenkuBot:
This is an automated pull request.
/deploy extra-values=global.renku.cli_version=2.7.0

PR #3379, opened 2023-11-21
chore: selenium tests to install CLI version fetched from core

jachro:
This PR changes the behaviour of selenium tests in a way that they call the `/api/renku/apiversion` API and read the new `cli_version` property. This way the tests will always use CLI that is compatible (in fact the same) as installed on renku-core.

**NOTICE:**
* this PR cannot be merged before https://github.com/SwissDataScienceCenter/renku-python/pull/3655 is merged and released
* a following env var definition has to be taken out from the `helm-chart/renku/templates/tests/test-renku.yaml`; it cannot be done now as it will be brought with https://github.com/SwissDataScienceCenter/renku/pull/3370.
```
      {{ if .Values.global.renku.cli_version }}
      - name: RENKU_CLI_VERSION
        value: '{{ .Values.global.renku.cli_version }}'
      {{ end }}
 ```

/deploy

### CODE REVIEW

**Comment by jachro** [comment]
The selenium mtests run and failed as the response from https://ci-renku-3379.dev.renku.ch/api/renku/apiversion
```
{"result":{"maximum_api_version":"2.2","latest_version":"2.9.1","supported_project_version":10,"minimum_api_version":"1.0"}}
```
didn't contain the expected `cli_version` property. @Panaetius was that change adding the property part of the lastest release on renku-python?

rokroskar:
@Panaetius I guess we can close this?

PR #3383, opened 2023-11-23
chore: periodic jena compaction

jachro:
This PR:
* upgrades Jena chart to 0.0.23
* adds relevant config values to enable periodic compaction every Monday at midnight

I did several manual tests, where one of them was to deploy the chart in some namespace while the other was to change Jena stateful set values on the `renku-kg-dev` env.


helm-chart/renku/values.yaml
```diff
@@ -1280,6 +1280,11 @@ jena:
   additionalEnvironmentVariables:
   - name: JVM_ARGS
     value: -Xmx2G -Xms2G
+  compacting:
+    - name: ADMIN_USER
+      value: admin
+    - name: COMPACTING_SCHEDULE
+      value: 'Mon *-*-* 00:00:00'
```
olevski, line 1287: Can we make this the default in the Jena chart and not mention these at all here?

If we keep them here can we add a comment to explain what format the schedule should follow. That does not seem like standard cronjob string to me.

jachro: I was thinking about that but the problem is that the `renku-jena` chart is pretty generic. For instance, it leaves for the user/admin the decision of what users will be created and what privileges they will have. In terms of the pattern, I only thought it'd be maybe good to leave it unset on the `renku-jena` chart so the user has to decide by themselves or simply uncomment my proposal. The `values.yaml` on `renku-jena` says this:

```
# In order to enable periodic datasets compacting, the following entries need to be specified.
# The defaults would work only with the default `shiro.ini`.
compacting: []
#  - name: ADMIN_USER
#    value: admin
#  - name: COMPACTING_SCHEDULE
#    value: Mon *-*-* 00:00:00 # every Monday at 00:00:00. More details at https://github.com/eikek/calev
```

What do you think?

eikek: Just a note on the pattern: this is not a cron style pattern, but a systemd.time one (I find this variant a lot easier to remember and understand)

olevski: Ok @jachro can you add a comment in the renku chart in this PR to explain what the format of the pattern is.

jachro: @olevski I added some explanation next to the entry in `values.yaml`.

olevski (approved):

PR #3389, opened 2023-11-29
feat(helm): Make Jena passwords optional in values file.

aledegano:
This commit continues on the topic of making the definition of the secrets in the values files optional, to simplify CI deployments.

As previous similar commits this change supports:

- Existing deployments where the secret is defined will keep working as usual

- New deployments where the secret is defined will keep working as they did before this change

- New deployments where the secret is NOT defined will get 2 random passwords which will be persisted across upgrades/redeployments

The three scenarios have been manually tested.

This PR doesn't change anything to users or admins, therefore I didn't write an entry in the changelog to avoid unnecessary noise, let me know if you would rather have one.

/deploy


rokroskar:
@aledegano can we get this either merged or closed? Should be rebased on `release-0.60.0` and (re)tested. Also, I don't know if it affects anything for the script that generates values, but perhaps the Jena secrets can be removed from the values template?

aledegano:
@rokroskar I have finally came back to this and I have:
- Rebased on the latest scheduled release
- Retested (also on my own to really check that it works without specifying the secrets)
- Removed the keys from the values template file

scripts/generate-values/base-renku-values.yaml.template
```diff
@@ -21,12 +21,6 @@ gateway:
 graph:
   gitlab:
     url: {gitlab_url}
-  tokenRepository:
-    tokenEncryption:
-      secret: <use `openssl rand -hex 8 | base64`>
-  webhookService:
-    hookToken:
-      secret: <use `openssl rand -hex 8 | base64`>
```
olevski, line 29: question: Why are these removed? They are not used at all? Because these are not the values you automated in this PR.

CHANGELOG.rst
```diff
@@ -3,6 +3,9 @@
 0.65.0
 ------
 
+0.64.1
```
olevski, line 6: I think it makes sense to add a single line about this change in the changelog.

olevski (changes requested):
This is great Alessandro! Just two small things.

CHANGELOG.rst
aledegano: Will do!

olevski (approved):

aledegano:
Change of plans: at this moment the job to test the `generate-values.sh` script is stuck because it cannot find a MacOS 11 where to test it. Rather than fixing the test or changing the target I will simply revert the change to the templated values file, because in the following PR that script will be entirely removed (as secrets won't need to be generated out-of-band anymore).

PR #3447, opened 2024-01-25
test(cypress): do not run cloud storage tests where not supported

lorenzo-cavazzi:
This prevents running tests for cloud storage on PRs where it's not supported.

/deploy


leafty (commented):
Code looks good 🎉 will ✅ after it runs

leafty (approved):
LGTM

PR #3449, opened 2024-01-29
update values file changelog

Panaetius:


helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,26 @@ For changes that require manual steps other than changing values, please check o
...
 
+## Upgrading to Renku 0.47.0
+
+We completely overhauled how mounting cloud storage in sessions works, relying on a new CSI driver based on RClone 
+which has to be installed in the cluster for things to work. Either install it as part of Renku using the flag 
+mentioned below or install the csi-rclone chart manually and set the correct storage class in the values for the 
+notebooks service.
+
+* NEW `noteboks.cloudstorage.enabled` - set to `true` to enable mounting cloud storage in sessions.
+* DELETE `notebooks.cloudstorage.s3.enabed` - superseeded by previous porperty.
```
rokroskar, line 16:
```suggestion
* DELETE `notebooks.cloudstorage.s3.enabed` - superseeded by previous property.
```

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,26 @@ For changes that require manual steps other than changing values, please check o
...
+* NEW `notebooks.cloudstorage.storageClass` - the storage class for the CSI Rclone chart, needed for new cloudstorage 
+  to work. The default `csi-rclone` should be fine unless already in use.
+* NEW `global.csi-rclone.install` - if `true` installs the csi-rclone chart alongside Renku. The chart is needed for 
+  cloud storage in sessions to work.
+* NEW `csi-rclone.storageClassName` - the storage class name the CSI drivers uses, should match what is configured in 
+  the `storageClass` property mentioned above.
+* NEW `csi-rclone.csiNodePlugin.tolerations` - Tolerations for the node plugin part of the CSI driver. Need to be set
```
rokroskar, line 23: is this the default or does it need to be set explicitly?

rokroskar (changes requested):
Thanks @Panaetius !

helm-chart/values.yaml.changelog.md
Panaetius: the storage class default is fine. the tolerations don't have a default, I was under the impression that's something specific to us?

rokroskar: Probably true - I was just wondering if this would potentially break some vanilla deployment of Renku but that's probably not the case.

rokroskar (approved):

aledegano (approved):

PR #3451, opened 2024-01-29
build: make configuration of privacy policy more flexible

ciyer:
Related to https://github.com/SwissDataScienceCenter/renku-ui/issues/2954

/deploy #persist renku-ui=main extra-values=ui.client.privacy.page.enabled=true

### DIFF
--- a/helm-chart/renku/templates/ui/ui-client-deployment-template.yaml
+++ b/helm-chart/renku/templates/ui/ui-client-deployment-template.yaml
@@ -25,14 +25,17 @@ spec:
         app: ui
         release: {{ .Release.Name }}
     spec:
-      {{- if and .Values.ui.client.privacy.enabled .Values.ui.client.privacy.page.enabled }}
+      {{- if .Values.ui.client.privacy.page.enabled }}
       volumes:
         - name: privacy
           configMap:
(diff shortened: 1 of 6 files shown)


helm-chart/renku/templates/ui/ui-client-deployment-template.yaml
```diff
@@ -25,14 +25,17 @@ spec:
         app: ui
         release: {{ .Release.Name }}
     spec:
-      {{- if and .Values.ui.client.privacy.enabled .Values.ui.client.privacy.page.enabled }}
+      {{- if .Values.ui.client.privacy.page.enabled }}
       volumes:
         - name: privacy
           configMap:
             name: {{ .Values.ui.client.privacy.page.configMapName | default (printf "%s-privacy-sample" (include "renku.fullname" .)) | quote }}
```
leafty, line 32: This does not match with the config map! It uses `{{ template "ui.fullname" . }}-privacy-sample` for the default one.

ciyer: Yes, I fixed that this morning.

ciyer:
# Testing

- Review the documentation for configuring the privacy policy and terms of service and check that it makes sense and the keys used for configuration are correct.
- Check that the keycloak login theme informs about the terms of service and privacy policy, and that the links resolve to the right location

image


image

PR #3474, opened 2024-02-07
[ShapeUp] Renku 1.0 Sessions

leafty:
Feature branch for the Renku 1.0 Sessions.


helm-chart/renku/templates/secrets.yaml
```diff
@@ -1,4 +1,11 @@
 ---
+{{- $data_service_kc_client_secret := randAlphaNum 64 | b64enc | quote }}
+
+{{- $renku_secret := lookup "v1" "Secret" .Release.Namespace (include "renku.fullname" .) }}
+{{- if $renku_secret }}
```
leafty, line 5: trying to fix new deployments:
```suggestion
{{- if and ($renku_secret) ($renku_secret.data) }}
```

PR #3480, opened 2024-02-07
feat: make configuration of privacy policy more flexible

rokroskar:
Related to https://github.com/SwissDataScienceCenter/renku-ui/issues/2954

/deploy #persist renku-ui=main extra-values=ui.client.privacy.page.enabled=true

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -696,12 +696,13 @@ ui:
     # privacy.page.configMapName value. As a reference, you can use the sample configMap generated when
     # enabling the feature.
     privacy:
-      enabled: false
       page:
         enabled: false
         #configMapName: privacy-page
-        #configMapKey: privacy_statement
+        #configMapPolicyKey: privacy_statement
+        #configMapTermsKey: terms
       banner:
+        enabled: false
         content: |
           This website requires cookies in order to ensure basic functionality. By clicking
           or navigating the site, you consent to the use of cookies in accordance with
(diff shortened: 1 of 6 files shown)


leafty (commented):
Chart changes look good :+1:

Should the documentation also contain the procedure to enable the "Terms and Conditions" required action on Keycloak?

Note: waiting for the KC theme image to be ready before greenlighting.

helm-chart/renku/values.yaml
```diff
@@ -295,7 +295,7 @@ keycloakx:
     enabled: false
   extraInitContainers: |
     - name: theme-provider
-      image: renku/keycloak-theme:4.1.3
+      image: ciyerr/renku-theme:latest
```
leafty, line 298:
```suggestion
      image: renku/keycloak-theme:4.1.5
```

leafty (approved):

rokroskar:
@ciyer now I'm confused - I thought we decided against requiring Terms&Conditions approval on login? In favor of just informing?

ciyer:
> @ciyer now I'm confused - I thought we decided against requiring Terms&Conditions approval on login? In favor of just informing?

This is correct, and this is what the PR does. I think @leafty changed the configuration in the Keycloak admin to see how much effort it is to require approval (not much, just flip two switches).

leafty:
> Got it - can we just have one sentence somewhere pointing to keycloak docs?

We can't, our Keycloak is too out-of-date :see_no_evil: https://www.keycloak.org/documentation-archive.html

We could merge as is, and add that documentation later; I don't think it's worth delaying things here.

rokroskar:
Ok lets leave as-is for now. It's 99% likely a non-issue.

PR #3512, opened 2024-02-22
feat: enable NetPol for Solr.

aledegano:
This change to the values files of Solr enables a network policy to Solr such that only `search-api` and `search-provision` are allowed to connect to it.
/deploy

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -498,7 +498,17 @@ solr:
     - renku-search
   javaMem: "-Xmx512M"
   networkPolicy:
-    enabled: false # Awaiting instructions from: https://github.com/bitnami/charts/issues/23424
+    allowExternal: false
+    extraIngress:
+      - ports:
+          - port: 8983
+        from:
+          - podSelector:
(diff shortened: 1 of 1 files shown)


jachro (approved):

lokijuhy (approved):
I hearby approve the addition of a blank line in the changelog

rokroskar:
> I hearby approve the addition of a blank line in the changelog

you _could_ also consider requesting a change to the changelog... ;)

aledegano:
There is an issue with this change: enabling the NetPol breaks the probes, so Solr keeps restarting (even though from the logs it seems to be fine...).
I'm investigating further.

rokroskar:
@aledegano this passes now - I nuked the namespace and redeployed.

aledegano:
> @aledegano this passes now - I nuked the namespace and redeployed. The deployment was borked and there was an issue with the Solr credentials.

Thank you, yes it works now. I also verified that `search-api` and `search-provision` are indeed the only 2 pods that are allowed to connect to Solr.

I am merging this now then.

PR #3539, opened 2024-03-11
chore: Remove `ui.client.url` from values as it is unused.

aledegano:
The value is not used anymore and we should remove it to avoid confusion.

/deploy


rokroskar:
cc @SwissDataScienceCenter/renku-ui-maintainers

leafty (approved):

PR #3547, opened 2024-03-21
Prometheus metrics for data service

Panaetius:
/deploy renku-data-services=prometheus-metrics renku-gateway=master


rokroskar:
Is there a dashboard? If yes, could you add a link in the PR description?

PR #3552, opened 2024-03-26
doc: improve the custom CA configuration part

sgaist:
The current version shows the configuration at the top level of the yaml while it should be under the global scope.

More details about the global scope in the [helm documentation](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/).

### DIFF
--- a/docs/how-to-guides/admin/prerequisites/certificates.rst
+++ b/docs/how-to-guides/admin/prerequisites/certificates.rst
@@ -48,16 +48,17 @@ configuration values:
 
 .. code-block:: yaml
 
+  global:
+    certificates:
+      customCAs:
+        - secret: renku-tls # must match the secretName field of the CA Certificate object.
+
   ingress:
     enabled: true
     annotations:
       cert-manager.io/cluster-issuer: null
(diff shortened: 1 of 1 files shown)


docs/how-to-guides/admin/prerequisites/certificates.rst
```diff
@@ -47,6 +47,10 @@ You can then use it for your deployment by setting annotations such as the follo
 configuration values:
 
 .. code-block:: yaml
+  global:
```
rokroskar, line 50: I believe you need a blank line here

rokroskar (changes requested):
Thanks @sgaist !! There is one formatting error and the tests complain. Note that I changed the base branch to be the branch for the next release, so you will need to do a rebase here.

docs/how-to-guides/admin/prerequisites/certificates.rst
sgaist: Indeed !

aledegano (approved):
Thank you for this fix to the docs!

rokroskar (approved):

PR #3557, opened 2024-03-27
chore: Upgrade search service to 0.0.38

eikek:
/deploy


rokroskar (changes requested):
can you please just quickly add a note about this version to the changelog?

rokroskar (approved):

eikek:
> can you please just quickly add a note about this version to the changelog?

I fixed the version in the changelog. I'm not very sure what to add, tbh. Currently there is this generic line about the new service being added. But no details are mentioned. Should it be more detailed? This bump from 0.0.37 -> 0.0.38 only tweaks the solr ingestion a bit to have some fields indexed case-insensitive. I can add this to the changelog if that is desired, it only seems to be the only detail then?

rokroskar:
it's fine imho, the detail is in the release text for the component. Thanks!!

PR #3570, opened 2024-04-03
chore: set bitnami-common version in the parent chart

rokroskar:
`bitnami/common` is used in several places and can cause version conflicts.

### DIFF
--- a/helm-chart/renku/requirements.yaml
+++ b/helm-chart/renku/requirements.yaml
@@ -36,6 +36,7 @@ dependencies:
     repository: "oci://registry-1.docker.io/bitnamicharts"
     version: "8.9.2"
     condition: solr.enabled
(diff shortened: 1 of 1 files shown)


helm-chart/renku/requirements.yaml
```diff
@@ -36,3 +36,8 @@ dependencies:
     repository: "oci://registry-1.docker.io/bitnamicharts"
     version: "8.9.2"
     condition: solr.enabled
+  - name: common
```
olevski, line 39: Just a reminder to add a comment here to explain why this is pinned here.

rokroskar: done

Panaetius (approved):

olevski (approved):

PR #3574, opened 2024-04-04
chore: Mute selenium dependency updates

eikek:
Ignores automatic updates to selenium-java. I think it should go into master, since this is the branch updates are calculated from (as it is still the default branch)


rokroskar:
this will make us forget about the selenium updates eventually - is that something we're ok with?

eikek:
As far as I understood, we want to get rid of it anyways as everything will be in the cypress tests now? We already removed some test cases.

rokroskar (approved):

Issue #3555, opened 2024-03-27
docs: add instructions solr/search


rokroskar:
We need to add documentation about how to bootstrap the solr chart, esp how to set the password. The new search values should also be documented in the values changelog.

PR #3576, opened 2024-04-05
Add lines about new search and solr components

eikek:
This is my attempt to provide some notes on the search and solr components,  addressing #3555


rokroskar (approved):
Thanks!!

PR #3579, opened 2024-04-09
chore: run cypress in parallel

olevski:


cypress-tests/cypress/support/commands/general.ts
```diff
@@ -24,6 +24,10 @@ function getDataCy(value: string, exist?: boolean) {
   return cy.get(`[data-cy=${value}]`);
 }
 
+export function getRandomString(length: number=6) {
+  return Math.random().toString(20).substr(2, length)
+}
```
lorenzo-cavazzi, line 29: Just nitpick details: no need to specify the type when it's automatically inferred by the default value. Also, `.substr` 
is deprecated ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr)), better using `.substring`.

```suggestion
export function getRandomString(length = 6) {
  return Math.random().toString(20).substring(2, length)
}
```

lorenzo-cavazzi: You could also set the default to `8` and avoid passing the argument in the other files since you always use 8 -- this is ofc totally discretionary, feel free to keep it as is :smile:

rokroskar:
< 10 minutes 🤯 🚀 🌔

lorenzo-cavazzi (approved):

PR #3586, opened 2024-04-12
fix: parallel cypress tests

olevski:
I think the reason for the failures when things run in parallel is that the tests do not wait sufficiently long for the redirects of the login process to finish. And they try to visit a page before the user has fully logged in.

Hopefully the check of the url after the login process finishes is enough to remedy this. 

/deploy

The changes that helped are:
- Adding a wait right after the login completes, for some reason the main problem was that after the login was complete calling gitlab still randomly resulted in 401s half the time. But after waiting for 10s Gitlab does not respond with 401s.
- Removing the login from the before hook and adding it to every beforeEach hook. I did the same with the project creation. This is better because beforeEach hooks are retried where as a single failure in the before hooks cannot be retried and results in the test failing.

### DIFF
--- a/cypress-tests/cypress/support/commands/general.ts
+++ b/cypress-tests/cypress/support/commands/general.ts
@@ -3,7 +3,7 @@ import { TIMEOUTS } from "../../../config";
 export const validateLogin = {
   validate() {
     // This returns 401 when not properly logged in
-    cy.request("/ui-server/api/user");
+    cy.request("/ui-server/api/data/user");
   },
 };
 
--- a/cypress-tests/cypress/support/commands/general.ts
+++ b/cypress-tests/cypress/support/commands/general.ts
@@ -2,6 +2,9 @@ import { TIMEOUTS } from "../../../config";
 
 export const validateLogin = {
   validate() {
+    // If we send a request to the user endpoint on Gitlab too quickly after we log in then
+    // it sometimes randomly responds with 401 and sometimes with 200 (as expected). This wait period seems to
+    // allow Gitlab to "settle" after the login and properly recognize the token and respond with 200.
     cy.wait(10000);
     // This returns 401 when not properly logged in
     cy.request("ui-server/api/data/user").its("status").should("eq", 200);
--- a/cypress-tests/cypress/support/commands/general.ts
+++ b/cypress-tests/cypress/support/commands/general.ts
@@ -2,8 +2,16 @@ import { TIMEOUTS } from "../../../config";
 
 export const validateLogin = {
   validate() {
+    cy.wait(10000);
     // This returns 401 when not properly logged in
-    cy.request("/ui-server/api/data/user");
+    cy.request("ui-server/api/data/user").its("status").should("eq", 200);
+    // This is how the ui decides the user is logged in
+    cy.request("ui-server/api/user").then((response) => {
+      expect(response.status).to.eq(200);
+      expect(response.body).property("username").to.not.be.empty;
+      expect(response.body).property("username").to.not.be.null;
+      expect(response.body).property("state").to.equal("active");
+    });
   },
 };
 
--- a/cypress-tests/cypress/support/commands/login.ts
+++ b/cypress-tests/cypress/support/commands/login.ts
@@ -99,8 +99,14 @@ function registerAndVerify(props: RegisterAndVerifyProps) {
   })
   cy.get("header").should("be.visible");
   cy.get("footer").should("be.visible");
-  cy.wait(1000);
+  cy.wait(10000);
   cy.request("ui-server/api/data/user").its("status").should("eq", 200);
+  cy.request("ui-server/api/user").then((response) => {
+    expect(response.status).to.eq(200);
+    expect(response.body).property("username").to.not.be.empty;
+    expect(response.body).property("username").to.not.be.null;
+    expect(response.body).property("state").to.equal("active");
+  });
 }
 
 type RobustLoginProps = {
--- a/cypress-tests/cypress/support/commands/login.ts
+++ b/cypress-tests/cypress/support/commands/login.ts
@@ -1,9 +1,9 @@
 const renkuLogin = (credentials: { username: string; password: string }[]) => {
-  for (const { username, password } of credentials) {
-    cy.get("#username").type(username);
-    cy.get("#password").type(password, { log: false });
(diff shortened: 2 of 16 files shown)


olevski:
This needs a new renku-actions release.

cypress-tests/cypress/support/commands/general.ts
```diff
@@ -2,8 +2,16 @@ import { TIMEOUTS } from "../../../config";
 
 export const validateLogin = {
   validate() {
+    cy.wait(10000);
```
lorenzo-cavazzi, line 5: Waiting for a specific time [is strongly discouraged by the guidelines](https://docs.cypress.io/api/commands/wait#Time) -- I'll look into turning on the linter in these tests during the cooldown.

Unless it's a very short time span, like <=1s, it would be much better to wait for something to show up, or a response from an API, or any other event to happen.
Could you give a look into that? If it takes too long, please add a comment in the code about why we need to wait 10s here.

cypress-tests/cypress/support/commands/login.ts
```diff
@@ -28,7 +28,7 @@ const register = (
 
   // ? wait to be assess whether tokens were refreshed automatically or we really need to register
   cy.wait(1000); // eslint-disable-line cypress/no-unnecessary-waiting
-  cy.request({ failOnStatusCode: false, url: "ui-server/api/user" }).then(
+  cy.request({ failOnStatusCode: false, url: "ui-server/api/data/user" }).then(
```
lorenzo-cavazzi, line 31: This points to the data service instead of the gateway. I would expect to authenticate with the gateway and check there if we are good to go. Has the authentication flow changed?

cypress-tests/cypress/support/commands/login.ts
```diff
@@ -91,21 +91,22 @@ function registerAndVerify(props: RegisterAndVerifyProps) {
...
-  cy.request("ui-server/api/user").its("status").should("eq", 200);
+  cy.location().should((loc) => {
+    const baseURL = new URL(Cypress.config("baseUrl"));
+    expect(["/", ""]).to.include(loc.pathname);
+    expect(loc.search).to.eq("");
+    expect(loc.hostname).to.eq(baseURL.hostname);
+  })
+  cy.get("header").should("be.visible");
+  cy.get("footer").should("be.visible");
+  cy.wait(10000);
```
lorenzo-cavazzi, line 102: Same as above

cypress-tests/cypress/support/commands/projects.ts
```diff
@@ -69,39 +69,46 @@ interface NewProjectProps extends ProjectIdentifier {
...
-    "contain",
-    newProjectProps.name.toLowerCase()
-  );
-  cy.get(`[data-cy="header-project"]`, { timeout: TIMEOUTS.vlong }).should(
-    "be.visible"
-  );
-  cy.get("ul.nav-pills-underline").should("be.visible");
+  const namespace = newProjectProps.namespace ?? Cypress.env("TEST_USERNAME");
+  const slug = encodeURIComponent(`${namespace}/${newProjectProps.name}`);
+  cy.request({failOnStatusCode: false, method: "GET", url: `/ui-server/api/projects/${slug}`}).then((response) => {
```
lorenzo-cavazzi, line 74: I feel this change is unnecessary since we can re-run single file specs but not single blocks: no need to recheck the project every time.
Do you mind renaming the function to something more descriptive like `createProjectWhenMissing` or `checkAndCreateProject`? I added a couple of comments to the spec files review before realizing what it actually does :grin:

cypress-tests/cypress/support/commands/projects.ts
```diff
@@ -69,39 +69,46 @@ interface NewProjectProps extends ProjectIdentifier {
 }
 
 function createProject(newProjectProps: NewProjectProps) {
```
leafty, line 71: This is now a misnomer. Use something like `getOrCreateProject()`

cypress-tests/cypress/support/commands/projects.ts
leafty: @lorenzo-cavazzi this is the entire point of this PR. It's to move the project creation into `beforeEach` blocks so that the project creation hook can be re-attempted if it fails.

cypress-tests/cypress/support/commands/projects.ts
olevski: Sure I will rename it.

cypress-tests/cypress/support/commands/projects.ts
olevski: Exactly what @leafty said.

cypress-tests/cypress/support/commands/login.ts
olevski: The gateway has no endpoint to show user info or if a user is logged in or not. That request goes through the gateway where the credentials are added and then to the data service which has a database of all users and can return user information if presented with a valid token.

I dont think it really matters tbh where the request is served from. Going forward that will be the way to check if the user is logged in.

So I added this mostly for the future. Currenly the UI decides if the user is logged in by hitting Gitlab via the gateway. And that is the check I added right below.

cypress-tests/cypress/support/commands/general.ts
olevski: Lorenzo we already have waits in other places... I.e. to wait for metadata indexing or when we call visitAndLoadProject to wait for things to render fully. 

I did not want to add more but there is no way I could think of to address this. The problem is that hitting the user endpoint on gitlab (via the gateway) right after completing the log in sometimes resulted in a 401 and sometimes in 200 (as expected). I think the problem is that our tests are just too fast for Gitlab or Gitlab is getting overwhelmed and fails to recognize or issue the tokens. It could also be the gateway but I looked through the logs of failed tests and did not see any errors or issues. Adding the wait completely removed this problem and Gitlab more consistently was able to log the user in and recognize the credentials.

cypress-tests/cypress/support/commands/projects.ts
olevski: I went for `createProjectIfMissing`.

olevski:
@lorenzo-cavazzi as per our call I added comments about why the `cy.wait` is necessary.

lorenzo-cavazzi (approved):

PR #3591, opened 2024-04-17
feat: add secrets to sessions

lorenzo-cavazzi:
Feature branch for the build "Secrets in sessions"

https://www.notion.so/Session-Secrets-Build-Team-c6d62656018841109dfe52d831b97f9f

/deploy renku-ui=lorenzo/shape-up-secrets-in-sessions renku-data-services=main renku-gateway=limit-user-secret renku-notebooks=secrets-sidecar extra-values=secretsStorage.encryptionKey=vMqdVXcSIwjwONvGiNE2vjzRjmi8NJ8Y9lTB8m3l/Xg=


helm-chart/renku/templates/setup-job-platform-init.yaml
```diff
@@ -0,0 +1,96 @@
...
+          image: "{{ .Values.platformInit.image.repository }}:{{ .Values.platformInit.image.tag }}"
+          args: [ "platform-init.py" ]
+          securityContext:
+            runAsUser: 1000
+            runAsGroup: 1000
+            allowPrivilegeEscalation: false
+            runAsNonRoot: true
+          env:
+            - name: K8S_NAMESPACE
+              value: {{ .Release.Namespace}}
```
olevski, line 34:
```suggestion
              value: {{ .Release.Namespace }}
```

helm-chart/renku/templates/setup-job-platform-init.yaml
```diff
@@ -0,0 +1,96 @@
...
+            - name: RENKU_FULLNAME
+              value: {{ template "renku.fullname" . }}
+            - name: PLATFORM_INIT_CONFIG
+              value: {{ .Values.global.platformConfig| default (printf "{}") | quote }}
```
olevski, line 38:
```suggestion
              value: {{ .Values.global.platformConfig | default (printf "{}") | quote }}
```

helm-chart/renku/values.yaml
```diff
@@ -5,6 +5,8 @@
 ## Global variables
 ## Shared values/secrets
 global:
+  platformConfig: |
+    {}
```
olevski, line 9: I see that the `platform-init` script uses json to parse this. Lets switch to yaml. It is just nicer to write it and it is nicer to have yaml in a file that already has a bunch of yaml like the values file.

scripts/platform-init/platform-init.py
```diff
@@ -0,0 +1,202 @@
...
+@dataclass
+class Config:
+    k8s_namespace: str
+    renku_fullname: str
+    secret_service_public_key: str | None
+    secret_service_private_key: str | None = field(repr=False)
+
+    @classmethod
+    def from_env(cls):
+        config_map = json.loads(os.environ.get("PLATFORM_INIT_CONFIG", "{}"))
```
olevski, line 21: See the other comment in the values file. It is nicer to use YAML. Otherwise we have YAML for most of the values file and then a huge section with json.

And when we switch to yaml people can still use json if they want to.

helm-chart/renku/values.yaml
```diff
@@ -5,6 +5,8 @@
 ## Global variables
 ## Shared values/secrets
 global:
+  platformConfig: |
```
olevski, line 8:
```suggestion
  ## YAML string that contains all application-level Renku configuration options.
  platformConfig: |
```

scripts/platform-init/platform-init.py
```diff
@@ -0,0 +1,202 @@
...
+                },
+                type="Opaque",
+            ),
+        )
+
+    elif (
+        existing_private_key != config.secret_service_private_key
+        and config.secret_service_private_key is not None
+    ):
+        # update private key
```
olevski, line 112: Can we actually support updating the private key? What happens to all the secrets that are sitting in k8s secrets or in the database that are encrypted with the old key?

Same question for the public key.

olevski: Also I think we should support two cases only:
1. Both private and public key are specified in the config and they are proper and valid
2. Neither public nor private key are specified in the config, in this case we generate the secret only if the secret does not already exist

Right now we are independently checking for the pivate and public keys and also generating them independently. I think this is too complicated and I dont think anyone would or should be using this option. I.e. provide a public key but let renku generate the private one.

Lastly there is no documentation on what algorithm/standard should be used for the keys. And I assume this matters or it does not? Like RSA, eliptic curve, etc...

helm-chart/renku/templates/secrets-storage/pdb.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: PodDisruptionBudget
+metadata:
+  name: {{ template "renku.fullname" . }}-secrets-storage
+  labels:
+    app: renku-secrets-storage
+    chart: {{ template "renku.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  maxUnavailable: 50%
```
olevski, line 12: This may be like this in other places. But `maxUnavailable` will round up. So if you have 1 pod you will get fully disrupted. We should switch to `minAvailable` so that if we have 1 pod we will not get fully disrupted.

```suggestion
  minAvailable: 50%
```

helm-chart/renku/templates/secrets-storage/network-policy.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      app: {{ template "renku.fullname" . }}-secrets-storage
+      release: {{ .Release.Name }}
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+        - podSelector:
```
olevski, line 14:
```suggestion
        # Allow ingress to the secret service only from notebook service
        - podSelector:
```

helm-chart/renku/templates/secrets-storage/secret.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "renku.fullname" . }}-secrets-storage
```
olevski, line 4: What is the difference between this secret and the one generated by the init script? Why dont we generate all of them with the init script?

olevski (changes requested):
A few questions and minor changes.

docs/topic-guides/secrets/secrets.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Secrets in RenkuLab
+===================
+
+This is still a placeholder.
+
+Add or change secrets
+---------------------
+
+WIP
```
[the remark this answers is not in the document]
Panaetius: we already have a PR for that. but I can remove it here

scripts/platform-init/platform-init.py
Panaetius: rotating keys is a nice to have that still needs to be handled.

I guess we could also just not require a public key at all and always generate it.

helm-chart/renku/templates/secrets-storage/secret.yaml
Panaetius: this one is for symmetric encryption at rest in the DB. but yes, it should also be in platform init, it's a leftover from the first iteration of the service.

scripts/platform-init/platform-init.py
olevski: > I guess we could also just not require a public key at all and always generate it.

I think that is fine

> rotating keys is a nice to have that still needs to be handled.

For this you would have to have some kind of watch in the secret service to restart it when the secret changes. I dont know how easy it is to implement this though. An easier way is to take the sha256 hash of the secret and make it an environment variable in the secret service. Then when the secret changes the secret service will automatically update. See https://helm.sh/docs/chart_template_guide/function_list/#sha256sum

olevski: And taking the sha256 digest still does not address the issue of how you decrypt already encrypted secrets that may be still lying around. So perhaps it is better to keep this simple in the first implementation and improve this later during cooldown.

olevski (approved):

Panaetius (approved):

helm-chart/renku/templates/data-service/deployment.yaml
```diff
@@ -142,5 +152,17 @@ spec:
         - name: server-options
           configMap:
             name: {{ template "renku.fullname" . }}-server-options
+        - name: encryption-key
+          secret:
+            secretName: {{ template "renku.fullname" . }}-secrets-storage
```
leafty, line 157: Where is this defined? There does not seem to be any new definition for a secret with this name.

PR #3596, opened 2024-04-23
chore: update renku actions 1.11.1

olevski:
This should make the cypress tests run more reliably.

/deploy


olevski:
Replaced by #3586

PR #3607, opened 2024-05-03
fix: cloning repositories in sessions

olevski:
Steven from the academic team encountered this issue.

With the update of the notebook service to handle multiple repos the path where the project is cloned was set to a different parameter from gitlab which seems to ignore the case of the letter in the name. But a lot of other parameters - most important of which is the working directory for the session - was set to the case sensitive value from gitlab.

This resulted in projects which have a path with upper and lowercase letters to be cloned in the wrong directory.

For more info see the PR in the notebook service.

/deploy

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,26 @@
 .. _changelog:
 
+0.51.1
+------
+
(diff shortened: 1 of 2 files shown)


aledegano (approved):
LGTM!

Panaetius (approved):

lokijuhy (approved):

PR #3625, opened 2024-05-16
chore: add check to prevent merging prs with custom deploy command

Panaetius:
Adds a new check to make sure that the \deploy string doesn't contain a custom component version, when targetting a `release-*` branch.

e.g. `\deploy extra-values=a=10` is ok, but `\deploy renku-notebooks=my-pr` is not. 

This is to ensure that PRs aren't merged when they still depend on custom settings, as this can break other PRs that don't have the correct deploy string set.


rokroskar (approved):
Thanks @Panaetius - we should remember to add this check as required for `release-` branches

PR #3631, opened 2024-05-17
Update search services to 0.1.0

eikek:
/deploy


rokroskar:
This also includes using namespaces in search, correct?

eikek:
> This also includes using namespaces in search, correct?

Yes, namespaces are also supported. It requires messages to be in v2, though. Otherwise we don't get the namespace in the message and can't ingest it.

rokroskar:
👍 Can you please add a blurb in the user-facing part of the release notes :)

eikek:
> 👍 Can you please add a blurb in the user-facing part of the release notes :)

There already is from the previous version 😄 This update has no user facing changes, only improvements. The important changes are mentioned in CHANGELOG.rst. That's why I only updated the link to the new release. But I'm happy to elaborate more, only there is not much more to say I think.

rokroskar (approved):

PR #3640, opened 2024-05-27
feat: add trusted proxies configuration to data-service

leafty:
See: SwissDataScienceCenter/renku-data-services#225

/deploy


olevski:
And lastly please target the new release branch. `release-0.53.x`

leafty:
> And lastly please target the new release branch. `release-0.53.x`

Can we cleanup that branch first? The changes are crazy if I do the switch.

leafty:
> Also shouldn't the proxy count be 2? Because in most cases the ui will call this through the ui-server, which will then go through the gateway and finally reach the data service.

I think our network hops do not behave as proper reverse proxies should, so for now using `1` as the proxies count value makes the HTTP scheme detection work.

leafty:
> @leafty this requires changes on both the ui-server and the gateway so that this secret is used and set in incoming requests, correct? Or am I misunderstanding the examples/config from the sanic docs here: https://sanic.dev/en/guide/advanced/proxy-headers.md#getting-started

Yes, I removed the forwarded secret value. We can try to address this in the future.

Panaetius (approved):

PR #3641, opened 2024-05-27
chore: bump template versions

rokroskar:
Updates project template versions.

/deploy


olevski:
@rokroskar please set the target to `release-0.53.x`

rokroskar:
I think maybe I'd like to make it 0.52.1... so it can go out sooner rather than later. Or just make the change in the values file... because there are some important updates, I just forgot to get it merged before 🙈

PR #3650, opened 2024-05-30
docs: update release file

lorenzo-cavazzi:
Now that we use ShapeUp to organize most of the work within the Renku Team, we might want to expand the guidelines for releasing to optimize the process and prevent conflicts between different features.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -5,29 +5,53 @@ There are two types of Renku releases: planned and unplanned.
 A release consists of: 
 
 * Helm chart changes to reflect the new versions of individual components
+* Updates to the documentation and the integration tests (both `cypress` and `acceptance`)
 * Changes to `CHANGELOG.rst`
 * A new tag in this repository, which results in packaging and pushing out new helm charts and corresponding docker images
 
 This procedure should be followed for *any* release:
 
-* Create a release branch (e.g. `0.46.x`), if one does not already exist, with the [release action](https://github.com/SwissDataScienceCenter/renku/actions/workflows/create-release-branch.yml).
-* Create a `CHANGELOG` entry for the release and open a PR; create a deployment, this is the reference for the release.
-* Note that any PR that should go into the release needs to target the release branch _not_ `master`.
-* All release branches should be protected.
+* Create a release branch (e.g. `0.46.x`), if one does not already exist, with the [release action](https://github.com/SwissDataScienceCenter/renku/actions/workflows/create-release-branch.yml). Mind that release branches are protected.
+* Merge new PRs into the release branch. Neither bugfixes nor features should be merged directly into `master`.
+  * Mind that the PRs pointing to a release branch cannot be merged if they break the platform, even temporarily. Be sure all tests pass by using the `/deploy` string without referencing other components or setting extra values; those should be included in the helm chart folder.
+  * There must be no PRs pointing to a release branch when releasing. If a change won't make it to that release, please point it to the following one (or `master` until it targets another release branch).
+* Be sure `CHANGELOG.rst` is up-to-date and reflects the changes in the release branch before tagging a release. This might require a final cleanup PR.
 
-Acceptance tests have to pass on all release branches before merging. 
+Acceptance tests have to pass on all release branches before merging.
 
 ## Planned release
 
-A release is planned either at the beginning of a build cycle or at the start of the cooldown period. 
-Depending on the features being developed, there may be more than one release per build cycle, especially 
-if a feature requires significant infrastructure changes. 
+A release is planned at the end of a build cycle and the end of the cooldown period. 
+Depending on the features being developed, there may be more than one release per build cycle, especially if a feature requires significant infrastructure changes.
 
-For larger features that require the coordination of several teams, consider using 
-a feature branch to which the individual component updates can be merged. This feature
-should then get merged into the release branch once the feature is complete. 
+Since we usually include multiple features in the same release, please use feature branches to merge all the changes related to a single feature. Ideally, the commits should be squashed together when merging into the release branch, or reduced to a minimum (you can squash _some_ commits together in the feature branch before merging into the release branch).
 
-Note that there might be multiple planned release branches and PRs active at the same time. 
+### Merge feature branches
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -5,29 +5,53 @@ There are two types of Renku releases: planned and unplanned.
...
+### Merge feature branches
+
+Depending on the specific feature, it might be necessary to include changes from multiple components. Please coordinate with the other teams to ensure that all the necessary changes are merged at the same time to avoid temporarily breaking the release branch and minimize the impact on other teams.
+
+Ideally, you should include all the necessary documentation and integration test changes in the same PR.
+
+Before merging a feature branch, please always drop a message on the `#dev-core` channel to inform the other teams about the upcoming changes. This will allow them to plan their work accordingly; even if you don't expect any conflicts, it's always better to be safe than sorry, especially toward the end of the build sprint when teams need to wrap up their work and must point their feature branches to the release branch.
+
+```mermaid
+flowchart LR
```
Panaetius, line 38:
```suggestion
flowchart TD

```
might be more readable?

RELEASE.md
```diff
@@ -5,29 +5,53 @@ There are two types of Renku releases: planned and unplanned.
...
-* Create a `CHANGELOG` entry for the release and open a PR; create a deployment, this is the reference for the release.
-* Note that any PR that should go into the release needs to target the release branch _not_ `master`.
-* All release branches should be protected.
+* Create a release branch (e.g. `0.46.x`), if one does not already exist, with the [release action](https://github.com/SwissDataScienceCenter/renku/actions/workflows/create-release-branch.yml). Mind that release branches are protected.
+* Merge new PRs into the release branch. Neither bugfixes nor features should be merged directly into `master`.
+  * Mind that the PRs pointing to a release branch cannot be merged if they break the platform, even temporarily. Be sure all tests pass by using the `/deploy` string without referencing other components or setting extra values; those should be included in the helm chart folder.
+  * There must be no PRs pointing to a release branch when releasing. If a change won't make it to that release, please point it to the following one (or `master` until it targets another release branch).
```
Panaetius, line 17: It took me a bit to understand that the first point refers to merging into the release branch but the second one is about merging the release branch into master, I think because it's nested under a point that talks about merging PRs into the release branch. Because the release itself is also a PR, just talking about release branches and feature branches and using that wording consistently makes it easier to understand? And if we can keep it short there's a higher chance of people actually reading it.

I'm thinking like
```
* Create a release branch (e.g. `0.46.x`), if one does not already exist, with the [release action](https://github.com/SwissDataScienceCenter/renku/actions/workflows/create-release-branch.yml). # I don't think we need to mention that they are protected, it's just a fact of life
* create feature branches off of the release branch for new features.
  * a feature branch must contain all the changes needed for that feature.
* squash merge feature branches into the release branch.
  * do not merge feature branches if they break the platform, even temporarily.
  * make sure all tests pass with the `/deploy` action.
  * make sure the `/deploy` action does not reference specific component versions, these should be set in the chart.
  * make sure the `CHANGELOG.rst` is up to date and reflects the changes made.
* The release manager merges the release branch into master when planned
  * If a feature branch won't make it into a planned release, point it to the next release branch.
  * This might need a final cleanup PR to consolidate the changelog
```

RELEASE.md
```diff
@@ -5,29 +5,53 @@ There are two types of Renku releases: planned and unplanned.
...
-Acceptance tests have to pass on all release branches before merging. 
+Acceptance tests have to pass on all release branches before merging.
 
 ## Planned release
 
-A release is planned either at the beginning of a build cycle or at the start of the cooldown period. 
-Depending on the features being developed, there may be more than one release per build cycle, especially 
-if a feature requires significant infrastructure changes. 
+A release is planned at the end of a build cycle and the end of the cooldown period. 
+Depending on the features being developed, there may be more than one release per build cycle, especially if a feature requires significant infrastructure changes.
```
Panaetius, line 25:
```suggestion
A release is planned at the end of a build cycle and the end of the cooldown period. 
More releases can be done if needed, in which case you should coordinate with the release manager.
```

RELEASE.md
```diff
@@ -5,29 +5,53 @@ There are two types of Renku releases: planned and unplanned.
...
 
-For larger features that require the coordination of several teams, consider using 
-a feature branch to which the individual component updates can be merged. This feature
-should then get merged into the release branch once the feature is complete. 
+Since we usually include multiple features in the same release, please use feature branches to merge all the changes related to a single feature. Ideally, the commits should be squashed together when merging into the release branch, or reduced to a minimum (you can squash _some_ commits together in the feature branch before merging into the release branch).
```
Panaetius, line 27: I included this in my suggestion above so we could remove it here

RELEASE.md
lorenzo-cavazzi: Thanks! That reads much better.
I slightly changed the 3rd bullet point to make it more readable. Feel free to suggest further changes

Panaetius (approved):

Panaetius:
actually one thing not covered is that the thing with the highest risk of impacting other people is probably not merging feature branches into the release branch, but making releases of individual components.

I.e. a new version of notebooks needs a values change and breaks without the new values. TO make a feature branch for a release, we need to first make a release of notebooks to bump the version in the feature branch.
So there is at least like a day between merging to main on notebooks and merging to the release branch in the renku repo. Anyone else that is working on notebooks during this time, who sees that their PR is outdated and updates it, is in a "broken" state until the feature branch is merged. I.e. they now need custom values or base off of the unmerged feature branch. 

I don't think this is something we can prevent, other than also dropping a message when merging notebooks that you shouldn't rebase?

lorenzo-cavazzi:
@Panaetius I initially added a comment about that, but then I realized addressing the conflicts inside the single repos isn't trivial and it might be left to the frontend and backend teams.
Since we usually tag short before merging the feature branches, perhaps we can specify that the message should come _before_ tagging the single repos instead or just before merging the feature -- otherwise, we might get a lot of messages on slack when approaching the release date (I fear people would start ignoring them in that case)

PR #3653, opened 2024-06-03
feat: support for rotating secrets storage keys

Panaetius:
/deploy


scripts/platform-init/platform-init.py
```diff
@@ -86,20 +102,24 @@ def init_secret_service_secret(config: Config):
...
             ),
         )
-    elif existing_private_key is None and config.secret_service_private_key is not None:
+    elif config.secret_service_private_key is not None:
         # create private key from config
+        logging.info("Setting private key from values")
+        if existing_private_key:
+            # delete old private key
+            logging.info("deleting old private key secret")
+            v1.delete_namespaced_secret(private_key_secret, config.k8s_namespace)
```
olevski, line 111: suggestion(blocking): Just fail here. Because it is very likely someone is making a mistake in this case. And the key that is being deleted is still in use. The way I understand this branch of the logic is that there is an existing private key secret in k8s and also a secret has been set in the configuration but the previous secret field is not set. So in this case most likely there are secrets in the DB encrypted with the value in the k8s secret. In this case the only viable solution is to do a rotation - otherwise we may get in trouble.

scripts/platform-init/platform-init.py
```diff
@@ -86,20 +102,24 @@ def init_secret_service_secret(config: Config):
                 type="Opaque",
             ),
         )
-    elif existing_private_key is None and config.secret_service_private_key is not None:
+    elif config.secret_service_private_key is not None:
```
olevski, line 105: suggestion: We are on python 3.12 in this script. So I think this whole section of if/elif/else should be converted to match/case because it will be easier to follow what case/conditions are being met in each branch.

scripts/platform-init/platform-init.py
```diff
@@ -109,10 +129,22 @@ def init_secret_service_secret(config: Config):
             ),
         )
     else:
-        # just load key to create public key from
+        # just load key to create public key from and unset previous secret if set
+        logging.info("Leaving private key unchanged")
         private_key = serialization.load_pem_private_key(
             existing_private_key.encode(), password=None
         )
+        v1.patch_namespaced_secret(
```
olevski, line 137: question: My understanding is that in this branch of the logic nothing has changed, so why do we have to patch the secret? Isn't  `existing_private_key` just the result of reading the `private_key_secret`? So this patches the contents of the secret with the same thing that is already there.

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,14 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+## Upgrading to Renku 0.54.0
+
+* NEW ``global.platformConfig``: The YAML string can now contain a new key, `secretServicePreviousPrivateKey` which allows for rotating the secret-storage private key.
+  To rotate keys, set this to the previous `secretServicePrivateKey` value and set a new key for `secretServicePrivateKey`. Secrets-storage will then rotate all secrets
+  once its started. You can monitor the progress of the rotation in prometheus using the `secrets_rotation_count` (for total secrets rotated so far) and `secrets_rotation_state`
+  (either `running`, `finished` or `errored`) for the overall state of the rotation. Please make sure to unset `secretServicePreviousPrivateKey` once rotation is finished
+  as a matter of best practice.
```
olevski, line 14:
```suggestion
  as a matter of best practice.
  
  NOTE: Make sure that you do not redeploy or rollback the Renku Helm chart while a key rotation is underway. Even if the deployment is broken it is best to wait for the key rotation to finish before attempting another deployment or a rollback.
```

olevski (changes requested):

scripts/platform-init/platform-init.py
Panaetius: if we rotated secrets before, there's an old and new key in the secret, and this removes the old key entry (otherwise secrets storage would just try to rotate again, which would be a nop but still take cpu cycles/hit the DB)

scripts/platform-init/platform-init.py
Panaetius: This happens when you rotate the secret, you copy the key to previousKey and set a new key in key. I'll change it so it only does this if a previousKey was set.

scripts/platform-init/platform-init.py
Panaetius: Another solution would be to use two separate k8s secrets for the new and old key and delete the old one. but that's one of the annoying parts with the platform init script, since it happens after helm deploy, in the chart it's harder to conditionally mount secrets as at deploy we don't know if there will be a secret. We can read the config values from the values.yaml and then parse them and do some {{ if ... }} based on them, but then we have python code and helm template that have logic based on a value which isn't that clean. Just having one secret with variable properties that's always mounted is easier and then it's just handled in one place.

Panaetius:
I've refactored the code quite a bit to have more sanity checks, be easier to follow and cover all the edge cases.

olevski (approved):

PR #3690, opened 2024-06-21
fix: get cypress tests to go through the GitLab OAuth screen

leafty:
Fixes the cypress tests getting stuck at the login step.

/deploy


ciyer (approved):
Thanks!!

PR #3696, opened 2024-06-26
chore: do not repeat image tags in values.yaml

leafty:
/deploy

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -1328,7 +1328,7 @@ gateway:
   reverseProxy:
     image:
       repository: renku/renku-revproxy
-      tag: *&gateway_image_tag
+      tag: *gateway_image_tag
       pullPolicy: IfNotPresent
     metrics:
       enabled: true
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -672,7 +672,7 @@ ui:
     replicaCount: 1
     image:
       repository: renku/renku-ui
-      tag: "3.28.1"
+      tag: &ui_image_tag "3.28.1"
       pullPolicy: IfNotPresent
       ## Optionally specify an array of imagePullSecrets.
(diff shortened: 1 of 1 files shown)


aledegano (commented):
Honestly I really think it's not a good idea to use anchors in the values file (actually I wouldn't use them anywhere...).

I think it would be a much better approach if we could instead define a common value (e.g. `ui-images-tag`) and use only that instead.

leafty:
> Honestly I really think it's not a good idea to use anchors in the values file (actually I wouldn't use them anywhere...).
> 
> I think it would be a much better approach if we could instead define a common value (e.g. `ui-images-tag`) and use only that instead.

Any rationale for not using anchors and references? This PR improves the current situation where someone could forget to update e.g. `renku/secrets-mount` and break a release.

Changing the structure of the values and chart to define tags as globals is a much deeper change.

We should not let "better" be the enemy of "good".

aledegano:
> > Honestly I really think it's not a good idea to use anchors in the values file (actually I wouldn't use them anywhere...).
> > I think it would be a much better approach if we could instead define a common value (e.g. `ui-images-tag`) and use only that instead.
> 
> Any rationale for not using anchors and references? This PR improves the current situation where someone could forget to update e.g. `renku/secrets-mount` and break a release.
> 
> Changing the structure of the values and chart to define tags as globals is a much deeper change.
> 
> We should not let "better" be the enemy of "good".

You called this branch `dry-image-tags` but we are not "drying" anything here, just automating the repetitiveness. Which arguably makes the problem worse because the repetition is still there but it will be harder to catch once something goes wrong.

For instance it will not be clear at all to admins which value is the anchor and which are the anchored as it's not clearly stated.
Which in turns might lead to a very confusing situation when changing one value will change many other values (the anchor) but changing another value will only change it once (the anchored).

I am all for incremental improvement, but this is no improvement at all for the end user of the values file.

PR #3706, opened 2024-07-04
feat: Allow db common to be hardcoded

JGodin-C2C:
# What 
This allows the db common to be hardcoded in the value file. For instance, it will allow idempotence

# why 
This will solve a problem met when using, for example, sealed secret 

# Further thought

I am not sure if this is a good idea, or if one should remove the whole secret management when a secret is provided by the user and let the user manage the secret by himself, leading to a more flexible way of managing this particular secret.

### DIFF
--- a/helm-chart/renku/templates/renku-common-db-secret.yaml
+++ b/helm-chart/renku/templates/renku-common-db-secret.yaml
@@ -5,6 +5,9 @@
 {{- if $secret }}
 {{- $db_password = index $secret.data "password" }}
 {{- end -}}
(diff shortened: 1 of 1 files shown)


JGodin-C2C:
@aledegano I also doing some stuff here :blush:

JGodin-C2C:
@aledegano I am not working on the project anymore, but this may be of your interest ? 
Else, feel free to close this MR 

Des beks .

aledegano:
Thank you for your contribution.
I believe that more changes would be required, but currently they are not a priority.
I will close the PR.

PR #3741, opened 2024-08-23
tests: fix test to match new defaults for data sources

ciyer:
Fixes to the tests to match new defaults for data sources.

/deploy renku-ui=main


ciyer:
No longer relevant.

PR #3753, opened 2024-09-04
feat: upgrade csi-rclone to 0.3.0

olevski:
This upgrades the version of CSI rclone.

It is needed for handling encrypted secrets for storage and for the new amalthea operator.

Also it turns out we have missed to add a few versions of csi rclone that were already released to renku.


rokroskar (approved):

aledegano (commented):
AFAIR upgrading csi-rclone is a bit delicate: if there are sessions mounting volumes with it at the time of the upgrade they might get bricked.
Can you please put a clearly visible warning to admins?

olevski:
@aledegano I added a note for administrators in the changelog.

aledegano (approved):

rokroskar (approved):

PR #3758, opened 2024-09-05
(docs) redirect to Community portal for Renku 2.0 docs

lokijuhy:
We've noticed people interested in Renku 2.0 coming to our docs and doing our tutorial, which isn't a good representation of what Renku 2.0 is all about! This change directs those people to the [Community Portal](https://renku.notion.site/Renku-Community-Portal-2a154d7d30b24ab8a5968c60c2592d87) instead.


docs/tutorials.rst
```diff
@@ -3,6 +3,14 @@
...
 ---------
 
+.. note::
+
+     **We're building the next version of Renku!** If you're looking for a tutorial for Renku 2.0,
+     please see the Renku 2.0 documentation on our `Community Portal
+     <https://renku.notion.site/Renku-Community-Portal-2a154d7d30b24ab8a5968c60c2592d87>`_ instead.
+     To learn more about the big changes coming in Renku, check out our `blog post
+     <https://blog.renkulab.io/renku-2/>`_.
+
```
rokroskar, line 13: should we be more explicit here and say something like "The tutorial linked below is outdated and refers to the legacy version of Renku that we will no longer support." ?

lokijuhy: Good idea!

I feel like "no longer support" might scare current users more than it needs to. I chose "no longer under active development" instead.

rokroskar (approved):

PR #3767, opened 2024-09-13
fix: bump gateway to 1.0.2

leafty:
Fixes the `Authorization` header for the Renku 2.0 search service.

/deploy


eikek:
I can confirm that this fixes the issue, I was able to find my private project.

rokroskar (approved):

PR #3781, opened 2024-09-25
test: add dashboard v2 e2e tests

andre-code:
Add e2e tests for dasboard V2.

/deploy renku-ui=dashboard-renku2.0 renku-data-services=main


cypress-tests/cypress/support/commands/projectsV2.ts
```diff
@@ -0,0 +1,77 @@
...
+function getUserNamespaceAPIV2(): Cypress.Chainable<string | null> {
+  return cy.request({ failOnStatusCode: false, method: "GET", url: `api/data/namespaces?minimum_role=editor` })
+    .then((response) => {
+      if (response.status === 200) {
+        const userNamespace = response.body?.filter((namespace) => namespace.namespace_kind === "user");
+        return userNamespace && userNamespace.length > 0 ? userNamespace[0].slug : null;
+      }
+      return null;
+    });
+}
```
olevski, line 22: question: I am confused about what the purpose of this should be? If you want to get the user's own namespace then you should set the minimum_role to "owner". I dont think that user namespaces can have an editor role - they just have one role - owner and one member the user they are tied to. So this will probably work but I think it will be less confusing if the role is owner.

cypress-tests/cypress/support/commands/projectsV2.ts
```diff
@@ -0,0 +1,77 @@
...
+        "be.visible"
+      );
+    }
+  });
+}
+function deleteProjectFromAPIV2(id: ProjectIdentifier) {
+  cy.request({failOnStatusCode: false, method: "GET", url: `api/data/projects/${id.namespace}/${id.slug}`})
+    .then((response) => {
+      const projectId = response.body["id"];
+      if (response.status == 200) {
```
olevski, line 52: suggestion: You can eliminate this if statement and the call to the GET method if you want because deleting a non-existing project should return a 204 as well. But you can check for a 204 or 404 in the call to the DELETE endpoint. And both of those will be acceptable status codes.

Alternatively you can also flip the order and remove the if statement. I.e. first call DELETE and make sure it is 204 or 404 and then call GET on the project id and make sure it is 404.

cypress-tests/cypress/e2e/dashboardV2.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+    cy.getDataCy("search-card").should("have.length.at.least", 1);
+    cy.getDataCy("search-card").should("contain.text", `${prefixProjectTitle} ${projectIdentifier.slug}`);
+
+    // logout and search for the project and log back in
+    cy.logout();
+    cy.visit("v2");
+    cy.getDataCy("projects-container").contains("No 2.0 projects.");
+    cy.getDataCy("view-other-projects-btn").click();
+    cy.getDataCy("search-results").should("not.contain.text", `${prefixProjectTitle} ${projectIdentifier.slug}`);
+    cy.robustLogin();
```
olevski, line 75: suggestion: I think the three sections in this single `it` clause should be 3 separate `it` clauses. Something like:
- `it("Can find a private project in the dashboard")`
- `it("Can find a private project in the search")`
- `it("Cannot find a private project when logged out")`

And the describe section above should be something like `Basic private project functionality` rather than `Basic Dashboard functionality`. Because only one set of statements in this test (the first three) test anything about the dashboard.

cypress-tests/cypress/support/commands/projectsV2.ts
```diff
@@ -0,0 +1,77 @@
...
+  slug: string;
+  namespace?: string;
+};
+
+interface NewProjectV2Props extends ProjectIdentifier {
+  visibility?: "public" | "private";
+  name: string;
+}
+
+function getUserNamespaceAPIV2(): Cypress.Chainable<string | null> {
```
olevski, line 13: It would be nice if we explained briefly that this uses the API only.


```suggestion
/** Get the namespace of the logged in user from the API. */
function getUserNamespaceAPIV2(): Cypress.Chainable<string | null> {
```

cypress-tests/cypress/support/commands/projectsV2.ts
```diff
@@ -0,0 +1,77 @@
...
+  return cy.request({ failOnStatusCode: false, method: "GET", url: `api/data/namespaces?minimum_role=editor` })
+    .then((response) => {
+      if (response.status === 200) {
+        const userNamespace = response.body?.filter((namespace) => namespace.namespace_kind === "user");
+        return userNamespace && userNamespace.length > 0 ? userNamespace[0].slug : null;
+      }
+      return null;
+    });
+}
+function createProjectIfMissingAPIV2(newProjectProps: NewProjectV2Props) {
```
olevski, line 23:
```suggestion
/** Create a project (if the project is missing) by using only the API. */ 
function createProjectIfMissingAPIV2(newProjectProps: NewProjectV2Props) {
```

cypress-tests/cypress/support/commands/projectsV2.ts
```diff
@@ -0,0 +1,77 @@
...
+        });
+    } else {
+      cy.visit(`v2/projects/${newProjectProps.namespace}/${newProjectProps.slug}`);
+      cy.get(`[data-cy="project-name"]`, { timeout: TIMEOUTS.vlong }).should(
+        "be.visible"
+      );
+    }
+  });
+}
+function deleteProjectFromAPIV2(id: ProjectIdentifier) {
```
olevski, line 48:
```suggestion
/** Delete a project by using only the API. */ 
function deleteProjectFromAPIV2(id: ProjectIdentifier) {
```

olevski (changes requested):
Thanks Andrea just a few minor changes and suggestions.

andre-code:
Thanks, @olevski, for your review. I’ve applied your suggestions in the latest commit

rokroskar:
Is this meant to be going to `master`?

cypress-tests/cypress/support/commands/projectsV2.ts
leafty, line 1: I think we should not add any more commands to cypress. Commands should be highly re-usable, e.g. `getDataCy` which we use everywhere.

These scripts can be simple methods we import into the relevant tests.

Also to note that adding commands requires doing some typing gymnastics with the `Cypress` instance, so it should be limited to methods used in large amounts of tests.

leafty (changes requested):

andre-code:
> Is this meant to be going to `master`?

No, it should go to [release-0.58.0](https://github.com/SwissDataScienceCenter/renku/tree/release-0.58.0)

andre-code:
Thanks, @leafty  and @olevski , for your code review. I’ve added your suggestions. I agreed to leave the commands for something more reusable across all tests, and I’ve included that change as well.

cypress-tests/cypress/e2e/dashboardV2.cy.ts
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  it("Can see own project on the dashboard", () => {
+    cy.visit("v2");
+    cy.getDataCy("dashboard-project-list").find("a").should("have.length.at.least", 1);
+    cy.getDataCy("dashboard-project-list").find("a").should("contain.text", `${prefixProjectTitle} ${projectIdentifier.slug}`);
+    cy.getDataCy("dashboard-project-list").find("a").should("contain.text", projectIdentifier.slug);
+  });
+
+  it("Can find project in the search", () => {
```
leafty, line 64:
```suggestion
  it("Can find project in the search results", () => {
```

cypress-tests/cypress/e2e/dashboardV2.cy.ts
```diff
@@ -0,0 +1,80 @@
...
+    cy.visit("v2");
+    cy.getDataCy("view-my-projects-btn").click();
+    cy.getDataCy("search-card").should("have.length.at.least", 1);
+    cy.getDataCy("search-card").should("contain.text", `${prefixProjectTitle} ${projectIdentifier.slug}`);
+  });
+});
+
+describe("Dashboard v2, Non-Authenticated Users", () => {
```
leafty, line 72: nitpick:

```suggestion
describe("Dashboard v2 - Non-Authenticated user", () => {
```

cypress-tests/cypress/e2e/publicProject.cy.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from "../support/commands/projects";
 import { validateLogin, getRandomString } from "../support/commands/general";
 
+
```
leafty, line 9: ?

leafty (commented):
Code looks good, please merge this with the corresponding UI release, not before.

andre-code:
Thank you for your feedback, @leafty. I've addressed all your suggestions.

olevski (approved):

leafty (approved):
Tests are good :+1:

leafty:
@andre-code you merged before the release of data-services and renku-ui, which means this breaks the release branch.

PR #3790, opened 2024-10-02
Update search services to 0.6.1

eikek:
/deploy

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -85,25 +85,6 @@ global:
     ## for a new project based on its own version.
     ## NOTE should only be set for CI deployments and development purposes.
     cli_version:
-  ## Events related values definitions
-  events:
-    streams:
-      projectCreated: "project.created"
-      projectUpdated: "project.updated"
-      projectRemoved: "project.removed"
-      projectAuthAdded: "projectAuth.added"
-      projectAuthUpdated: "projectAuth.updated"
-      projectAuthRemoved: "projectAuth.removed"
-      userAdded: "user.added"
-      userUpdated: "user.updated"
-      userRemoved: "user.removed"
-      groupAdded: "group.added"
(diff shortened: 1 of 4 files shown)


helm-chart/renku/values.yaml
```diff
@@ -88,22 +88,22 @@ global:
...
+      USER_ADDED: "user.added"
+      USER_UPDATED: "user.updated"
+      USER_REMOVED: "user.removed"
+      GROUP_ADDED: "group.added"
+      GROUP_UPDATED: "group.updated"
+      GROUP_REMOVED: "group.removed"
+      GROUPMEMBER_ADDED: "groupMember.added"
+      GROUPMEMBER_UPDATED: "groupMember.updated"
+      GROUPMEMBER_REMOVED: "groupMember.removed"
+      DATASERVICE_ALLEVENTS: "data_service.all_events"
```
olevski, line 106: I am not sure why are these even in the values file at all. So this is technically a breaking change in the helm chart. But things like this should not be in the values file at all. This type of deep application configuration is something that a renku administrator has no way to set correctly. Nor would they know or understand what it is about. Even if they knew what it is about chances are that they are not free to set this to just any value.

So since we are touching this, we have 2 options:
- roll back this change and then we have no changes in the values file
- remove this section from the values file altogether (I think this is even better) and hardcode it in the deployment for the search where it is currently used.

olevski (changes requested):

helm-chart/renku/values.yaml
eikek: I'm now moving this section into a template. Thanks for the discussion!

olevski (approved):

eikek:
Hello @SwissDataScienceCenter/renku-product-management sorry for the ping :-) may I ask you to have a look at this pr? thanks!! 🙏🏼

PR #3792, opened 2024-10-03
feat: update gateway and add RBAC for new amalthea

olevski:
Update the gateway and add RBAC for new amalthea.

/deploy

### DIFF
--- a/helm-chart/renku/templates/notebooks/role.yaml
+++ b/helm-chart/renku/templates/notebooks/role.yaml
@@ -67,5 +67,17 @@ rules:
       - list
       - get
       - watch
+  - apiGroups:
+      - amalthea.dev
+    resources:
+      - amaltheasessions
+    verbs:
+      - create
+      - update
+      - delete
+      - patch
+      - list
+      - get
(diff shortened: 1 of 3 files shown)


helm-chart/renku/templates/notebooks/role.yaml
```diff
@@ -67,5 +67,17 @@ rules:
       - list
       - get
       - watch
+  - apiGroups:
+      - amalthea.dev
+    resources:
+      - amaltheasessions
```
aledegano, line 73: Is it gonna be okay to create a role that has permissions on a resource that doesn't exist (yet)?

olevski: I will try this on dev with a madeup CR. One sec.

olevski: Yes I just made a role on dev with a completely made up CR that does not exist. It was successfully created.

aledegano (approved):

lokijuhy (approved):

PR #3804, opened 2024-10-10
fix: Set the correct path of CA certs for OpenJDK based containers.

aledegano:
/deploy

### DIFF
--- a/.github/workflows/check-acceptance-test-code.yml
+++ b/.github/workflows/check-acceptance-test-code.yml
@@ -3,12 +3,23 @@ name: Check acceptance tests dependencies and code
 on:
   pull_request:
   merge_group:
+
+env:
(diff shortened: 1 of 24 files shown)


Panaetius (dismissed):

aledegano:
FYI this PR is breaking Search-API and Search-Provision, since for some reason and despite using the same base container, they are not happy about their change to CAcerts path.

Panaetius:
they don't fully use the same base container, e.g. graph uses eclipse-temurin:21-jre-alpine but search uses eclipse-temurin:21-jre (not alpine). That could be the reason, maybe it's a different path on alpine?

aledegano:
> they don't fully use the same base container, e.g. graph uses eclipse-temurin:21-jre-alpine but search uses eclipse-temurin:21-jre (not alpine). That could be the reason, maybe it's a different path on alpine?

I have found the problem: search-api and search-provision didn't have the init-container for the certificates. Now it should be fixed.

PR #3809, opened 2024-10-15
build: fix scala actions

leafty:


.github/workflows/check-acceptance-test-code.yml
```diff
@@ -3,12 +3,23 @@ name: Check acceptance tests dependencies and code
...
   formatting-check:
     name: Scala dependencies and code check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4.1.7
+      - name: Setup JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
```
Panaetius, line 20: nitpick: both graph and search use java 21, so might be good to use the same version in the acceptance tests. though I don't know if ubuntu 24.04 had 17 as a default.

leafty: The Dockerfile for the acceptance tests says 17 -> https://github.com/SwissDataScienceCenter/renku/blob/master/acceptance-tests/Dockerfile.

Panaetius (approved):

PR #3813, opened 2024-10-15
fix: Set the correct path of CA certs for OpenJDK based containers.

aledegano:
Also add the certificates init-container to search-api and search-provision.

Thanks to @JGodin-C2C for reporting the issue.

/deploy

PR #3817, opened 2024-10-17
fix: Set the correct path of CA certs for OpenJDK based containers.

aledegano:
Also add the certificates init-container to search-api and search-provision.
This is a port of #3813 to the `0.59.0` release.

/deploy


Panaetius:
I rebased 0.59.0 on master so with the release of 0.58.1 this PR should not be needed anymore

aledegano:
> I rebased 0.59.0 on master so with the release of 0.58.1 this PR should not be needed anymore

ah, right! good , thank you!

PR #3839, opened 2024-11-20
feat: consolidate network policies

olevski:
/deploy extra-values=networkPolicies.allowAllIngressFromNamespaces={default},networkPolicies.allowAllIngressFromPods[0].podSelector.matchLabels.app=allowAllIngress


Panaetius (approved):

Panaetius (commented):
There's still the setup-job-network-policy, not sure if you wanted to move that as well?

olevski:
>There's still the setup-job-network-policy, not sure if you wanted to move that as well?

I left that one on purpose. Because it already sits there near all the rest of the setup job helm temoplates.

helm-chart/renku/templates/network-policies.yaml
```diff
@@ -1,3 +1,16 @@
...
+kind: NetworkPolicy
+metadata:
+  name: default-deny-all-ingress
+  labels:
+    app: {{ template "renku.name" . }}
+    chart: {{ template "renku.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  podSelector: {}
```
aledegano, line 11: Will this policy deny ingress to **all** pods in the cluster? e.g. also pods that have nothing to do with Renku?...
Maybe I am wrong, but it reads like that...
If I am not wrong then we should make this policy be "deny all ingress to Renku pods".

helm-chart/renku/templates/notebooks/network-policy.yaml
```diff
@@ -1,108 +0,0 @@
...
-    - ports:
-        - protocol: UDP
-          port: 53
-        - protocol: TCP
-          port: 53
----
-apiVersion: networking.k8s.io/v1
-kind: NetworkPolicy
-metadata:
-  name: {{ template "renku.notebooks.fullname" . }}-sessions
```
aledegano, line 55: I am a bit unsure about the removal of this policy: it is pretty convenient to forbid Sessions to connect around the cluster where they are not supposed to.

Keep in mind that we (and potentially others) manage PG, Redis, Gitlab and other services independently from Renkulab, so the moment we deployed this version of Renkulab we would have to make sure that all those services have corresponding policies to deny user-sessions to connect to them.

Everything else you changed with this PR is great and useful, but I would keep this policy around, for egress only too.

WDYT?

helm-chart/renku/templates/network-policies.yaml
olevski: This denies ingress to ALL pods in the namespace where the helm chart is installed. Not to all pods in the cluster.

helm-chart/renku/templates/notebooks/network-policy.yaml
olevski: So this prevents sessions connecting to anything else in the renku release namespace. But it does not deny access to other namespaces. The problem with the egress policy is the "except internal IPs part". This part is sometimes wrong depending on the cluster setup and it prevents sessions from working because when they try to connect to things outside of the cluster the ips map to the ip ranges that are usually just private/internal.

helm-chart/renku/templates/network-policies.yaml
olevski: I did not want to call it "renku pods" because you may have other things in the namespace other than renku pods.

helm-chart/renku/templates/notebooks/network-policy.yaml
olevski: So Alessandro as we discussed over slack. I have added the following:
- deny all egress from user sesssions
- allow all egress form user sessions (but only for requests that target the release namespace) and this is fine because inside the release namespace we deny all ingress by default

olevski:
@aledegano @Panaetius this is ready.

helm-chart/renku/templates/notebooks/network-policy.yaml
aledegano: Okay, sounds good.
Thank you for getting to the bottom of this!

aledegano (approved):

PR #3861, opened 2024-12-22
feat: turn off notebook service

olevski:
This turns off the notebook service and uses the data service to manage JupyterServers.

/deploy

I tested in another throwaway CI deployment that when we update the sessions from v1 and v2 sessions are still visible to users. For both registered and anonymous users.

### DIFF
--- a/helm-chart/values.yaml.changelog.md
+++ b/helm-chart/values.yaml.changelog.md
@@ -11,8 +11,6 @@ Please follow this convention when adding a new row
   This feature is experimental - enable at your own risk! It depends on Shipwright (version >= 0.15.x) 
   which must be installed independently from the Renku Helm chart.
 
-## Upgrading to Renku 0.62.0
-
 * EDIT ``notebooks.replicaCount`` - set to zero by default because the data service is handling all sessions now, the notebook service will be fully decomissioned and removed from the helm chart in a subsequent PR.
 
 ## Upgrading to Renku 0.62.0
--- a/helm-chart/values.yaml.changelog.md
+++ b/helm-chart/values.yaml.changelog.md
@@ -13,7 +13,7 @@ Please follow this convention when adding a new row
 
 ## Upgrading to Renku 0.62.0
 
-* EDIT ``notebooks.replicaCount`` - set to zero by default because the data service is handling all sessions now, the notebook service will be fully decomissioned and remove from the helm chart in a subsequent PR.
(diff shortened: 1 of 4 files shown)


helm-chart/values.yaml.changelog.md
```diff
@@ -7,6 +7,10 @@ Please follow this convention when adding a new row
 
 ## Upgrading to Renku 0.62.0
 
+* EDIT ``notebooks.replicaCount`` - set to zero by default because the data service is handling all sessions now, the notebook service will be fully decomissioned and remove from the helm chart in a subsequent PR.
```
Panaetius, line 10:
```suggestion
* EDIT ``notebooks.replicaCount`` - set to zero by default because the data service is handling all sessions now, the notebook service will be fully decomissioned and removed from the helm chart in a subsequent PR.
```

helm-chart/values.yaml.changelog.md
```diff
@@ -7,6 +7,10 @@ Please follow this convention when adding a new row
 
 ## Upgrading to Renku 0.62.0
```
aledegano, line 8:
```suggestion
## Upgrading to Renku 0.66.0
```

aledegano (approved):

helm-chart/values.yaml.changelog.md
olevski: Omg. Ok I will fix this. Also will wait for the custom envs build PR to be merged then rebase and merge this.

bethcg (approved):
Just add the .  at the end of each of the improvements in CHANGELOG.rst to be consistent with the rest of the document.

olevski:
> Just add the . at the end of each of the improvements in CHANGELOG.rst to be consistent with the rest of the document.

I will open a separate PR for this.

lorenzo-cavazzi:
I'm not sure this is working fine. I can start v1 sessions but the UI doesn't show any and I can't connect.
This is the same error I'm getting in this renku-notebooks PR, so perhaps the latest version of renku-notebooks isn't working. Or values are missing in the PR deployments? In that case, you could try adding them in the `/deploy` string

olevski:
> I'm not sure this is working fine. I can start v1 sessions but the UI doesn't show any and I can't connect.
This is the same error I'm getting in this renku-notebooks PR, so perhaps the latest version of renku-notebooks isn't working. Or values are missing in the PR deployments? In that case, you could try adding them in the /deploy string

@lorenzo-cavazzi I just deleted your session with kubectl. I had a lot of leftover sessions from the acceptance tests so I cleaned everything up. I saw your session but did not see how old it was. And I deleted everything.

PR #3878, opened 2025-01-27
feat!: create a session image from a code repository

### LINKED ISSUE #3877: [Shape Up] Define a custom environment from a repository
See: https://renku.notion.site/Define-a-custom-environment-from-a-repo-1570df2efafc80f4b6b3ceb5a09d1569

This feature adds a new flow where the user can create an environment from a code repository.

leafty:
Closes #3877.

This feature adds a new option when creating a session launcher: users can now choose to build a session image from one of their code repositories.

Deployment: https://ci-renku-3878.dev.renku.ch/.

/deploy extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.pushSecretName=flora-docker-secret,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/flora-dev/


helm-chart/renku/values.yaml
```diff
@@ -1498,6 +1498,32 @@ dataService:
   trustedProxies:
     proxiesCount: 1
     realIpHeader: x-real-ip
+  # Build session images from code repositories with Shipwright
+  imageBuilders:
+    # Experimental feature: enable at your own risk!
```
olevski, line 1503: I think we can add a bit more information here. At least that it requires shipwright to be installed independently from the renku helm chart.

CHANGELOG.rst
```diff
@@ -3,6 +3,16 @@
 0.66.0
 ------
 
+TODO: feat! Define a custom environment from a repository
```
olevski, line 6: Remove the TODO here? And it to the user- facing features section.

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,11 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+## Upgrading to Renku 0.66.0
+
+* NEW `dataService.imageBuilders` to configure session image builds using Shipwright.
+  This feature is experimental and should not be enabled by RenkuLab administrators. Enable the feature at your own risk!
```
olevski, line 11: Same as above - we can at least add one more sentence that enabling this requires Shipwright to be installed independently from the renku helm chart.

olevski (changes requested):

olevski (approved):

aledegano:
> @SwissDataScienceCenter/yat we should make a proper robot account + repository in harbor and update the values for dev and dev-ci deployments so that the extra values aren't needed here.

yes, as soon as I am done with #3937, which will make changing these values so much easier.

rokroskar:
I believe the change that is needed will be the same. As in, these are non-standard values that need to be specified anyway.

leafty:
> I believe the change that is needed will be the same. As in, these are non-standard values that need to be specified anyway.

The defaults can work on dev, provided the buildstrategy and the secret are correct.

rokroskar:
> > I believe the change that is needed will be the same. As in, these are non-standard values that need to be specified anyway.
> 
> The defaults can work on dev, provided the buildstrategy and the secret are correct.

By default the feature is disabled so you have to add at least that to the values. But yes, if we just use the default secret and build strategy names that's the only thing that should be required.

helm-chart/renku/values.yaml
```diff
@@ -1498,6 +1498,33 @@ dataService:
...
+    buildRunTimeoutSeconds: 3600
+    ## The node selector for BuildRuns
+    nodeSelector:
+      renku.io/node-purpose: image-build
+    ## The tolerations for BuildRuns
+    tolerations:
+      - effect: NoSchedule
+        key: renku.io/dedicated
+        operator: Equal
+        value: image-build
```
rokroskar, line 1527: should this be commented out to keep the value functional in simple deployments?

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,13 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+## Upgrading to Renku 0.66.0
+
+* NEW `dataService.imageBuilders` to configure session image builds using Shipwright.
+  This feature is experimental and should not be enabled by RenkuLab administrators.
```
rokroskar, line 11: it should _not_ be enabled?

helm-chart/renku/templates/data-service/rbac.yaml
```diff
@@ -80,6 +80,18 @@ rules:
...
+    resources:
+      - buildruns
+    verbs:
+      - create
+      - update
+      - delete
+      - patch
+      - list
+      - get
+      - watch
```
olevski, line 94: Does this blow up if the cluster does not have the buildrun CRDs installed?

Same for the notebooks role changes below.

helm-chart/values.yaml.changelog.md
rokroskar: we list Shipwright as a requirement and the feature as experimental - I think that's fine

helm-chart/renku/values.yaml
leafty: This follows the current efforts from @aledegano to reduce the amount of custom values for CI/CD deployments. If you want to switch `enabled: true`, then having to adjust the values here is expected for now.

helm-chart/renku/templates/data-service/rbac.yaml
olevski: I just tried to create a role (with a role binding and service account) just like this in a bare kind cluster. It worked. So this will not blow things up if shipwright or tekton is not installed.

Panaetius (approved):

CHANGELOG.rst
```diff
@@ -10,7 +10,7 @@ User-Facing Changes
 
 - **UI**: Create a custom environment for your session based on an environment definition in a code repository.
   Current support includes creating python environments defined by an environment.yaml file or similar.
-  Learn more in our `documentation <https://renku.notion.site/How-to-create-a-custom-environment-from-a-code-repository-1960df2efafc801b88f6da59a0aa8234>`__.
+  This feature is only available on RenkuLab.io, Learn more in our `documentation <https://renku.notion.site/How-to-create-a-custom-environment-from-a-code-repository-1960df2efafc801b88f6da59a0aa8234>`__.
```
Panaetius, line 13:
```suggestion
  This feature is only available on RenkuLab.io, learn more in our `documentation <https://renku.notion.site/How-to-create-a-custom-environment-from-a-code-repository-1960df2efafc801b88f6da59a0aa8234>`__.
```

PR #3881, opened 2025-01-28
Add a cypress test for testing project basic operations: create, edit, delete

### LINKED ISSUE #3880: Tests - basic operations on the project: create, edit, delete.

lorenzo-cavazzi:
See title

fix #3880

### DIFF
--- a/cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
@@ -2,14 +2,14 @@ import {
   getRandomString,
   validateLoginV2,
 } from "../../support/commands/general";
+import { ProjectIdentifierV2 } from "../../support/types/project.types";
 import { User } from "../../support/types/user.types";
 import {
   deleteProjectFromAPIV2,
   getProjectByNamespaceAPIV2,
-  ProjectIdentifierV2,
 } from "../../support/utils/projectsV2.utils";
 
-const sessionId = ["dashboardV2", getRandomString()];
+const sessionId = ["projectBasics", getRandomString()];
 
 beforeEach(() => {
   // Restore the session (login)
--- a/cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
@@ -1,12 +1,18 @@
-import { getRandomString, validateLoginV2 } from "../../support/commands/general";
-import { deleteProjectFromAPIV2, getProjectByNamespaceAPIV2, ProjectIdentifierV2 } from "../../support/utils/projectsV2.utils";
+import {
+  getRandomString,
+  validateLoginV2,
+} from "../../support/commands/general";
+import { User } from "../../support/types/user.types";
+import {
+  deleteProjectFromAPIV2,
+  getProjectByNamespaceAPIV2,
+  ProjectIdentifierV2,
+} from "../../support/utils/projectsV2.utils";
 
-// ! const sessionId = ["dashboardV2", getRandomString()];
-const sessionId = ["dashboardV2", "123"];
-const username = Cypress.env("TEST_USERNAME");
+const sessionId = ["dashboardV2", getRandomString()];
 
 beforeEach(() => {
-  // Restore the session
+  // Restore the session (login)
   cy.session(
     sessionId,
     () => {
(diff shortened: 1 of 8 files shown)


cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      cy.getDataCy("navbar-project-new").click();
+      cy.location("hash").should("eq", "#create-project");
+      cy.getDataCy("new-project-modal").should("exist");
+      cy.getDataCy("project-name-input").type(projectName);
+      cy.getDataCy("project-slug-toggle").click();
+      cy.getDataCy("project-slug-input").should("have.value", projectPath);
+      cy.getDataCy("project-visibility-public").click();
+      cy.getDataCy("project-description-input").type(projectDescription);
+      cy.getDataCy("new-project-modal").contains(
+        `The URL for this project will be renkulab.io/v2/projects/${username}/${projectPath}`
```
leafty, line 63: Oh my, oh my, oh my, this is actually hard-coded 🙈 Ouch! 🤕

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      }
+    });
+  });
+
+  it("Project - create, edit and delete", () => {
+    // Create a new project
+    cy.visit("/v2");
+    cy.getUserData().then((user: User) => {
+      const username = user.username;
+      cy.get("#plus-dropdown").should("exist").click();
```
leafty, line 53: `.click()` will already test presence, visibility and uniqueness.

```suggestion
      cy.get("#plus-dropdown").click();
```

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      cy.getDataCy("navbar-project-new").click();
```
leafty, line 54:
```suggestion
      cy.getDataCy("navbar-project-new").contains("Project").click();
```

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      cy.location("hash").should("eq", "#create-project");
```
leafty, line 55: I'm not sure this line is very useful. The important part is to check that a form is presented and that completing it will make a new project, not the technical details about what the browser does.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      cy.getDataCy("new-project-modal").should("exist");
```
leafty, line 56: Add a test to check that the modal (or modal header) contains "Create a new project".

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      cy.getDataCy("project-name-input").type(projectName);
+      cy.getDataCy("project-slug-toggle").click();
+      cy.getDataCy("project-slug-input").should("have.value", projectPath);
+      cy.getDataCy("project-visibility-public").click();
+      cy.getDataCy("project-description-input").type(projectDescription);
+      cy.getDataCy("new-project-modal").contains(
+        `The URL for this project will be renkulab.io/v2/projects/${username}/${projectPath}`
+      );
+      cy.getDataCy("project-create-button").click();
```
leafty, line 65:
```suggestion
      cy.getDataCy("project-create-button").contains("Create").click();
```

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      cy.location("pathname").should(
+        "eq",
+        `/v2/projects/${username}/${projectPath}`
+      );
```
leafty, line 69: Instead of testing this, I think it is more important to test that the new page is the newly created project page, e.g. it should contain the project title and the project description.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      // Change settings
+      const modifiedProjectName = `${projectName} - modified`;
+      const modifiedProjectDescription = `${projectDescription} - modified`;
+      cy.getDataCy("project-settings-link").click();
```
leafty, line 74:
```suggestion
      cy.getDataCy("project-settings-link").contains("Settings").click();
```

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+      cy.getDataCy("project-settings-general")
+        .get(".alert-success")
+        .contains("The project has been successfully updated.");
+      cy.getDataCy("project-name").should("contain", modifiedProjectName);
+      cy.getDataCy("project-description").should(
+        "contain",
+        modifiedProjectDescription
+      );
+
+      // Revert settings
```
leafty, line 103: Why do this?

cypress-tests/cypress/support/commands/general.ts
```diff
@@ -19,6 +20,38 @@ export const validateLogin = {
...
+
+      expect(response.body).property("id").to.not.be.empty;
+      expect(response.body).property("id").to.not.be.null;
+      expect(response.body).property("username").to.not.be.empty;
+      expect(response.body).property("username").to.not.be.null;
+    });
+  },
+};
+
+export const getUserData = () => {
```
leafty, line 36:
```suggestion
export function getUserData(): User {
```

cypress-tests/cypress/support/commands/general.ts
```diff
@@ -43,6 +76,7 @@ export function getRandomString(length = 8) {
 export default function registerGeneralCommands() {
   Cypress.Commands.add("getIframe", getIframe);
   Cypress.Commands.add("getDataCy", getDataCy);
+  Cypress.Commands.add("getUserData", getUserData);
```
leafty, line 79: Do not add this as command, a utility function works perfectly well.

See: https://github.com/SwissDataScienceCenter/renku/pull/3781#discussion_r1781173793.

leafty (commented):
Thanks for adding this! 👏 

Some comments below.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: This shouldn't be necessary. `navbar-project-new` should already point to the element we want, disentangling it from the text/content that might change ("Project"? "New project"? Just an icon?).

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: I would say it's a similar case of https://github.com/SwissDataScienceCenter/renku/pull/3881/files#r1934094160
The idea behind using `getDataCy` is to disentangle getting the proper element from matching some specific text that might easily change in a future UI release.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: Good point, I'll change it!
Similar to https://github.com/SwissDataScienceCenter/renku/pull/3881/files#r1933949049

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
leafty: That text content had never changed and it is quite important to know it says "Project" and not "Dataset" or something else.

https://docs.cypress.io/app/core-concepts/best-practices#Text-Content

> If the answer is yes because the word Submit is critical and should not be changed - then use cy.contains() to target the element. This way, if it is changed, the test will fail.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: Oh, it doesn't really cover any additional features. I'll remove it :+1:

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: fixed in 2f9a1256175006ed30996b31ee8ed90d95f90617

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: fixed in 2f9a1256175006ed30996b31ee8ed90d95f90617

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: fixed in 2f9a1256175006ed30996b31ee8ed90d95f90617

lorenzo-cavazzi:
@leafty thank you for the review!
I implemented all the suggestions except the ones asking to check for the element text after getting it though `getDataCy`. 
The reason is here https://github.com/SwissDataScienceCenter/renku/pull/3881#discussion_r1934124831 . I'm happy to further discuss it offline if you disagree or prefer a different approach.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
leafty: No, the idea is to CHECK that the button says "Project" and not something else. This is a major interaction ("Create a new project"), and so we should check that the important titles and buttons are present and correct.

Again, from the Cypress best practices:
> A rule of thumb is to ask yourself this:
> If the content of the element changed would you want the test to fail?
> If the answer is yes: then use [cy.contains()](https://docs.cypress.io/api/commands/contains)
> If the answer is no: then use a data attribute.

Here the text should not change unless there is a product decision to change it, so it good to do `cy.getDataCy("navbar-project-new").contains("Project")`, this will fail if we accidentally update the button text.

Note that we DO use the `data-cy` attribute to select the button, but we add the check on the button text because we do not want that button to change.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
leafty: Please remove or comment this test.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,119 @@
...
+      cy.getDataCy("project-description").should(
+        "contain",
+        modifiedProjectDescription
+      );
+
+      // Delete project
+      cy.getDataCy("project-settings-link").click();
+      cy.getDataCy("project-delete")
+        .should("exist")
+        .contains("Are you sure you want to delete this project?");
```
leafty, line 106: What about this? This could change as text content goes.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: Why would you want the test to fail if the text changes from "Project" to something else like "New project"? Do you think that's relevant for the acceptance test?

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: I thought `project-delete` was only the section, and matching the text would guarantee we have the proper rights to delete the project. But the section don't even show up without permissions.

I'll remove it :+1:

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: I am checking to ensure we are showing the correct expected URL.
What would be a reasonable alternative here? Perhaps just `/${username}/${projectPath}`?

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
leafty: If you use `matchCase: false` then that would not fail -> https://docs.cypress.io/api/commands/contains

It would fail upon a inserting a typo (e.g. "Projec") and would need to be updated only if a very important change is made in the UI.

lorenzo-cavazzi: And it will succeed even if we write "Delete project" :smile: 
I feel this discussion is getting a little off-topic.

To get back on track, here are the problems with accepting this suggestion:

* There is no clear rule for deciding when to check the text. You say "Project" and "Create" are important, I say "Public" and "Private" are important (what if we mess up there and users create private projects that are actually public?).  Either we always trust the `data-cy` property is correct, or we always doubt it and check the content.
* If we check for UI typos, we make the acceptance tests responsible for checking the UI. I'm very worried about going down that way because it overlaps with the [UI Cypress tests](https://github.com/SwissDataScienceCenter/renku-ui/tree/main/tests) and is not the goal of our acceptance tests.

P.S. Since the goals of the Cypress acceptance tests were a little hidden and not very clear in the Readme file, I made a PR to improve that #3884.

leafty: OK, we can go without matching text for now in the tests here.

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  it("Project - create, edit and delete", () => {
+    // Create a new project
+    cy.visit("/v2");
+    getUserData().then((user: User) => {
+      const username = user.username;
+      cy.getDataCy("navbar-new-entity").click();
+      cy.getDataCy("navbar-project-new").click();
+      cy.getDataCy("new-project-modal").should("exist");
```
leafty, line 56: From the `data-cy` conversation, I think we should instead check against the presence of a new project _form_ and not a _modal_. The fact that the form is presented inside a modal is not important and could change (a page, an off canvas, etc.) but there should always be a `data-cy=new-project-form` element presented to the user.

lorenzo-cavazzi: Yes, this makes sense! I dropped the use of `new-project-modal` and match the form instead 2061f524d8cb48cebaf33483b3385e1d6d00be6a

cypress-tests/cypress/e2e/v2/projectBasics.cy.ts
lorenzo-cavazzi: Oh, true. I made this more resilient to the URL changes and, following your other suggestion on dropping the matcher for `new-project-modal`, I now match the URL element more precisely 2061f524d8cb48cebaf33483b3385e1d6d00be6a

PR #3904, opened 2025-02-12
Add a Cypress test to search entities

lorenzo-cavazzi:
See title

fix https://github.com/SwissDataScienceCenter/renku/issues/3901

### DIFF
--- a/cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
@@ -1,0 +1,174 @@
+import {
+  getRandomString,
+  getUserData,
+  validateLoginV2,
+} from "../../support/commands/general";
+import { User } from "../../support/types/user.types";
+import {
+  createGroupIfMissingAPI,
+  deleteGroupFromAPI,
+  getGroupFromAPI,
+} from "../../support/utils/group.utils";
+import {
+  createProjectIfMissingAPIV2,
+  deleteProjectFromAPIV2,
+  getProjectByNamespaceAPIV2,
+} from "../../support/utils/projectsV2.utils";
+
+const sessionId = ["searchEntities", getRandomString()];
+
+describe("Search for resources: groups, projects, users", () => {
+  // Define projects and groups details
+  // ? Random string is currently the only way to get good results from search
+  // ? since searching for multiple words may return a bunch a unexpected results.
+  const stringRandomOne = getRandomString();
+  const stringRandomTwo = getRandomString();
+  const stringFirst = "first";
+  const stringSecond = "second";
+  const stringProject = "project";
+  const stringGroup = "group";
+  const projects = {
+    first: `${stringProject}-${stringFirst}-${stringRandomOne}-${getRandomString()}`,
+    second: `${stringProject}-${stringSecond}-${stringRandomTwo}-${getRandomString()}`,
+  };
+  const groups = {
+    first: `${stringGroup}-${stringFirst}-${stringRandomTwo}-${getRandomString()}`,
+    second: `${stringGroup}-${stringSecond}-${stringRandomOne}-${getRandomString()}`,
+  };
+
+  // Create the requires resources
+  before(() => {
+    // Login
+    cy.session(
+      sessionId,
+      () => {
+        cy.robustLogin();
+      },
+      validateLoginV2,
+    );
+
+    // Create groups and projects -- fail early if any of the resources are not created
+    getUserData().then((user: User) => {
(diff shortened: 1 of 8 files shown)


cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    cy.getDataCy("search-input").clear().type(stringRandomOne);
+    cy.getDataCy("search-button").click();
+    cy.wait("@searchQueryOne");
+    cy.getDataCy("search-card").should("have.length", 1);
+
+    cy.getDataCy("search-filter-type-group").filter(":visible").click();
+    cy.getDataCy("search-filter-type-project").filter(":visible").click();
+    cy.wait("@searchQueryOne");
+    cy.getDataCy("search-card").should("have.length", 1);
```
olevski, line 152: suggestion: Same as above - and in this case it is even easier because there is only one result. Lets assert that we got the right projects or groups back and not something else.

cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
```diff
@@ -0,0 +1,180 @@
...
+      ),
+    ).as("searchQueryOne");
+    cy.getDataCy("search-input").clear().type(stringRandomOne);
+    cy.getDataCy("search-button").click();
+    cy.wait("@searchQueryOne");
+    cy.getDataCy("search-card").should("have.length", 2);
+    cy.intercept(
+      new RegExp(
+        `(?:/ui-server)?/api/search/query\\?q=.*?${stringRandomTwo}.*`,
+      ),
```
olevski, line 132: question/suggestion: If you intercept a less specific path to the search api isn't it easier? I.e. if you intercept all requests to the search api then you can always just wait for those interceptions to complete and you would not have `@searchQueryOne` and `@searchQueryTwo`? You can just wait for `@searchQuery`.

olevski (changes requested):
Just a few minor changes.

cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
```diff
@@ -0,0 +1,180 @@
...
+          deleteGroupFromAPI(grp);
+        }
+      });
+    });
+  });
+
+  it("Can search for entities and filter works", () => {
+    cy.visit("/v2");
+
+    // Search for string
```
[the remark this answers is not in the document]
lorenzo-cavazzi: We could, but that would trigger the login again. It seems overkill for a bunch of actions that take so little time. I would keep a slightly less clean output but a much faster execution time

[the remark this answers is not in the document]
lorenzo-cavazzi: It makes sense to separate into different sections when we check features that might fail independently and the interactions aren't too tightly related.
While checking projects, data connectors are different from code repos. But here we are trying a few filters without covering everything, so knowing it failed exactly on the "date creation" filter doesn't seem too relevant.

Also, the comments wouldn't be suitable separators.  I want the click on the project filter to preserve the previous search; if I moved that to a separate section, I would need to repeat most previous interactions.
If you think it's valuable, I could still separate the parts "Search with filters" and "Search for users," but it's a little hard to clearly define where to split something into a separate section.

cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
lorenzo-cavazzi: I now more precisely check for group and project names wherever I check the length 893dff814f5af245d8ba87291c206f98f313f061

cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
lorenzo-cavazzi: You're right, no need to differentiate. Here I used a generic matcher 893dff814f5af245d8ba87291c206f98f313f061

lorenzo-cavazzi:
@olevski Thank you for the review! I answered inline.

Mind that 893dff814f5af245d8ba87291c206f98f313f061 contains an additional change: a short waiting time after the creation of the resources since I noticed the search engine might need a (very short) time to index the groups and occasionally failed locally. That seems to happen only when we have more existing resources (hundreds?)

cypress-tests/cypress/e2e/v2/searchEntities.cy.ts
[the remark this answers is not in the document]
olevski: Ok no problem we leave it as is for now

olevski (approved):

PR #3906, opened 2025-02-13
Add a Cypress test to check anonymous users can only access public resources

lorenzo-cavazzi:
See title.

Mind that this will be rebased after closing #3887 and #3901

fix https://github.com/SwissDataScienceCenter/renku/issues/3903

### DIFF
--- a/cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
@@ -1,0 +1,202 @@
+import {
+  getRandomString,
+  getUserData,
+  validateLoginV2,
+} from "../../support/commands/general";
+import { User } from "../../support/types/user.types";
+import {
+  createProjectIfMissingAPIV2,
+  deleteProjectFromAPIV2,
+  getProjectByNamespaceAPIV2,
+} from "../../support/utils/projectsV2.utils";
+
+const sessionId = ["anonymousNavigation", getRandomString()];
+
+describe("Anonymous users can only access public resources", () => {
+  // Define some project details
+  let randomString: string;
+  let privateProjectName: string;
+  let publicProjectName: string;
+
+  function resetRequiredResources() {
+    randomString = getRandomString();
+    privateProjectName = `anon-private-${randomString}`;
+    publicProjectName = `anon-public-${randomString}`;
+  }
+
+  function browserLogout() {
+    // Logging out from the browser ensures cookies are cleared
+    cy.visit("/v2");
+    cy.getDataCy("navbar-toggle-user-menu").click();
+    cy.getDataCy("navbar-logout").click();
+    cy.getDataCy("renku-root").should("be.visible");
+  }
+
+  // Restore the session (login) and create the required projects
+  beforeEach(() => {
+    // Login
+    cy.session(
+      sessionId,
+      () => {
+        cy.robustLogin();
+      },
+      validateLoginV2,
+    );
+
+    // Create projects with new names to be sure re-runs don't break.
+    resetRequiredResources();
+    getUserData().then((user: User) => {
+      for (const proj of [privateProjectName, publicProjectName]) {
+        createProjectIfMissingAPIV2({
+          name: proj,
+          namespace: user.username,
+          slug: proj,
+          visibility: proj.includes("public") ? "public" : "private",
+        }).then((response) => {
+          // eslint-disable-line max-nested-callbacks
+          if (response.status >= 400) {
+            throw new Error("Failed to create project");
+          }
+        });
+      }
+    });
+  });
+
+  // Cleanup the project after the test
+  afterEach(() => {
+    // Login -- mind we might be logged out
+    cy.session(
+      sessionId,
+      () => {
+        cy.robustLogin();
+      },
+      validateLoginV2,
+    );
+
+    // Delete the projects
+    getUserData().then((user: User) => {
+      [privateProjectName, publicProjectName].forEach((proj) => {
+        getProjectByNamespaceAPIV2({
+          namespace: user.username,
+          slug: proj,
+        }).then((response) => { // eslint-disable-line max-nested-callbacks
+          if (response.status === 200) {
+            deleteProjectFromAPIV2({
+              id: response.body.id,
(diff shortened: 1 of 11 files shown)


cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
```diff
@@ -0,0 +1,202 @@
...
+  let privateProjectName: string;
+  let publicProjectName: string;
+
+  function resetRequiredResources() {
+    randomString = getRandomString();
+    privateProjectName = `anon-private-${randomString}`;
+    publicProjectName = `anon-public-${randomString}`;
+  }
+
+  function browserLogout() {
```
leafty, line 27: Please use `cy.session()` and not this, it is prone to errors. Cypress already supports having separate cookies etc. when giving different session IDs.

lorenzo-cavazzi: I restore the session in the `beforeEach` section, and then manually log in and out throughout the test using `browserLogout` and `robustLogin`.
What do you mean by "use cy.session() and not this"?

leafty: See https://github.com/SwissDataScienceCenter/renku/blob/425dfd82ba1307edfd3d06ddfe273ccac4f37d3c/cypress-tests/cypress/e2e/privateProject.cy.ts#L57

leafty: > then manually log in and out throughout the test using browserLogout and robustLogin.

This is exactly what is error prone. Prefer using cy.session to switch between users as per the cypress docs https://docs.cypress.io/api/commands/session#Switching-sessions-inside-tests

olevski: @leafty you mean that instead of logging out through the ui the best thing is to just  switch to another random session. Or a session that essentially is the "empty" session of a user that is not logged in. Right?

I think we can reserve a random session ID or a specific session ID for this purpose. A user that is not logged in and just switch to it.

leafty: Yes, exactly. Logging out through the UI is slow and I had troubles when I refactored the gateway, which is why the code I pointed earlier was changed to use `cy.session()` to switch between users.

Logging out through the UI should only be done in one or two tests which specifically check the log-in then log-out flow.

leafty: To be clear, what my comment is about: remove the `browserLogout()` function and use `cy.session()` instead.

cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
```diff
@@ -0,0 +1,202 @@
...
+    });
+
+    // Verify the previously private project is now publicly visible
+    cy.visit("/v2");
+    cy.getDataCy("navbar-login").should("be.visible");
+    cy.getDataCy("navbar-link-search").click();
+    cy.getDataCy("search-input").clear().type(randomString);
+    cy.getDataCy("search-button").click();
+    cy.wait("@searchQuery");
+    cy.getDataCy("search-card").should("have.length", 2);
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Good point :+1:  Done here 2af4cb18a4a673c7181e22f03a62b881e2ce3b07
P.S. I used the `randomString` string wherever we had more than one resource to simplify the logic while keeping the test effective.

cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
```diff
@@ -0,0 +1,202 @@
...
+    publicProjectName = `anon-public-${randomString}`;
+  }
+
+  function browserLogout() {
+    // Logging out from the browser ensures cookies are cleared
+    cy.visit("/v2");
+    cy.getDataCy("navbar-toggle-user-menu").click();
+    cy.getDataCy("navbar-logout").click();
+    cy.getDataCy("renku-root").should("be.visible");
+  }
```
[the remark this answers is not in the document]
olevski: Sounds good.

cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
```diff
@@ -60,6 +61,9 @@ describe("Anonymous users can only access public resources", () => {
         });
       }
     });
+
+    // ? Wait a little time to allow the search engine to index resources
+    cy.wait(TIMEOUTS.short);
```
leafty, line 66: why not query the search API with a loop and timeout here? this will prevent flakes and avoid guessing the indexing time.

Feel free to ignore in this PR, but then track this in an issue.

leafty: the other solution is to repeatedly search with the UI (click again on the search button) until we get the expected result.

cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
lorenzo-cavazzi: I switched to using sessions instead of manually logging in/out https://github.com/SwissDataScienceCenter/renku/pull/3906/commits/bcce7bfc38beb88f927da32a6ebc2daddb6060eb

I agree it's a better solution, especially since I ended up logging out/in multiple times, slowing down the test.

cypress-tests/cypress/e2e/v2/anonymousNavigation.cy.ts
lorenzo-cavazzi: I like your first suggestion! Implemented here https://github.com/SwissDataScienceCenter/renku/pull/3906/commits/f1814398bec0ffc5eb4aa430d2f3752492c58df2

leafty (approved):
LGTM

olevski (approved):

olevski:
My bad I misclicked and dismissed @leafty's approval. That was a mistake.

leafty:
I will dismiss yours then 😆

PR #3909, opened 2025-02-14
feat: Make Redis and OIDC client secrets optional

aledegano:
These are the last two secrets that were needed to be generated out-of-bands before deploying Renkulab.

Thus the `generate-values` script is not really necessary anymore and in its place there is a `minimal-deployment` values file that in a future commit will also be used as a base for CI deployments.

The documentation has been updated to reflect the removal of the script.

Further cleanup and simplification of the Helm chart will result in a smaller and simpler minimum-deployment file, with the ultimate goal of being able to issue `helm install renku/renku` resulting in a functioning development deployment.


minimal-deployment/minimal-deployment-values.yaml
```diff
@@ -0,0 +1,47 @@
+amalthea:
+  deployCrd: false
```
olevski, line 2: For a functional deployment from nothing in a fresh cluster this should be true.

```suggestion
  deployCrd: true
```

minimal-deployment/minimal-deployment-values.yaml
```diff
@@ -0,0 +1,47 @@
...
+amalthea-sessions:
+  deployCrd: false
```
olevski, line 4: same as above

```suggestion
  deployCrd: true
```

olevski (changes requested):
This is awesome @aledegano!!! Just two really small suggestions - feel free to ignore if you wish.

olevski (approved):

PR #3912, opened 2025-02-18
Add a Cypress test to check sessions run correctly

lorenzo-cavazzi:
See title.

/deploy renku-ui=main

fix https://github.com/SwissDataScienceCenter/renku/issues/3907

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -149,12 +149,13 @@ jobs:
             useSession,
             checkWorkflows,
             rstudioSession,
+            v2/anonymousNavigation,
             v2/dashboardV2,
-            v2/projectBasics,
             v2/groupBasics,
+            v2/projectBasics,
             v2/projectResources,
             v2/searchEntities,
-            v2/anonymousNavigation
+            v2/sessionBasics,
           ]
 
     steps:
--- a/cypress-tests/cypress/e2e/v2/sessionBasics.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/sessionBasics.cy.ts
@@ -1,0 +1,70 @@
+import {
+  getRandomString,
+  getUserData,
+  validateLoginV2,
+} from "../../support/commands/general";
+import { ProjectIdentifierV2 } from "../../support/types/project.types";
+import { User } from "../../support/types/user.types";
+import {
+  createProjectIfMissingAPIV2,
+  deleteProjectFromAPIV2,
+  getProjectByNamespaceAPIV2,
+} from "../../support/utils/projectsV2.utils";
+
+// !const sessionId = ["sessionBasics", getRandomString()];
+const sessionId = ["sessionBasics", 1];
+
+describe("Start a session that consumes project resources", () => {
+  // Define required resources
+  let projectNameRandomPart: string;
+  let projectName;
+  let projectPath;
+  let projectIdentifier: ProjectIdentifierV2;
+
+  function resetRequiredResources() {
+    projectNameRandomPart = getRandomString();
+    projectName = `session-basics-${projectNameRandomPart}`;
+    projectIdentifier = {
+      slug: projectPath,
+      id: null,
+      namespace: null,
+    };
+  }
+
+  beforeEach(() => {
+    // Restore the session (login)
+    cy.session(
+      sessionId,
+      () => {
+        cy.robustLogin();
+      },
+      validateLoginV2
+    );
+
+    // Create a project
+    resetRequiredResources();
+    getUserData().then((user: User) => {
+      projectIdentifier.namespace = user.username;
+      createProjectIfMissingAPIV2({
+        name: projectName,
+        namespace: user.username,
+        slug: projectName,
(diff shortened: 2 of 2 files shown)


cypress-tests/cypress/e2e/v2/sessionBasics.cy.ts
```diff
@@ -0,0 +1,195 @@
...
+      timeout: TIMEOUTS.long,
+    }).should("be.visible");
+    cy.getDataCy("session-header").contains(session.name);
+    cy.get("[data-cy=session-iframe]", { timeout: TIMEOUTS.vlong }).should(
+      "be.visible",
+    );
+
+    // Check the project resources are available in the session.
+    cy.getIframe("[data-cy=session-iframe]").within(() => {
+      // ? Mind the following commands target the VSCode web interface where we have little control.
```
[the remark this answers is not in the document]
lorenzo-cavazzi: I found it more convenient than other solutions to easily check that code repos and data sources are mounted correctly (E.G. the file browser is neat and there are additional attributes to check folder depth, while Jupiter was harder to test with Cypress since its UI occasionally refreshes components leading to flakyness). Bonus point for (likely) being a popular solution for our users.

I'm open to more lightweight alternatives -- as long as their UI can be easily tested with Cypress.

.github/workflows/pull-request-test.yml
```diff
@@ -149,12 +149,13 @@ jobs:
...
             rstudioSession,
+            v2/anonymousNavigation,
             v2/dashboardV2,
-            v2/projectBasics,
             v2/groupBasics,
+            v2/projectBasics,
             v2/projectResources,
             v2/searchEntities,
-            v2/anonymousNavigation
+            v2/sessionBasics,
```
leafty, line 158: Nitpick: why is this not just `v2/sessions`?

cypress-tests/cypress/e2e/v2/sessionBasics.cy.ts
```diff
@@ -0,0 +1,195 @@
...
+  it("Start a session with VSCode and check it has the necessary resources", () => {
+    const dataConnectorName = `giab-${getRandomString()}`;
+    const repositoryName = "renku-ui";
+    const session = {
+      arguments:
+        '["/entrypoint.sh jupyter server --ServerApp.ip=0.0.0.0 --ServerApp.port=8888 --ServerApp.base_url=$RENKU_BASE_URL_PATH --ServerApp.token=\\"\\" --ServerApp.password=\\"\\" --ServerApp.allow_remote_access=true --ContentsManager.allow_hidden=true --ServerApp.allow_origin=* --ServerApp.root_dir=\\"/home/jovyan/work\\""]',
+      command: '["sh","-c"]',
+      image: "registry.renkulab.io/rok.roskar/vscode-example:0c10599",
+      name: "vscode-launcher",
+      url: "/vscode",
```
leafty, line 75: I would use `renku/renkulab-vscodium-python-runimage:ubuntu-59fcea4` which is an official image from the Renku team and not something hosted at `registry.renkulab.io/rok.roskar`.

That image also has simpler arguments: `command=["bash"]`, `arguments=["/entrypoint.sh"]`, `workdir=/home/ubuntu/work`, `mountdir=/home/ubuntu/work`

cypress-tests/cypress/e2e/v2/sessionBasics.cy.ts
```diff
@@ -0,0 +1,195 @@
...
+    }).should("be.visible");
+    cy.getDataCy("session-header").contains(session.name);
+    cy.get("[data-cy=session-iframe]", { timeout: TIMEOUTS.vlong }).should(
+      "be.visible",
+    );
+
+    // Check the project resources are available in the session.
+    cy.getIframe("[data-cy=session-iframe]").within(() => {
+      // ? Mind the following commands target the VSCode web interface where we have little control.
+      const findAndExpandFolder = (name: string, depth: number) => {
```
leafty, line 166: style:
```suggestion
      function findAndExpandFolder(name: string, depth: number) {
```

leafty (approved):
Looks good

cypress-tests/cypress/e2e/v2/sessionBasics.cy.ts
lorenzo-cavazzi: Oh, good suggestion! I only had to change a few other details and it works well :+1: 
https://github.com/SwissDataScienceCenter/renku/pull/3912/commits/1097b60a1f96f059e8693ed9636821e202ea4a14

.github/workflows/pull-request-test.yml
lorenzo-cavazzi: I followed a similar pattern to the names I gave to the other files. I thought adding other tests for the same target (E.G. sessionWithCredentials) would avoid overlapping names, but it isn't very important.
I can simplify it, but I would rather change all the other names too to use a similar pattern everywhere (groupBasics --> groups, ...)

leafty (approved):
LGTM

Note: I wanted to watch the cypress video but I can't. It seems the uploaded artefacts are overwriting each other.

PR #3922, opened 2025-02-28
feat: allow setting timeouts for builds

Panaetius:
tested in https://github.com/SwissDataScienceCenter/renku-data-services/pull/674


olevski (approved):
This looks fine to me. You may want to get someone who is working more on the build to double check if you want.

Also the other commits that went into `build/session-env-builders` have not really been reviewed. So we will review the whole feature branch PR anyhow.

PR #3934, opened 2025-03-04
revert: Configure data services to connect to solr

olevski:
Refs: e831264a

/deploy


olevski:
We figured out that the Solr helm chart will not create additional cores after solr has been initialized at least once. So the new feature that we thought was ready was not handling the creation of solr cores.

We will need to add this functionality to create solr cores somwhere and then add this feature back in.

This will have to be done in a different release. Not in 0.65.0 which will be released today.

lorenzo-cavazzi (approved):
:+1:

PR #3937, opened 2025-03-05
feat: Update Renku-Actions to 1.15.0

aledegano:
- chore(ci): Update renku actions to v1.15.0
- use base values from local files instead of the GH secret
- Add default resources values file
- fix: Cypress, useSession tolerate multiple retry


lorenzo-cavazzi:
Just a comment: it would be easier to check the tests through the videos if you use 1.15.1 instead of 1.15.0

olevski (changes requested):
This is great Alessandro! Everything looks good.

The only minor thing is to mention somewhere in the values file (At the top) and also probably in the changelog that the default resource requests in the values file are not suitable for production and can only be used for a small CI deployment.

Actually the values yaml changelog should mention this too. Because  when you upgrade to this version of Renku you will go from potentially no resource requests/limits in some cases to very low values. I can imagine this can cause a lot of trouble for admins.

aledegano:
> This is great Alessandro! Everything looks good.
> 
> The only minor thing is to mention somewhere in the values file (At the top) and also probably in the changelog that the default resource requests in the values file are not suitable for production and can only be used for a small CI deployment.
> 
> Actually the values yaml changelog should mention this too. Because when you upgrade to this version of Renku you will go from potentially no resource requests/limits in some cases to very low values. I can imagine this can cause a lot of trouble for admins.

Thank you, all done.

olevski (approved):

PR #3938, opened 2025-03-05
chore: update all actions to use ubuntu-24.04

olevski:
NOTE: This is merging straight into `master`.

/deploy

### DIFF
--- a/.github/workflows/auto-reqs-update-pr.yml
+++ b/.github/workflows/auto-reqs-update-pr.yml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch
 jobs:
   auto-pull-request:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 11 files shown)


.github/workflows/auto-reqs-update-pr.yml
```diff
@@ -3,7 +3,7 @@ on:
   workflow_dispatch
 jobs:
   auto-pull-request:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
aledegano, line 6: just curious: is this change to be explicit instead of implicit (that I would very much approve of), or there is a factual change?

olevski: There is no actual change. I just thought it is nicer to have this pinned and to be uniform everywhere.

Panaetius (approved):

PR #3949, opened 2025-03-12
test: cleanup sessions after the tests

lorenzo-cavazzi:
Hanging sessions aren't ideal since the CI cleanup action fails.
This PR deletes any running session for the projects it creates

/deploy


rokroskar:
When is 0.66 supposed to be released? If not very soon, then maybe this should go straight to master?

Panaetius:
like 10 days from now would be the regular end of cooldown release, unless we want to release the buildpack stuff earlier

aledegano (approved):

leafty:
Ironic: https://github.com/SwissDataScienceCenter/renku/actions/runs/13858120465/job/38779556515

PR #3966, opened 2025-03-26
chore: update keycloak

rokroskar:
/deploy-azure


olevski:
@rokroskar I propose that this goes out into 0.69.0 or a later release. 0.68.0 will be tagged at noon tomorrow.

PR #3976, opened 2025-04-06
feat: add option to turn off legacy services

olevski:
/deploy


wesjdj:
I think `enableV1Services` could be a better name. There might be other services we want to disable in the future, and that makes the distinction clearer.

olevski:
@wesjdj I made the change you requested and renamed the option in the values file.

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,10 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+## Upgrading to Renku 0.68.0
+
+* NEW `enableLegacyServices` used to indicate whether services needed exclusively for Renku V1 are deployed or not. The support for this feature is experimental and it should not be used in production at all.
```
rokroskar, line 10:
```suggestion
* NEW `enableV1Services` used to indicate whether services needed exclusively for Renku V1 are deployed or not. The support for this feature is experimental and it should not be used in production at all.
```

rokroskar: The statement "... should not be used in production at all" is a bit strange - we will definitely be using this in production afaik?

olevski: Yes. And we can change that line when we are ready to use it in production. For now no one should really be touching this. I consider the audience of this doc to be other admins in the wild who may not know all the intricacies and background on adding this option. These admins should really not be touching this flag at all that is why I added the line.

rokroskar: right - but we _have_ to use it, no?

olevski:
This will go into 0.69.0 or a later  release

PR #3980, opened 2025-04-10
chore: add CORE_SERVICE_URL to data services

olevski:
Adds the core service url to the data service so that it can get metadata about renku v1 projects.

The PR for adding the actual endpoint/functionality in the data service will be done in a separate pr.


olevski:
This will go into 0.69.0 or later release

PR #3982, opened 2025-04-11
feat: Add init script for Harbor.

aledegano:
This commit adds an initialization script to configure a Harbor instance to be used as image registry for Renkulab v2.

### DIFF
--- a/scripts/harbor-init/harbor-init.go
+++ b/scripts/harbor-init/harbor-init.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"flag"
+	"fmt"
 	"io"
 	"log"
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 )
 
 const (
@@ -16,10 +18,41 @@ const (
 	HARBOR_API_ENDPOINT = "/api/v2.0"
 )
 
+type SensitiveString string
+
+func (s SensitiveString) String() string {
+	return "<redacted sensitive string>"
+}
+
+func (s SensitiveString) Format(w fmt.State, v rune) {
+	_, err := w.Write([]byte(s.String()))
+	if err != nil {
+		panic(err)
+	}
+}
+
+func (s *SensitiveString) Set(value string) error {
+	// Set the sensitive string value
+	*s = SensitiveString(value)
+	return nil
+}
+
 type Login struct {
+	Client   *http.Client
 	Url      string
 	Username string
-	Password string
+	Password SensitiveString
+}
+
+func (l *Login) String() string {
+	return l.Url + " " + l.Username + " " + "<redacted password>"
+}
+
+func (l *Login) Format(w fmt.State, v rune) {
+	_, err := w.Write([]byte(l.String()))
+	if err != nil {
+		panic(err)
+	}
 }
 
 type Projects []struct {
@@ -39,15 +72,14 @@ type ResponseRobot struct {
 }
 
 func request(method string, login Login, endpoint string, body []byte) (*http.Response, []byte, error) {
-	client := &http.Client{}
 	req, err := http.NewRequest(method, login.Url+endpoint, bytes.NewBuffer(body))
 	if err != nil {
 		return nil, nil, err
 	}
-	req.SetBasicAuth(login.Username, login.Password)
+	req.SetBasicAuth(login.Username, string(login.Password))
 	req.Header.Set("Content-Type", "application/json")
 
-	resp, err := client.Do(req)
+	resp, err := login.Client.Do(req)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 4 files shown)


scripts/harbor-init/harbor-init.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return resp, respBody, nil
+}
+
+func main() {
+	urlFlag := flag.String("url", "https://harbor.example.com", "Harbor URL")
+	adminFlag := flag.String("admin", "admin", "Harbor admin username")
+	passwordFlag := flag.String("password", "", "Harbor admin password")
```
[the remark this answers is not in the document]
olevski: If you make this you may need to do `string(sensitiveStringVar)` when you use it in a place that requires `string` but you are certain that it cannot be accidentally printed  in the logs in your code.

scripts/harbor-init/harbor-init.go
```diff
@@ -0,0 +1,236 @@
...
+	robotSecretFlag := flag.String("secret", "", "Robot account secret")
+	flag.Parse()
+	// Initialize logger to print to stdout
+	logger := log.New(os.Stdout, "INFO: ", log.Ldate|log.Ltime)
+	login := Login{}
+	// Get login from environment variable
+	if *urlFlag == "" {
+		logger.Fatal("url flag is not set")
+	}
+	login.Url = *urlFlag + HARBOR_API_ENDPOINT
```
olevski, line 77: suggestion(non-blocking): Not sure what happens if `urlFlag` has a slash at the end. Beacuse `HARBOR_API_ENDPOINT` also has a slash at the beginning. Then you would have something like `https://harbor.io//api/v2.0` and hopefully this does not cause trouble. A call to `strings.TrimRight` will address this problem. But feel free to ignore if you wish.

scripts/harbor-init/harbor-init.go
```diff
@@ -0,0 +1,236 @@
...
+const (
+	// Harbor API version
+	HARBOR_API_ENDPOINT = "/api/v2.0"
+)
+
+type Login struct {
+	Url      string
+	Username string
+	Password string
+}
```
olevski, line 23: suggestion: Same as the password flag. You should set `String()` and `Format()` method here to not print the password. Or you can make the new type I suggested and make `Password` be a `SensitiveString`.

scripts/harbor-init/harbor-init.go
```diff
@@ -0,0 +1,236 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer resp.Body.Close()
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, nil, err
+	}
+	return resp, respBody, nil
```
olevski, line 59: suggestion(non-blocking): If the only reason you are doing this is to deserialize the json from the body... then there is an easier way:

```go
var deserializeIntoHere SomeType
err = json.NewDecoder(res.Body).Decode(&deserializeIntoHere)
```

So this method can just return the response and you can deserialize outside of this function if you want. But this is also fine - just fyi. If you want ignore this.

scripts/harbor-init/harbor-init.go
```diff
@@ -0,0 +1,236 @@
...
+	logger.Println("Authenticating with Harbor server")
+	resp, body, err := request("GET", login, "/audit-logs", nil)
+	if err != nil {
+		logger.Fatal("Error connecting to Harbor: ", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		logger.Fatal("Failed to authenticate:", resp.Status, err, string(body))
+		return
+	}
+	logger.Println("Authenticated successfully")
```
olevski, line 109: question: It may be that this is not needed if you send the credentials in basic auth header every time with the request method? Or if you want to keep cokies around then probabl you dont want to initialize http.Client from scratch on every call. I am pretty sure that wipes the cookies. Or you may need to add some kind of cookie jar provider to the http client if you want to reuse cookies.

scripts/harbor-init/harbor-init.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// Check if robot account already exists
+	logger.Println("Checking if robot account exists:", robotAccountName)
+	resp, body, err = request("GET", login, "/projects/"+projectName+"/robots", nil)
+	if err != nil {
+		logger.Fatal("Error creating robot account: ", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		logger.Fatal("Failed to get robot account:", err)
+		return
```
olevski, line 158:
```suggestion
```

olevski (changes requested):
Thanks Alessandro. A few minor suggestions. Some of them feel free to ignore if you wish.

scripts/harbor-init/harbor-init.go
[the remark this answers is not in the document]
aledegano: done

scripts/harbor-init/harbor-init.go
aledegano: added trimright

scripts/harbor-init/harbor-init.go
aledegano: not done, I am not sure how to achieve what I need (i.e. deserialize into different types)

scripts/harbor-init/harbor-init.go
aledegano: the client is now init in the main and passed around with the `login` type

scripts/harbor-init/harbor-init.go
olevski: ok np.

PR #3986, opened 2025-04-15
Revert "chore: consolidate security contexts (#3975)"

olevski:
This reverts commit 4f56c743e318bcb512af551c6409924bb10ef220.

/deploy

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,15 +3,9 @@
 0.68.0
 ------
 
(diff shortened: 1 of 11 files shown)


olevski:
This was causing the ui image to not start up fully. So we should revert it. To implement this fully we need to make some more deeper changes on the ui images.

lorenzo-cavazzi (approved):

aledegano (approved):

PR #3987, opened 2025-04-15
chore: bump renku core and templates versions

lorenzo-cavazzi:
This PR bumps the renku-core version and updated the templates

/deploy #legacy

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -327,14 +327,6 @@ keycloakx:
         - |
           echo "Copying theme..."
           cp -Rfv /renku_theme/* /theme
-      securityContext:
-        allowPrivilegeEscalation: false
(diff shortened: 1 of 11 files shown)


helm-chart/renku/values.yaml
```diff
@@ -1643,4 +1643,4 @@ affinity: {}
 versions:
   latest:
     image:
-      tag: v2.9.2
+      tag: v2.9.4
```
olevski, line 1646: This whole section is not needed. It is leftover from an old version of the helm chart. I cannot find .Values.versions being used anywhere in the helm chart.

And github for some reason will not allow me to do a suggestion here. You can just remove `versions` and anything inside of it.

olevski (changes requested):

olevski (approved):

PR #3989, opened 2025-04-24
feat: Add data tasks deployment

eikek:
Adds configuration for the data tasks deployment. This runs background tasks in the scope of data services in one python runtime.

/deploy


eikek:
> Looks good Ike. But I think the order is now like what we discussed during the meeting. This references the data services repo. So we should first merge the data services code and make a release on data services and then update the image version here.
> 
> An alternative to this is that we merge this right away. But for that you should set the replicas to 0 on the deployment.

Yes, I agree. Your first proposal is also what I recall from other talks about this. We do the release on data-services and then bump the image version in this PR. 👍🏼

olevski (approved):

PR #4000, opened 2025-05-07
fix: legacy acceptance tests

olevski:
/deploy #legacy


olevski:
This fixes a problem because it seems that the title used when the session is staring has been changed. It used to say starting and now it says launching. If you check on dev.renku.ch you will see the old text. But the new release has the new text.

The rstudio/jupyter sessions tests still fail because a fix on the data services that updates the session status correctly has not been merged. Once it is merged. Then the legacy acceptance tests will pass.

lorenzo-cavazzi (approved):
Lgtm :+1:

PR #4003, opened 2025-05-09
fix: update gateway to 1.5.1

olevski:
/deploy


CHANGELOG.rst
```diff
@@ -23,6 +23,7 @@ User-Facing Changes
 - **Core Service**: Fix a bug where removing activities wouldn't actually remove them.
 - **Data services**: Fix an issue where conflicting mount points would prevent sessions from starting.
 - **UI**: Fix an issue where pausing or resuming a session would crash the whole tab in Firefox.
+- **Gateway**: Do not crash if the metrics client is not enabled.
```
leafty, line 26: I don't think you need to add a changelog entry for a bug introduced and then fixed in the same release.

PR #4006, opened 2025-05-12
feat: update renku UI 3.35.0

andre-code:
Update renku-ui to 3.55.0, which includes:

- Card design for session launchers and sessions on the project page.
- Support for migrating projects from Renku v1 to Renku 2.0.
- Session UX improvements.

/deploy


rokroskar (approved):
🚀 Thanks!

PR #4007, opened 2025-05-12
[test pr]

leafty:
/AzureDeploy renku-data-services=fix-upgrade-kr8s


leafty:
Note: current k8s watcher token expires at `Mon May 12 2025 13:44:12 GMT+0200 (Central European Summer Time)`.

PR #4012, opened 2025-05-14
fix: update data service to 0.43.2

olevski:
This fixes a failing DB migration in the user preferences table.

/deploy


olevski:
I already tested the migration with first deploying an older version of renku, adding stuff and then adding this upgrade (in a separate CI deployment). And things worked. The DB migration does not fail anymore.

PR #4028, opened 2025-05-23
feat: make idlesness check consider active requests to session

Panaetius:
/deploy

This PR updates amalthea to 0.19.0, with the major change being that the authproxy sidecar now tracks active connections to the session and prevents idleness culling if there are still requests from the user, even if CPU is below the threshold. This is to prevent users who are just reading code in the tab but not doing any computation from having the session suddenly stop out from under them.

it also fixes URL paths in v2 sessions so they can work properly with e.g. RStudio.
As well as introducing arm images for amalthea.


wesjdj (approved):
LGTM

PR #4038, opened 2025-06-02
fix: update tests for DC

andre-code:
PR to update test of DC


olevski:
@andre-code this should target `release-2.4.0` right? With the reshuffling of the version we just did, that is the case, right?

andre-code:
> @andre-code this should target `release-2.4.0` right? With the reshuffling of the version we just did, that is the case, right?

Yes, I've already changed the base branch to `release-2.4.0`

leafty (approved):

PR #4055, opened 2025-06-19
test: update legacy tests for Sunset project dataset

andre-code:
PR to update legacy tests when form to add a new project is not available anymore

/deploy renku-ui=andrea/turn-off-renku-legacy-project-dataset-creation #legacy


andre-code:
We no longer support legacy tests.

PR #4065, opened 2025-06-25
chore: simplify ui config for openshift

olevski:
Makes it easier to run the ui client on open shift.

/deploy extra-values=ui.client.privacy.page.enabled=true

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -536,7 +536,7 @@ ui:
     replicaCount: 1
     image:
       repository: renku/renku-ui
-      tag: "4.1.0"
+      tag: "4.2.0"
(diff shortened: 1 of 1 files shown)


leafty (changes requested):
The behavior of the UI values has changed, now it is not possible to activate the terms pages without providing a correct value. Beforehand, activating the terms page with defaults would show a template page.

See current deployment:
Screenshot 2025-06-26 at 13 47 44

olevski:
> The behavior of the UI values has changed, now it is not possible to activate the terms pages without providing a correct value. Beforehand, activating the terms page with defaults would show a template page.

@leafty I fixed this. Now the sample values show up when the pages are enabled but no custom content has been set.

leafty (approved):

PR #4082, opened 2025-07-03
Upgrade UI to 4.2.1

ciyer:
/deploy


olevski:
The base was master. But it was meant to be `2.3.0`. So I changed it . Thanks for doing the bugfix @ciyer .

PR #4086, opened 2025-07-08
chore: add buildpack base images

rokroskar:
Adds an action to build images for the renkulab global environments from our buildpacks.

Project that tests these builds: https://renkulab.io/p/renku-team/test-new-global-images


.github/workflows/build-global-renkulab-images.yml
```diff
@@ -0,0 +1,53 @@
...
+      - uses: buildpacks/github-actions/setup-pack@v5.0.0
+      - name: Build image
+        run: |
+          tags=$(echo "${{ steps.meta.outputs.tags }}" | tr '\n' ',')
+          tags=${tags::-1}
+          pack build \
+            ${{ env.DOCKER_PREFIX }}/py-${{matrix.FLAVOR}}-${{matrix.FRONTEND}}:latest \
+            -p global-images/${{matrix.FLAVOR}} \
+            --builder ghcr.io/swissdatasciencecenter/renku-frontend-buildpacks/selector:0.0.6 \
+            --run-image ghcr.io/swissdatasciencecenter/renku-frontend-buildpacks/base-image:0.0.6 \
```
leafty: Suggestion: move these to `env`:
```yaml
BUILDER_IMAGE: ghcr.io/swissdatasciencecenter/renku-frontend-buildpacks/selector:0.0.6
RUN_IMAGE: ghcr.io/swissdatasciencecenter/renku-frontend-buildpacks/base-image:0.0.6
```

.github/workflows/build-global-renkulab-images.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3.2.0
```
leafty: Not needed

```suggestion
```

global-images/base/requirements.txt
olevski, line 1: question: What is the point of `base`? This is a base image that other can use to build their own images I guess? This will add an arbitrary python version. If you use environment.yaml (which can still be empty) you can pick the python version. But this is also something we can improve or change in the future.

.github/workflows/build-global-renkulab-images.yml
```diff
@@ -0,0 +1,53 @@
+name: Build global base images
+
+on:
+  push:
+    paths:
+      - "global-images/**"
+      - ".github/workflows/build-global-renkulab-images.yml"
```
olevski, line 7: issue: since we tag the images with "latest" every time this runs lets run this on pushes to master only, in addition to the restrictions you already have for code changes.

global-images/README.md
olevski, line 1: nitpick(non-blocking): IMO if you put the command for building the image in a `Makefile` and then you use the `Makefile` in the github action you can cut down a lot of the docs and you allow anyone to easily build the images without copying and pasting a large command.

The makefile can you use env vars to decide which flavour and frontend to build.

But we can implement this later at some other point. It is really not a problem.

olevski (changes requested):

global-images/README.md
rokroskar: yeah good idea - where should the Makefile live? in the global-images directory?

global-images/base/requirements.txt
rokroskar: I'm hoping to keep it simple(r) by using pip and maybe we can fix the buildpack to let us specify the python version. But yes, I take your point... I can change it to use conda for now.

rokroskar: oh `base` is just an image with basically nothing in it. I could call it `basic` I suppose... or do you have other naming suggestions? I don't really expect others to build on top of it, though it wouldn't be impossible.

global-images/README.md
olevski: Yeah I think it makes sense if it is in the global-images directory.

global-images/base/requirements.txt
olevski: I dont have an issue with the name. I am just not sure how do we expect the image to be used. 

As for being able to pick the python version. That is not that simple to fix because it simply comes from the underlying python buildpack. And I am not sure there is a way to specify the version in our buildpack for some cases (i.e. pip) and then use the original buildpack logic for other cases (poetry, conda).

rokroskar: I renamed it to "basic" and used conda so we can have python 3.12

rokroskar:
@olevski @leafty I've addressed the comments: 

- cleaned up the GHA file
- added a Makefile
- tag latest only on merges to main branch
- added caching (we can remove this if it creates too much noise... cuts down build times by ~30 seconds)

leafty (dismissed):
LGTM

olevski (approved):

PR #4089, opened 2025-07-11
feat: make swagger-ui image configurable

sgaist:
## Describe your changes

Change the swagger template so that the image can also be configured when deploying Renku.

This allows to customize swagger in the same way as other components.

Useful for example when deploying on OpenShift where the default image cannot run due to security constraints.


olevski:
@sgaist please target branch `release-2.5.0`

olevski:
Everything else looks good.

olevski (approved):

PR #4092, opened 2025-07-15
test: update the Cypress tests to support changes from the new group search page

lorenzo-cavazzi:
# WIP

/deploy renku-ui=lorenzo/group-search


lorenzo-cavazzi:
Integrated into https://github.com/SwissDataScienceCenter/renku/pull/4100

PR #4104, opened 2025-07-23
release 2.6.0

RenkuBot:
Release 2.6.0

This PR is auto-generated by [actions/github-script](https://github.com/actions/github-script).

/deploy #legacy


leafty:
Something weird is going on with the diff page of the PR. The values file has been changed, some components have been bumped up a version but this is not showing up in the diff page.

olevski (approved):

helm-chart/renku/values.yaml
```diff
@@ -741,14 +741,15 @@ ui:
       image:
         repository: renku/renku-ui
         tag: "0.11.16"
+    supportLegacySessions: true
```
rokroskar: shouldn't this now be false by default?

olevski: Probably yes. I will change it.

olevski: Done. I just pushed the change directly here since it was a simple one-liner

PR #4107, opened 2025-07-31
feat: add docusaurus docs

olevski:
This does a few things:
- moves the legacy docs to a dedicated folder called `docs-legacy`, we can remove this fully later
- sets up docusaurus for the new docs
- adds a link to the user docs on notion for the user specific docs
- adds content for admin specific documentation
- updates the read the docs config so that it works with docusaurus - this is according to the documentation from read the docs

So when we merge this our read the docs page should switch to these new docs.

To start a development server for the docs and browse them simply run `cd docs && npm run start`.


.readthedocs.yaml
```diff
@@ -2,16 +2,14 @@ version: 2
...
-sphinx:
-  configuration: docs/conf.py
-
-python:
-  install:
-  - requirements: docs/requirements.txt
-
-submodules:
-  include: all
+    nodejs: "18"
```
leafty:
```suggestion
    nodejs: "22"
```

Version 18 is end of life, current LTS is 22.

docs-legacy/requirements.txt
leafty, line 1: This file needs to be updated to match the move to `docs-legacy`.

.github/workflows/pull-request-test.yml
leafty, line 1: We should also test the `docs` build and if possible provide a way to preview the docs.

.readthedocs.yaml
```diff
@@ -2,16 +2,14 @@ version: 2
...
+  jobs:
+    install:
+       - cd docs/ && npm install
```
leafty: Use `clean-install` for production builds (will install dependencies from the lockfile).

```suggestion
       - npm --prefix docs clean-install
```

.readthedocs.yaml
```diff
@@ -2,16 +2,14 @@ version: 2
...
+    build:
+      html:
+        # Build the site
+        - cd docs/ && npm run build
```
leafty:
```suggestion
        - npm --prefix docs run build
```

.github/workflows/pull-request-test.yml
olevski: Yes I am replacing the old test for the old legacy with this.

olevski:
> Should we add the same theming as for https://blog.renkulab.io/? It should be relatively OK to copy the theme files.

I will open a follow up pr for that right after this. We should def have the same theming.

.github/workflows/pull-request-test.yml
olevski: For preview I dont have a nice way now. For now we can just run things locally. I will add this in a followup.

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,157 @@
...
+        {
+          title: 'Docs',
+          items: [
+            {
+              label: 'For users',
+              to: '/docs/users',
+            },
+            {
+              label: 'For admins',
+              to: '/docs/admins',
```
leafty: Links to a 404 page, as the admin category doesn't have an index page.

This breaks the build step (returns a non-zero exit code).

```suggestion
              to: '/docs/admins/architecture/services',
```

docs/docs/02-admins/01-architecture/01-services.md
```diff
@@ -0,0 +1,69 @@
...
+Unless specifically stated every service below is either a Deployment or StatefulSet
+in Kubernetes.
+
+### Developed in-house
+
+- **Gateway**: Login, routing and injecting Keycloak access tokens.
+- **Data service**: The main backend service for Renku.
+- **Secret service**: Stores and manipulates user secrets.
+- **K8s watcher**: Watches Kubernetes events for user sessions and populates the information in Postgres.
+- **UI**: Serves static assets for the user interface.
```
leafty: Not anymore now that we can do SSR which means that the UI is not made of only static assets now.

olevski: I was trying to say that it is just a file server essentially. Is that not true? I guess it has the few endpoints that serve the configuration the ui needs. So I can say it is "a simple express  server for configuration and frontend assets". Is that better?

docs/docusaurus.config.ts
olevski: Huh it is strange that that works without a problem with `npm run`. I will change it.

docs/docs/02-admins/01-architecture/01-services.md
leafty: In the future we will have SSR, e.g. so that posting a link to a project in slack would unfurl with the project's description.

I would say: "The web server for the Renku user interface."

docs/docusaurus.config.ts
olevski: Ah no nvm. I was trying a different link - never tried to use the footer links.

.github/workflows/pull-request-test.yml
```diff
@@ -18,39 +18,41 @@ concurrency:
...
       - uses: actions/checkout@v4.2.2
-      - uses: actions/setup-java@v4
+      - uses: actions/setup-node@v4
         with:
-          distribution: "temurin"
-          java-version: "17"
-      - uses: actions/setup-python@v5
-        with:
-          python-version: 3.9
+          cache: npm
```
leafty, line 49:
```suggestion
          node-version: "22"
          cache: npm
```

leafty (approved):
LGTM

PR #4113, opened 2025-08-05
chore: do not use RENKUBOT_KUBECONFIG for Azure

olevski:
On Azure CI deployments we can just write the kubeconfig to a file and set the standard KUBECONFIG env variable to point to the file. That way all tools and libraries will use the correct KUBECONFIG.

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -120,8 +120,8 @@ jobs:
       - name: Get AKS credentials
         if: needs.check-deploy.outputs.deploy == 'true'
         run: |
-          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ runner.temp }}/renkubot-kube.config"
-          chmod 400 "${{ runner.temp }}/renkubot-kube.config"
+          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ github.workspace }}/renkubot-kube.config"
+          chmod 400 "${{ github.workspace }}/renkubot-kube.config"
 
       - name: renku build and deploy
         if: needs.check-deploy.outputs.deploy == 'true'
@@ -130,7 +130,7 @@ jobs:
           DOCKER_PASSWORD: ${{ secrets.RENKU_DOCKER_PASSWORD }}
           DOCKER_USERNAME: ${{ secrets.RENKU_DOCKER_USERNAME }}
           GITLAB_TOKEN: ${{ secrets.DEV_GITLAB_TOKEN }}
-          KUBECONFIG: "${{ runner.temp }}/renkubot-kube.config"
+          KUBECONFIG: "${{ github.workspace }}/renkubot-kube.config"
           RENKU_RELEASE: ci-renku-${{ github.event.number }}
           RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
           RENKU_VALUES: minimal-deployment/minimal-deployment-values.yaml
@@ -173,12 +173,12 @@ jobs:
       - name: Get AKS credentials
         if: needs.check-deploy.outputs.deploy == 'true'
         run: |
-          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ runner.temp }}/renkubot-kube.config"
-          chmod 400 "${{ runner.temp }}/renkubot-kube.config"
+          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ github.workspace }}/renkubot-kube.config"
+          chmod 400 "${{ github.workspace }}/renkubot-kube.config"
 
       - uses: SwissDataScienceCenter/renku-actions/test-renku@v1.19.0
         with:
-          kubeconfig: "${{ runner.temp }}/renkubot-kube.config"
+          kubeconfig: "${{ github.workspace }}/renkubot-kube.config"
           renku-release: ci-renku-${{ github.event.number }}
           gitlab-token: ${{ secrets.DEV_GITLAB_TOKEN }}
           s3-results-access-key: ${{ secrets.ACCEPTANCE_TESTS_BUCKET_ACCESS_KEY }}
@@ -305,8 +305,8 @@ jobs:
       - name: Get AKS credentials
         if: needs.check-deploy.outputs.deploy == 'true'
         run: |
-          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ runner.temp }}/renkubot-kube.config"
-          chmod 400 "${{ runner.temp }}/renkubot-kube.config"
+          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ github.workspace }}/renkubot-kube.config"
+          chmod 400 "${{ github.workspace }}/renkubot-kube.config"
 
       # Cleanup Azure deployments
       - name: renku teardown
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-test.yml
```diff
@@ -123,13 +123,9 @@ jobs:
           az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ github.workspace }}/renkubot-kube.config"
           chmod 600 "${{ github.workspace }}/renkubot-kube.config"
```
sambuc: I would use `400` (no need to have write access on that file afaik) and move it to `runner.temp` as suggested by flora elsewhere

.github/workflows/pull-request-test.yml
```diff
@@ -123,13 +123,9 @@ jobs:
           az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ github.workspace }}/renkubot-kube.config"
```
sambuc, line 123: I would move it to `runner.temp` as suggested by flora elsewhere

.github/workflows/pull-request-test.yml
```diff
@@ -138,7 +134,6 @@ jobs:
           RENKU_RELEASE: ci-renku-${{ github.event.number }}
```
sambuc, line 134: I would move it to runner.temp as suggested by flora elsewhere

.github/workflows/pull-request-test.yml
```diff
@@ -185,7 +180,7 @@ jobs:
           cat "${{ github.workspace }}/renkubot-kube.config" >> $GITHUB_ENV
           echo "EOF" >> $GITHUB_ENV
 
-      - uses: SwissDataScienceCenter/renku-actions/test-renku@v1.18.2
+      - uses: SwissDataScienceCenter/renku-actions/test-renku@v1.19.0
         with:
           kubeconfig: ${{ needs.check-deploy.outputs.deploy == 'true' && env.RENKUBOT_KUBECONFIG || secrets.RENKUBOT_DEV_KUBECONFIG }}
```
sambuc: here as well

.github/workflows/pull-request-test.yml
```diff
@@ -185,7 +180,7 @@ jobs:
           cat "${{ github.workspace }}/renkubot-kube.config" >> $GITHUB_ENV
           echo "EOF" >> $GITHUB_ENV
```
sambuc: This needs to be adapted as well afaik

.github/workflows/pull-request-test.yml
```diff
@@ -317,17 +312,12 @@ jobs:
           az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ github.workspace }}/renkubot-kube.config"
```
sambuc, line 308: Here as well, I would go with `400` and `runner.temp`

.github/workflows/pull-request-test.yml
```diff
@@ -317,17 +312,12 @@ jobs:
...
-      # Cleanup for both standard and Azure deployments
+      # Cleanup Azure deployments
       - name: renku teardown
-        uses: SwissDataScienceCenter/renku-actions/cleanup-renku-ci-deployments@v1.18.2
+        uses: SwissDataScienceCenter/renku-actions/cleanup-renku-ci-deployments@v1.19.0
         env:
           HELM_RELEASE_REGEX: "^ci-renku-${{ github.event.number }}$"
           GITLAB_TOKEN: ${{ secrets.DEV_GITLAB_TOKEN }}
-          RENKUBOT_KUBECONFIG: ${{ needs.check-deploy.outputs.deploy == 'true' && env.RENKUBOT_KUBECONFIG || secrets.RENKUBOT_DEV_KUBECONFIG }}
+          KUBECONFIG: "${{ github.workspace }}/renkubot-kube.config"
```
sambuc, line 317: here as well for `runner.temp`

.github/workflows/renku-dev-test.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
       github.event.client_payload.message == 'Helm test succeeded' }}
     runs-on: ubuntu-24.04
     steps:
-      - uses: SwissDataScienceCenter/renku-actions/test-renku@v1.18.2
+      - uses: SwissDataScienceCenter/renku-actions/test-renku@v1.19.0
         with:
           kubeconfig: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
```
sambuc: Should this not be adapted somehow?

sambuc (commented):
There are some spots which I believe should also be adapted, plus we can go about moving the file to `runner.temp` as well as making it read only.

.github/workflows/renku-dev-test.yaml
olevski: I think we can remove that whole workflow. We run on every PR. So that is enough.

That triggers on `repository_dispatch` and that is a webhook for something that happens outside of github. I dont even know if we trigger that at all.
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#repository_dispatch

Yeah I just checked. There are zero runs for that workflow.
https://github.com/SwissDataScienceCenter/renku/actions/workflows/renku-dev-test.yaml

olevski: I will remove this fully.

.github/workflows/pull-request-test.yml
olevski: This shows `RENKU_RELEASE` as what your comment refers to. I think you are talking about this though?
```
RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
```

I honestly dont know what that refers to. I dont want to break the values file any more than it is. So I would go back to this. The goal is to just get the tokens to not leak in the logs.

.github/workflows/pull-request-test.yml
```diff
@@ -120,25 +120,20 @@ jobs:
       - name: Get AKS credentials
         if: needs.check-deploy.outputs.deploy == 'true'
         run: |
-          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ github.workspace }}/renkubot-kube.config"
-          chmod 600 "${{ github.workspace }}/renkubot-kube.config"
-
-          echo "RENKUBOT_KUBECONFIG<<EOF" >> $GITHUB_ENV
-          cat "${{ github.workspace }}/renkubot-kube.config" >> $GITHUB_ENV
-          echo "EOF" >> $GITHUB_ENV
+          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev --file "${{ runner.temp }}/renkubot-kube.config"
```
leafty: [we can do this later]
File and env var already setup by `azure/aks-set-context@v4`.
```suggestion
          az aks get-credentials --resource-group renku-dev --name aks-switzerlandnorth-renku-dev
```

olevski: Ok agreed. Lets just get this running.

olevski: I will come back to this.

olevski:
So we need the kubeconfig to survive accross steps... I am not sure if `runner.temp` does this.

Now I see:
```
W0805 14:10:11.729713     135 loader.go:222] Config not found: /home/runner/work/_temp/renkubot-kube.config
Error: Kubernetes cluster unreachable: Get "http://localhost:8080/version": dial tcp [::1]:8080: connect: connection refused
Traceback (most recent call last):
```

olevski:
The problem is not the 400 permissions. It is probably `runnter.temp`. I will rollback to workspace and then do this in a followup.

sambuc (approved):

leafty (approved):
LGTM, merge into the default branch then rebase all PRs 🙏

PR #4115, opened 2025-08-06
chore: use runner.temp for saving kubeconfig

olevski:
/deploy


olevski:
It seems that scripts within an action cannot access the stuff from `runner.temp` context. If I try to do something from the current workflow things work. But if I need to use something in `runner.temp` from the scripts in any of the `renku-actions` then it does not work.

olevski:
> See: https://github.com/actions/runner/issues/965

> When running a "docker" action, the container is started with mounts that move ${{ runner.temp }} to another path (/home/runner/work/_temp":"/__w/_temp). The action should correctly set ${{ runner.temp }} inside the action but existing env vars will of course not be updated to match the change.

I just saw this but I poked around and saw the same thing.

leafty:
Wrote: https://github.com/orgs/community/discussions/168949

PR #4117, opened 2025-08-06
chore: add azure login to deploy action

olevski:
/deploy


olevski:
Does not work because the azure login github action has more magic for logging in than what is available in the azure python sdk or in the azure cli. This is included in javascript libraries that we cannot access easily from the shell or python code in our custom renku-deploy action.

PR #4119, opened 2025-08-07
feat: prepare Legacy sessions phase out

lorenzo-cavazzi:
Add a new value `ui.client.supportLegacySessions` to turn off the Legacy sessions on the UI side -- see [lorenzo/v1-sessions-warnings](https://github.com/SwissDataScienceCenter/renku-ui/pull/3789)

This PR includes the latest UI version 4.6.0 with a bunch of other minor improvements and bugfixes

/deploy extra-values=ui.client.supportLegacySessions=false

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -565,7 +565,7 @@ ui:
     replicaCount: 1
     image:
       repository: renku/renku-ui
-      tag: "4.5.0"
+      tag: "4.6.0"
       pullPolicy: IfNotPresent
       ## Optionally specify an array of imagePullSecrets.
       ## Secrets must be manually created in the namespace.
@@ -749,7 +749,7 @@ ui:
     keepCookies: []
     image:
       repository: renku/renku-ui-server
-      tag: "4.5.0"
+      tag: "4.6.0"
       pullPolicy: IfNotPresent
     imagePullSecrets: []
     nameOverride: ""
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -741,6 +741,7 @@ ui:
       image:
         repository: renku/renku-ui
         tag: "0.11.16"
+    supportLegacySessions: true
   server:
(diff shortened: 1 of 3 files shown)


helm-chart/renku/values.yaml
```diff
@@ -741,6 +741,7 @@ ui:
       image:
         repository: renku/renku-ui
         tag: "0.11.16"
+    supportLegacySessions: true
```
leafty, line 744: Why is this a parameter on the UI only? Sessions may still be started by calling the API if the value is not consumed on the API.

lorenzo-cavazzi: My understanding is that the Product doesn't think it's worth the effort to update the backend services, disabling the UI is all that's planned. I don't have a _solid_ reference, only https://www.notion.so/renku/Disable-Legacy-sessions-21e0df2efafc80fc920aed586557b74c?source=copy_link#22d0df2efafc80cc86b2fe82db531aab

leafty: If someone is using `renku-python` as a CLI, then they can still start sessions then. And possibly still create projects.

lorenzo-cavazzi: Yes, but that doesn't seem to be a common case that justifies the necessary backend work to fully disable Legacy session

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,10 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+## Upgrading to Renku 2.6.0
+
++* NEW `ui.client.supportLegacySessions` used to disabled Legacy sessions.
```
leafty:
```suggestion
+* NEW `ui.client.supportLegacySessions` used to disable Legacy sessions.
```

lorenzo-cavazzi: :pray:

PR #4139, opened 2025-08-15
fix: upgrade data services to v0.53.2

olevski:
/deploy


leafty (approved):

leafty:
CPUs can be requested whole again! 🎉

PR #4144, opened 2025-08-19
chore: update frontend-bp to 0.0.8

rokroskar:
Bumps the version of the buildpacks used for global images to 0.0.8.


rokroskar:
ah yep, didn't see that one!

PR #4168, opened 2025-09-03
fix: correct chart template rendering

leafty:
This fixes an issue which happens when the Renku helm chart is deployed with a name not containing the string "renku".

/deploy

### DIFF
--- a/helm-chart/renku/requirements.yaml
+++ b/helm-chart/renku/requirements.yaml
@@ -18,11 +18,11 @@ dependencies:
     condition: enableV1Services
   - name: amalthea
     repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-    version: "0.20.1"
+    version: "0.21.0"
     condition: enableV1Services
   - name: amalthea-sessions
     repository: "https://swissdatasciencecenter.github.io/helm-charts/"
-    version: "0.20.1"
+    version: "0.21.0"
   - name: dlf-chart
     repository: "https://swissdatasciencecenter.github.io/datashim/"
     version: "0.3.9-renku-2"


.github/workflows/pull-request-test.yml
```diff
@@ -134,7 +134,7 @@ jobs:
           DOCKER_USERNAME: ${{ secrets.RENKU_DOCKER_USERNAME }}
           GITLAB_TOKEN: ${{ secrets.DEV_GITLAB_TOKEN }}
           KUBECONFIG: "${{ github.workspace }}/renkubot-kube.config"
-          RENKU_RELEASE: ci-renku-${{ github.event.number }}
+          RENKU_RELEASE: thechosenone # TODO: undo before merging
```
olevski: question. I think you should undo the changes here, right?

.github/workflows/pull-request-test.yml
```diff
@@ -240,7 +240,7 @@ jobs:
           e2e-target: ${{ matrix.tests }}
           kubernetes-cluster-fqdn: ${{ needs.check-deploy.outputs.deploy == 'true' && 'dev.renku.ch' }}
           renku-reference: ${{ github.ref }}
-          renku-release: ci-renku-${{ github.event.number }}
+          renku-release: thechosenone # TODO: undo before merging
```
olevski: same as above.

olevski (changes requested):
Things look good just undo the changes. 

One more question. Why does this have to go straight to master rather than a release branch. I would rather merge it in a release branch and let it go through our regular cycle. Or if you want split the github actions changes (those can go to master directly) and the helm chart changes (these would go to a release branch.

.github/workflows/pull-request-test.yml
leafty: Yes, these need to be undone before merging. These are here to demonstrate that the fix works as expected by deploying with a release name that doesn't contain the word `renku`.

PR #4169, opened 2025-09-03
chore(ci): increase memory for the data service

olevski:
@leafty reported restarts when trying to run migrations from a shell from inside the data service. This should work without restarting for CI deployments.

### DIFF
--- a/minimal-deployment/minimal-deployment-values.yaml
+++ b/minimal-deployment/minimal-deployment-values.yaml
@@ -26,10 +26,10 @@ dataService:
         memory: 200Mi
   resources:
     limits:
-      memory: 700Mi
+      memory: 750Mi
     requests:
       cpu: 50m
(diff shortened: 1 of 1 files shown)


sambuc (approved):

minimal-deployment/minimal-deployment-values.yaml
```diff
@@ -26,10 +26,10 @@ dataService:
         memory: 200Mi
   resources:
     limits:
-      memory: 700Mi
+      memory: 750Mi
```
leafty, line 29: Can we put more here? Are we sure this is enough to run all types of `alembic` commands (sometimes needed after rebasing on top of other migrations).

olevski: I think we already run with some spare capacity. It may be enough. Lets see.

PR #4172, opened 2025-09-09
Feat upgrade data service 0.56.0

olevski:
/deploy

Bumps the data services version. Adds a separate flag in the values file to distinguish whether just the v1 services are shut down or both the v1 services and the internal gitlab is off.

This allows us to run Renku with internal services turned off but with the internal gitlab integration still functioning.


olevski:
Will create PRs so that the gateway follows the `enableInternalGitlab` flag. There are some minor changes needed there so that this works as expected.

PR #4175, opened 2025-09-16
feat: modify helm chart to handle internal gitlab

olevski:
/deploy renku-gateway=feat-work-with-disabled-v1-svcs-but-internal-gitlab extra-values=enableInternalGitlab=true,enableV1Services=false


rokroskar (dismissed):
tested, seems good

helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,15 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+## Upgrading to Renku 2.8.0
+
+* NEW `enableInternalGitlab`, set to `true` by default. Indicates that the Gitlab that comes with legacy (V1) Renku
```
rokroskar, line 10:
```suggestion
* NEW `enableInternalGitlab`, set to `false` by default. Indicates that the Gitlab that comes with legacy (V1) Renku
```

olevski: The value is true by default. Otherwise people upgrading will get a nasty surprise.

rokroskar (approved):

PR #4186, opened 2025-09-18
chore: update docs on resource pools and classes

olevski:


docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
-title: Resource pools and classes
+title: Resource Pools and Classes
 ---
 
-:::warning[Under construction]
-Coming soon
+Resource pools allow Renku administrators to:
+- control access to compute resources
+- assign specific users to specific nodes on the cluster
+- limit how much resources (i.e. memory, CPU and/or GPU) can users consume
```
rokroskar:
```suggestion
- limit how much resources (i.e. memory, CPU and/or GPU) users can consume
```

docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
+- control which users can access which resources
```
rokroskar:
```suggestion
```
says the same as the first item

docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
+5. Click `Add Resource Pool`.
+
+Resource pools offer a lot more options that what is shown in the creation form.
+After you create a resource pool you can change any of these options.
+
+## Resource Classes
+
+Resource classes represent a collection of:
+- CPU
+- memory
```
rokroskar:
```suggestion
- Memory
```

docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
+- GPU
+- Default disk storage
+- Maximum disk storage
+- Kubernetes tolerations
+- Kubernetes node affinities
+
+A resource class is what users select when they start a session and a single resource
```
rokroskar:
```suggestion
A resource class is what users select when they start a session. A single resource
```

docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
+pool can have many resource classes.
+
+:::info
+The node affinities and tolerations can be used to make it so that specific resource
+classes run on specific nodes in the Kubernetes cluster. For this admins need to also
+manually create taints on specific nodes directly in the Kubernetes cluster and add
```
rokroskar:
```suggestion
 create taints on specific nodes directly in the Kubernetes cluster and add
```

docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
+tolerations for these taints in the resource classes. See the Kubernetes documentation on
+[assigning pods to nodes](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) and
+[taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) 
+for more information on taints, tolerations, affinities and how they can be used
+to control scheduling.
+:::
+
+## Quotas 
+
+A resource pool may have a quota. The quota is the total amount of resources (CPU, GPU and memory)
```
rokroskar: I thought it _must_ have a quota? Except for the default pool?

docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
+
+## Idleness Control
+
+In Renku deployments with a lot of users and session churn user sessions can quickly
+accumulate if there is no automated way of tracking whether they are active or not and
+shutting them down.
+
+A renku session is considered idle if both of the following conditions are true:
+- the CPU usage in the session container the user is working in is below 0.300
+- the last request from the user browser to the session is 30 minutes or older
```
rokroskar, line 88: Does this mean that the minimum idleness is 30 minutes?

docs/docs/02-admins/03-operation/07-rps.md
```diff
@@ -1,7 +1,116 @@
...
+
+Amalthea, the Kubernets operator that manages Renku user sessions, periodically
+checks for the 2 conditions mentioned above. Based on the conditions it determines whether 
+each user session is idle or not. If the session is found to be idle repeatedly for a period
```
rokroskar:
```suggestion
each user session is idle or not. If the session is found to be idle for a period 
```

docs/docs/02-admins/03-operation/07-rps.md
olevski: No. The quota is optional on the non-default resource pools. On the default pool you must not have a quota.

docs/docs/02-admins/03-operation/07-rps.md
olevski: So this gets weird only in the first 30 minutes after the session is created. After that it is ok.

If you want to hibernate sessions after lets say 10 minutes of inactivity (any number <30 mins) then if the session is immediately idle after you create it it will get 30 minutes for it to be hibernated.

But after the first 30 minutes in the life of the session pass by then the idleness check will work normally. I.e. 10 minutes of idleness means that for 10 minutes the session has been repeatedly (every minute I think) found to be using less than 0.300 cpus and the last request for the session is 30 minutes or older.

It is much easier to hardcode the criteria (0.300 cpus and the 30 minute age for the requeset) rather than to add them to the CRD and keep checking all potentially incorrect combinations of that and the thresholds.

docs/docs/02-admins/03-operation/07-rps.md
rokroskar: ok maybe that's the case in the API but it's not the case when you make a resource pool in the admin panel. When I create a pool it automatically has a quota.

olevski: Ok I will make that sentence more ambiguous then. Not sure why the admin panel adds quotas by default.

olevski: I changed it to this:

> The quota of a resource pool is the total amount of resources (CPU, GPU and memory)
that all resource classes in the pool can have available and use at any one time for all sessions.

PR #4190, opened 2025-09-24
feat: add default pool service account

olevski:
Also adds the configuration for using remote clusters.

/deploy

### DIFF
--- a/helm-chart/renku/templates/data-service/deployment.yaml
+++ b/helm-chart/renku/templates/data-service/deployment.yaml
@@ -178,6 +178,8 @@ spec:
                 secretKeyRef:
                   name: {{ cat (include "renku.fullname" .) "-gateway" | nospace }}
                   key: oidcClientSecret
+            - name: DEFAULT_RESOURCE_POOL_SESSION_SERVICE_ACCOUNT
+              value: {{ default "" .Values.dataService.defaultResourcePoolSessionServiceAccount | quote }}
             {{- if .Values.dataService.remoteClustersKubeconfigSecretName }}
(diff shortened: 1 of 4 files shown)


helm-chart/renku/templates/data-service/deployment.yaml
```diff
@@ -178,6 +178,12 @@ spec:
                 secretKeyRef:
                   name: {{ cat (include "renku.fullname" .) "-gateway" | nospace }}
                   key: oidcClientSecret
+            - name: LOCAL_CLUSTER_SESSION_SERVICE_ACCOUNT
+              value: {{ default "" .Values.dataService.localClusterSessionServiceAccount | quote }}
```
sgaist: I am wondering whether ` LOCAL_CLUSTER_SESSION_SERVICE_ACCOUNT` should be set if `localClusterSessionServiceAccount` is empty.

While it's explained that setting it to an empty string means that It would not be set on the session, it feels like there's a misconfiguration somehow.

helm-chart/renku/templates/data-service/deployment.yaml
```diff
@@ -178,6 +178,12 @@ spec:
...
+            {{- if .Values.dataService.localClusterSessionServiceAccount }}
```
sgaist, line 181:
```suggestion
            {{- if .Values.dataService.remoteClustersKubeconfigSecretName }}
```

lorenzo-cavazzi (approved):

aledegano (approved):

sgaist (approved):

PR #4192, opened 2025-09-26
build: update renku-actions to v1.19.1

leafty:
Fixes the renku teardown action.


rokroskar:
if it fixes the action maybe merge this to `master`?

PR #4194, opened 2025-09-26
feat: script to register redirects for migrated V1 projects

ciyer:


scripts/generate-redirects-from-migrations/generate-redirects.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import (
+	"fmt"
+)
+
+func main() {
+	fmt.Println("This script creates redirect entries for migrated projects.")
+	fmt.Print("You will be prompted for access tokens for Renku and GitLab, respectively.\n\n")
```
olevski:
```suggestion
	fmt.Print("You will be prompted for access tokens for Renku and GitLab, respectively. Note that both tokens need to have admin rights.\n\n")
```

scripts/generate-redirects-from-migrations/generate-redirect-utils.go
```diff
@@ -0,0 +1,237 @@
...
+func ScanCredentials(name string) ServerCredentials {
+	var token string
+	hostEnvVar := fmt.Sprintf("%s_HOST", strings.ToUpper(name))
+	if envHost := os.Getenv(hostEnvVar); envHost != "" {
+		fmt.Printf("Using %s host from environment variable (%s): %s\n", name, hostEnvVar, envHost)
+		defaultHosts[name] = envHost
+	} else {
+		fmt.Printf("No %s host environment variable (%s) found, using default: %s\n", name, hostEnvVar, defaultHosts[name])
+	}
+	host := defaultHosts[name]
```
olevski: not the end of the world, but it seems strange to insert the custom value in the defaults map so that you can use it later...

Also it could be that we wont be able to find a default value for the host.


```suggestion
	hostEnvVar := fmt.Sprintf("%s_HOST", strings.ToUpper(name))
	var host string
	if envHost := os.Getenv(hostEnvVar); envHost != "" {
		fmt.Printf("Using %s host from environment variable (%s): %s\n", name, hostEnvVar, envHost)
		host = envHost
	} else {
		fmt.Printf("No %s host environment variable (%s) found, using default: %s\n", name, hostEnvVar, defaultHosts[name])
		var hostFound bool
		host, hostFound = defaultHosts[name]
		if !hostFound {
			fmt.Printf("Could not find default host for %s, exiting.\n", name)
			os.Exit(1)
		}
	}
	
```

scripts/generate-redirects-from-migrations/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+In this folder, run
+
+```
+go run .
+```
+
+You will be prompted for authentication tokens for Renku and GitLab. In both cases, admin tokens are required for the script wo work correctly.
+
+For Renku there is no way to create tokens -- the only option is to get the session cookie from the browser.
```
olevski:
```suggestion
You will be prompted for authentication credentials for Renku and GitLab. In both cases, the credentials should have admin privileges.

For Renku we expect value of the `_renku_session` cookie after the user logs in as an administrator in the browser.
```

olevski: Can you please clarify what type of token is accepted for Gitlab. I suppose any kind? Or you only tested with a personal access token? If so just specify the type of token you tested with. We dont have to try to check / confirm that all token types will work.

scripts/generate-redirects-from-migrations/generate-redirects.go
olevski: Please also add the text that explains that for renku we need the `_renku_session` cookie. It should be in the text that prompts for it, not just in the readme. Same for gitlab it should be clear which token should be used.

scripts/generate-redirects-from-migrations/README.md
ciyer: Good point. I only tested with a personal access token for the read_api scope, and I have updated the prompt to clarify this.

The personal access tokens are easy enough to create and manage, so I think this should be sufficient.

scripts/generate-redirects-from-migrations/generate-redirects.go
ciyer: Clarified in 7e2eb71f6cd5df8fffd4e9e2905722a20dd91bb4

scripts/generate-redirects-from-migrations/generate-redirect-utils.go
ciyer: Fixed in 7e2eb71f6cd5df8fffd4e9e2905722a20dd91bb4

olevski (approved):

PR #4216, opened 2025-10-20
chore(docs): update README on how to add image/video to docs

mohammad-alisafaee:
Updates `docs/README` to include some guidelines on how to update docs pages.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,8 +18,8 @@ npm run start
 
 This command starts a local development server and opens up a browser window. Most changes are reflected live without having to restart the server.
 
-Another useful command is `npm run clear` (accompanied with a server restart) which purges the locally 
-cached and built static files when significant changes are made but they do not
+Another useful command is `npm run clear` (accompanied by a server restart) which purges the locally 
+cached and built static files when significant changes are made, but they do not
 show up on the development server.
 
 ## Updating documentation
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -39,3 +39,44 @@ This command generates static content into the `build` directory and can be serv
...
+      max-width: 85%;
+    }
+    ```
+* When adding a video to a page, put it in the same folder as the page. You then need to import it in the markdown file using its relative path and put the imported name in the `src` attribute of the `video` tag. For example:
+    ```md
+    import video10 from './video-10.mp4';
+
+    <video controls width="100%" src={video10} />
+    ```
+* When linking to other documentation pages, always use absolute links (e.g. `/docs/users/sessions/guides/create-environment-with-custom-packages-installed`) instead of relative links (e.g. `../sessions/guides/create-environment-with-custom-packages-installed`) or relative paths (e.g. `../60-sessions/guides/20-create-environment-with-custom-packages-installed.md`). Never use file paths for links. Absolute links are more robust to changes in the folder structure and easier to modify when necessary. You can easily get the absolute link of a page by browsing to it in your browser and copying the URL's path from the browser's address bar.
```
olevski, line 82: I am not so sure about this. If you look at the docusaurus docs here: https://docusaurus.io/docs/markdown-features/links they seem to propose using relative file links at the bottom of the docusaurus docs page.

I think that using absolute links may always point  to the latest version of the docs - not the same one as where the originating page is from.

olevski: If you have added your links like this that is ok. We can change them later. But I am just not sure whether we want to propose this for all the docs.

mohammad-alisafaee: Thanks @olevski for the comment! I think the link you shared refers to absolute file paths (not URL paths), since it mentions the `@site/` alias, which is only used for file assets: https://docusaurus.io/docs/advanced/routing#file-paths-and-url-paths.

According to https://docusaurus.io/docs/docusaurus-core#useBaseUrl, the `/baseUrl/` prefix is automatically added to all absolute paths. Because `/baseUrl/` is locale-, and version-aware, it generates the correct links (i.e., with the right locale and version). Inspecting links in the new docs confirms they’re generated correctly. That said, I agree their documentation on this issue is a bit vague.

I propose we merge this PR and use absolute links for now. We don’t have major documentation differences between versions at the moment, and absolute links are easier to grep and replace later. I’ve created a follow-up issue to investigate further and adjust links and this README if needed: https://github.com/SwissDataScienceCenter/renku/issues/4217. I’ll pick it up when I’m back (or in the next cooldown).

PR #4230, opened 2025-10-30
doc(OpenShift): add full minimal example

sgaist:
## Describe your changes

This PR adds a complete minimal example of values file. By minimal, 

As a drive-by fix, the `sessionIngress` configuration has been updated to make use of `ingressClassName` in place of the deprecated annotation.

### DIFF
--- a/docs/docs/20-admins/20-installation/20-configuration.md
+++ b/docs/docs/20-admins/20-installation/20-configuration.md
@@ -243,6 +243,11 @@ The file above is just an example you will have to modify the options shown as f
 
 - Label the node(s) you want to use for the builds with `renku.io/node-purpose: image-build`
 
+### Build strategy
+
+The last action required to have your system ready is to deploy the [BuildStrategy
+for Shipwright](https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_pack_builder/manifests/buildstrategy.yaml).
+
 ### Configuration without Harbor and Shipwright
 
 This is the default and no further steps are needed.


docs/docs/20-admins/40-openshift/10-setup.md
```diff
@@ -332,3 +332,265 @@ networkPolicies:
...
+      requests:
+        cpu: 50m
+        memory: 250Mi
+  k8sWatcher:
+    resources:
+      limits:
+        memory: 200Mi
+      requests:
+        cpu: 20m
+        memory: 200Mi
```
olevski, line 452: question: @sgaist are all these resources and requests absolutely necessary to run on openshift? I have the impression they are not. And if that is the case I would remove them and mention at the top that admins should put in the right resource requests and limits. And just stick with the absolute minimum of what is required to run on openshift. All else looks good.

olevski (changes requested):
Thanks Samuel this looks really good. Just one small question. I am happy if we merge even without resolving it.

docs/docs/20-admins/40-openshift/10-setup.md
sgaist: These values are based on the minimal example for test/poc deployments.

There's a note at the top of the yaml but I haven't checked whether it appears online.

sgaist: After double checking, I confirm we can see the note.

I also explain in the text above that this is an example to test drive the deployment.

Would you rather like to have only a "production ready" version ?

olevski: No it is fine we can keep them.

docs/docs/20-admins/40-openshift/10-setup.md
```diff
@@ -332,3 +332,263 @@ networkPolicies:
...
+    requests:
+      cpu: 50m
+      memory: 75Mi
+dataService:
+  # added
+  rbac:
+    create: false
+
+  localClusterSessionServiceAccount: renku-amalthea-sessions-scc-handler
+  # end added
```
olevski, line 438: What is the purpose of these `# added` and `# end added` comments that are scattered throughout. I think these are leftovers from some check that we can get rid off?

olevski: There are a few more scattered throughout these 2 are not the only ones.

sgaist: I used them to separate the values that are required for Renku to run on OpenShift.

I thought it would allow readers to compare more easily a values file used to deploy on vanilla kubernetes VS OpenShift.

PR #4236, opened 2025-11-04
feat: deprecate running legacy services

leafty:
This update updates the default helm chart values to not deploy Renku legacy (v1) services. Deprecation notices are also documented; the `enableV1Services` key will be removed in the next Renku release.

### DIFF
--- a/helm-chart/values.yaml.changelog.md
+++ b/helm-chart/values.yaml.changelog.md
@@ -7,8 +7,6 @@ Please follow this convention when adding a new row
 
 ## Upgrading to Renku 2.11.0
 
-enableV1Services: false
-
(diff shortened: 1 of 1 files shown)


helm-chart/values.yaml.changelog.md
```diff
@@ -5,6 +5,15 @@ For changes that require manual steps other than changing values, please check o
 Please follow this convention when adding a new row
 * `<type: NEW|EDIT|DELETE> - *<resource name>*: <details>`
 
+## Upgrading to Renku 2.11.0
+
+enableV1Services: false
+
```
leafty:
```suggestion
```

PR #4242, opened 2025-11-09
test(cypress): improve acceptance tests

mohammad-alisafaee:
/deploy extra-values=enableInternalGitlab=false #slow

### DIFF
--- a/cypress-tests/cypress/e2e/slow/sessions.cy.ts
+++ b/cypress-tests/cypress/e2e/slow/sessions.cy.ts
@@ -104,6 +104,12 @@ describe("Slow Sessions", () => {
       .find("[data-cy=start-session-button]")
       .click();
 
+    // Skip "Session repositories not accessible" warning modal
+    cy.get("[data-cy=session-repositories-modal]", {
+      timeout: TIMEOUTS.long,
+    }).should("be.visible");
+    cy.getDataCy("session-repositories-modal-continue").click();
+
     // Verify the session progress page is shown
     cy.get("[data-cy=session-status-starting]", {
       timeout: TIMEOUTS.long,


cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+      // ? Currently, data connectors newly linked might not appear immediately
+      cy.visit(`/u/${user.username}`);
+      cy.getDataCy("data-connector-box")
+        .find(`[data-cy=data-connector-name]`)
+        .contains(name);
+
+      cy.getDataCy("data-connector-box")
+        .find(`[data-cy=data-connector-name]`)
+        .contains(name)
+        .click();
```
leafty: why is the same test done twice?


```suggestion
      cy.getDataCy("data-connector-box")
        .find(`[data-cy=data-connector-name]`)
        .contains(name)
        .click();
```

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+      cy.getDataCy("data-connector-box")
+        .find(`[data-cy=data-connector-name]`)
+        .contains(name)
+        .should("not.exist");
+    });
+  });
+
+  it("Link an existing data connector to a project", () => {
+    const name = `giab-link-${getRandomString()}`;
+    let slug = "";
```
leafty: Looks like you don't need this `slug` variable.

```suggestion
```

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+    // ? Currently, data connectors newly linked might not appear immediately
+    visitCurrentProject();
+    cy.getDataCy("data-connector-box")
+      .find(`[data-cy=data-connector-name]`)
+      .contains(name);
+
+    cy.getDataCy("data-connector-box")
+      .find(`[data-cy=data-connector-name]`)
+      .contains(name)
+      .click();
```
leafty:
```suggestion
    cy.getDataCy("data-connector-box")
      .find(`[data-cy=data-connector-name]`)
      .contains(name)
      .click();
```

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+
+    cy.getDataCy("data-connector-title").should("be.visible").contains(name);
+
+    // Click the dropdown toggle button to open the menu
+    cy.getDataCy("button-with-menu-dropdown")
+      .eq(1)
```
leafty: ?
```suggestion
```

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+      cy.getDataCy("data-connector-slug-toggle").click();
+      cy.getDataCy("data-connector-slug-input").should("have.value", name);
+      cy.getDataCy("data-connector-edit-update-button").click();
+      cy.getDataCy("data-connector-edit-success").should("be.visible");
+      cy.getDataCy("data-connector-edit-close-button").click();
+
+      // Delete the data connector from the user's namespace page
+      // Note: You can only delete data connectors from the owner's page (user/group page),
+      // not from a project page (where you can only unlink them)
+      cy.visit(`/u/${user.username}`);
```
lorenzo-cavazzi: This isn't wrong, but in general we avoid:
* hardcoded URLs (here is `/u/`): they can change.
* invoking `visit` to anything else than the landing page: we should navigate to the target page through existing links, as a user clicking on the UI would do. P.S. Mind that you might find `/v2` instead of `/` because we previously had a distinction between v1 and v2 home pages.

In this case, you can pick the namespace from the `project-info-card` box. Even better, we can add a reference `data-cy="project-namespace-link"` in the UI so that it's easy to pick the link element here in the tests and click on it.

After that, you can also drop `getUserData().then((user: User) => {` at line 146

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+      cy.getDataCy("data-connector-box")
+        .find(`[data-cy=data-connector-name]`)
+        .contains(name);
+
+      cy.getDataCy("data-connector-box")
+        .find(`[data-cy=data-connector-name]`)
+        .contains(name)
+        .click();
```
lorenzo-cavazzi: As pointed out already, you can do both the check and the click in one single step

```suggestion
      cy.getDataCy("data-connector-box")
        .find(`[data-cy=data-connector-name]`)
        .contains(name)
        .click();
```

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+        .scrollIntoView()
+        .click();
+      cy.getDataCy("data-connector-delete").should("be.visible").click();
+
+      // Confirm deletion by typing the slug
+      cy.getDataCy("delete-confirmation-input").type(name);
+      cy.getDataCy("delete-data-connector-modal-button").click();
+
+      // Verify the data connector is deleted
+      cy.visit(`/u/${user.username}`);
```
lorenzo-cavazzi: Same here

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +140,163 @@ describe("Project resources - work with code, data, environments", () => {
...
+    const name = `giab-link-${getRandomString()}`;
+    let slug = "";
+
+    // Create a data connector in the user namespace (not linked to the project)
+    getUserData().then((user: User) => {
+      slug = name;
+      const dataConnectorIdentifier = `${user.username}/${slug}`;
+
+      // Go to the user's page to create a data connector
+      cy.visit(`/u/${user.username}`);
```
lorenzo-cavazzi: same

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
mohammad-alisafaee: This was already included in the existing test, Add and modify data connectors, and I thought it's necessary for correct functionality: https://github.com/SwissDataScienceCenter/renku/blob/master/cypress-tests/cypress/e2e/v2/projectResources.cy.ts#L110-L122

mohammad-alisafaee: It’s now removed from all occurrences, including the existing test.

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
mohammad-alisafaee: This is the issue I mentioned in yesterday’s standup: there are two dropdown buttons with `data-cy="button-with-menu-dropdown"`, one for the project and one for the data connector. Although the project one is hidden beneath the data connector properties modal and not visible to users, Cypress still tries to click it. Using `eq(1)` was a way to select the one I actually care about.
I’ve now added a dedicated data-cy attribute to `ButtonWithMenuV2` in https://github.com/SwissDataScienceCenter/renku-ui/pull/3907 and replaced this with `getDataCy("data-connector-menu-dropdown")`.

mohammad-alisafaee: It’s a bit confusing to me, since the exact same scenario occurs in the `"Delete a data connector"` test and works as expected there.

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
mohammad-alisafaee: I've opened a PR in the `renku-ui` repo as you've suggested: https://github.com/SwissDataScienceCenter/renku-ui/pull/3907

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
mohammad-alisafaee: Removed, see https://github.com/SwissDataScienceCenter/renku/pull/4242#discussion_r2537151080.

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +136,173 @@ describe("Project resources - work with code, data, environments", () => {
...
+        .contains(name);
+    });
+  });
+
+  it("Unlink a data connector from a project", () => {
+    const name = `giab-link-${getRandomString()}`;
+
+    // Create a data connector not linked to a project
+    getUserData().then((user: User) => {
+      const slug = name;
```
lorenzo-cavazzi: Nitpick: nothing wrong here, but you use `slug` only on the next line, and then use `name` in all the following occurrences. You could drop this line and use `name` as well in the next line

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
```diff
@@ -140,6 +136,173 @@ describe("Project resources - work with code, data, environments", () => {
...
+      const dataConnectorIdentifier = `${user.username}/${slug}`;
```
lorenzo-cavazzi: Comment: since the namespace is stable across the test, instead of using `getUserData().then((user: User) => {` in this and the following tests, you could add a variable on top (E.G. `let dataConnectorNamespace = <string>null;`) and then fill it in with the existing `getUserData().then((user: User) => {` in the `before` section. It would be something similar to what we do for `projectIdentifier`.
This would avoid re-invoking the user API multiple times across these tests. Not a _huge_ optimization to be honest 🙊

cypress-tests/cypress/e2e/v2/projectResources.cy.ts
leafty: Well, you should narrow the element search or use `.first()` (https://docs.cypress.io/api/commands/first). Using `.eq(1)` is very confusing for anyone reading the test.

mohammad-alisafaee: I believe `.first()` doesn’t return the correct dropdown menu, since the one I want is actually the second one. In any case, the best solution is to add a specific `data-cy` marker, which is what I’ve done in the latest revision.

olevski (approved):

PR #4246, opened 2025-11-12
Chart changes to support gitlab redirects

ciyer:
# Summary

- Adds values used to configure the Gitlab redirects in the gateway
- Adds values + ingress template for redirecting requests to the domain that Gitlab is served from

See it in action at https://gitlab.renku-ci-ui-3902.dev.renku.ch/, for example https://gitlab.renku-ci-ui-3902.dev.renku.ch/test/project

### DIFF
--- a/helm-chart/renku/templates/gateway/configmap.yaml
+++ b/helm-chart/renku/templates/gateway/configmap.yaml
@@ -96,7 +96,8 @@ data:
         enabled: {{ .Values.gateway.metrics.enabled }}
         port: {{ .Values.gateway.metrics.port }}
     redirects:
-      enabled: {{ .Values.gateway.redirects.enabled }}
-      renkuBaseUrl: {{ include "renku.baseUrl" . | quote }}
-      redirectedHost: {{ .Values.gateway.redirects.redirectedHost | default "" | quote }}
+      gitlab:
+        enabled: {{ .Values.gateway.redirects.gitlab.enabled }}
+        renkuBaseUrl: {{ include "renku.baseUrl" . | quote }}
+        redirectedHost: {{ .Values.gateway.redirects.gitlab.redirectedHost | default "" | quote }}
 ---
--- a/helm-chart/renku/templates/gateway/gitlab-redirect-ingress.yaml
+++ b/helm-chart/renku/templates/gateway/gitlab-redirect-ingress.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.gateway.redirects.enabled -}}
+{{- if .Values.gateway.redirects.gitlab.enabled -}}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
@@ -8,23 +8,20 @@ metadata:
     chart: {{ template "renku.chart" . }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
+{{- with .Values.gateway.redirects.gitlab.ingress.annotations }}
   annotations:
-    cert-manager.io/cluster-issuer: letsencrypt-production
-    nginx.ingress.kubernetes.io/proxy-body-size: "0"
-    nginx.ingress.kubernetes.io/proxy-buffer-size: 8k
-    nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
-    nginx.ingress.kubernetes.io/use-regex: "true"
-    nginx.ingress.kubernetes.io/rewrite-target: /api/gitlab-redirect/$1
+{{ toYaml . | indent 4 }}
+{{- end }}
 spec:
-  ingressClassName: webapprouting.kubernetes.azure.com
+  ingressClassName: {{ .Values.gateway.redirects.gitlab.ingress.className }}
 {{- if .Values.ingress.tls }}
   tls:
   - hosts:
-    - {{ .Values.gateway.redirects.serveHost }}
+    - {{ .Values.gateway.redirects.gitlab.ingress.host }}
     secretName: gitlab-{{ .Values.global.renku.domain }}-tls
 {{- end }}
   rules:
-    - host: {{ .Values.gateway.redirects.serveHost }}
+    - host: {{ .Values.gateway.redirects.gitlab.ingress.host }}
       http:
         paths:
         - backend:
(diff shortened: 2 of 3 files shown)


helm-chart/renku/templates/gateway/configmap.yaml
```diff
@@ -95,4 +95,8 @@ data:
       prometheus:
         enabled: {{ .Values.gateway.metrics.enabled }}
         port: {{ .Values.gateway.metrics.port }}
+    redirects:
```
olevski, line 98: suggestion: the gateway does and in the future may do other redirects. I would call this `gitlabRedirects` or something that makes it clearer.

helm-chart/renku/templates/gateway/gitlab-redirect-ingress.yaml
```diff
@@ -0,0 +1,37 @@
...
+    heritage: {{ .Release.Service }}
+  annotations:
+    cert-manager.io/cluster-issuer: letsencrypt-production
+    nginx.ingress.kubernetes.io/proxy-body-size: "0"
+    nginx.ingress.kubernetes.io/proxy-buffer-size: 8k
+    nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
+    nginx.ingress.kubernetes.io/use-regex: "true"
+    nginx.ingress.kubernetes.io/rewrite-target: /api/gitlab-redirect/$1
+spec:
+  ingressClassName: webapprouting.kubernetes.azure.com
```
olevski: issue: This needs to be modifiable through the values file. Because if someone is running on another cloud or infra they will need to change it to what their provider has.

I would put an `ignress` field under `redirects` (or `gitlabRedirects`) in the values file  and add this and a few other configs with defaults for what we use.

helm-chart/renku/templates/gateway/gitlab-redirect-ingress.yaml
```diff
@@ -0,0 +1,37 @@
...
+    chart: {{ template "renku.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  annotations:
+    cert-manager.io/cluster-issuer: letsencrypt-production
+    nginx.ingress.kubernetes.io/proxy-body-size: "0"
+    nginx.ingress.kubernetes.io/proxy-buffer-size: 8k
+    nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
+    nginx.ingress.kubernetes.io/use-regex: "true"
+    nginx.ingress.kubernetes.io/rewrite-target: /api/gitlab-redirect/$1
```
olevski: issue: same as above this needs to be in the values file. Please add comments to explain that if someone is using a different ingress controller other than ingress nginx they should "convert" the annotations so they get the same result. I mean this is important only for the annotation that does the rewriting. The other annotations are not that important.

helm-chart/renku/values.yaml
```diff
@@ -1183,6 +1183,10 @@ gateway:
     targetMemoryUtilizationPercentage: 75
     targetCPUUtilizationPercentage: 75
   updateStrategy: {}
+  redirects:
+    enabled: false
+    redirectedHost: gitlab.renkulab.io
+    serveHost: gitlab.renkulab.io
```
olevski: suggestion: Please rename these two fields or add comments to explain what the difference is. `serverHost` can be definitely moved in the `ingress` field I proposed above.

olevski: `redirectHost` is the host the redirects will lead to?

olevski (changes requested):

helm-chart/renku/templates/gateway/gitlab-redirect-ingress.yaml
olevski: I am referring to the `ingressClassName` field here specifically.

helm-chart/renku/values.yaml
ciyer: There is no "redirectHost" field, but there is a "redirect**ed**Host" field, which is the host that gets redirected.

helm-chart/renku/templates/gateway/configmap.yaml
ciyer: Alternatively, should we leave the section as `redirects`, but move the information to a subsection, e.g., `gitlab`? So it will look like:

```yaml
  redirects:
    gitlab:
      enabled: false
      redirectedHost: gitlab.renkulab.io
      ingress:
        className: webapprouting.kubernetes.azure.com
        host: gitlab.renkulab.io
        annotations:
          cert-manager.io/cluster-issuer: letsencrypt-production # Use null as value if not using cert-manager to remove this entry
          nginx.ingress.kubernetes.io/proxy-body-size: "0" # Adjust to a reasonable value for production to avoid DOS attacks.
          nginx.ingress.kubernetes.io/proxy-request-buffering: "off" # Needed if GitLab is behind this ingress
          nginx.ingress.kubernetes.io/proxy-buffer-size: "8k" # Default is 4k, larger size necessary for keycloak
          nginx.ingress.kubernetes.io/use-regex: "true"
          nginx.ingress.kubernetes.io/rewrite-target: /api/gitlab-redirect/$1
```

olevski: That works Sekhar.

ciyer: Fixed in 7ce444aa619ba921c75ab4219d2ed1ee8f256ab7

helm-chart/renku/templates/gateway/gitlab-redirect-ingress.yaml
ciyer: Fixed in 7ce444aa619ba921c75ab4219d2ed1ee8f256ab7

helm-chart/renku/templates/gateway/gitlab-redirect-ingress.yaml
ciyer: Fixed in 7ce444aa619ba921c75ab4219d2ed1ee8f256ab7

olevski (approved):

PR #4254, opened 2025-11-18
test(cypress): various group tests

mohammad-alisafaee:
* Search data connector and project in groups
* Link and unlink group data connector to user project

### DIFF
--- a/cypress-tests/cypress/e2e/v2/groupBasics.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/groupBasics.cy.ts
@@ -266,8 +266,53 @@ describe("Group - create, edit and delete", () => {
     cy.contains(name).should("be.visible");
     cy.contains(project1DataConnectorName).should("be.visible");
 
+    // Move project2 to user's namespace
+    cy.visit("/");
+    cy.getDataCy("groups-container")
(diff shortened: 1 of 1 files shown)


cypress-tests/cypress/e2e/v2/groupBasics.cy.ts
```diff
@@ -101,4 +114,271 @@ describe("Group - create, edit and delete", () => {
...
+    const project1Identifier: ProjectIdentifierV2 = {
+      slug: project1Name,
+      id: null,
+      namespace: groupSlug,
+    };
+    const project2Identifier: ProjectIdentifierV2 = {
+      slug: project2Name,
+      id: null,
+      namespace: groupSlug,
+    };
```
lorenzo-cavazzi, line 132: I would remove the 2nd project. It doesn't add much to the test.

As an alternative, you could use the 2nd project to verify that moving resources away from the group works fine: you could move the project from the group namespace to the user namespace, wait for the indexing, and check on the group search page that the project no longer appears. Up to you :)

mohammad-alisafaee: I did what you've suggested.

PR #4256, opened 2025-11-18
test(cypress): more data connector tests

mohammad-alisafaee:
* Link and unlink data connector from another project
* Transfer data connector from project to group (change namespace)

### DIFF
--- a/cypress-tests/cypress/e2e/v2/projectResources.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/projectResources.cy.ts
@@ -10,6 +10,10 @@ import {
   deleteProjectFromAPIV2,
   getProjectByNamespaceAPIV2,
 } from "../../support/utils/projectsV2.utils";
+import {
+  createGroupIfMissingAPI,
+  deleteGroupFromAPI,
+} from "../../support/utils/group.utils";
 
 const sessionId = ["projectResources", getRandomString()];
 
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (changes requested):
The code works fine!
I suggest the following two changes:

- Move all the data connectors code to its own test (dataConnectorsBasics.cy.ts): we now have enough code to justify having a separate test. It will run slightly faster since tests run in parallel, and understanding what project resource is failing will be more immediate.
- Move the resources creation and deletion to the `before` and `after` sections. Currently, if some tests fail and need to be re-run, there will likely be stale projects/groups left around since the deletion section won't be reached and a new project/group will be created for each attempt. Using the `after` section you will only create one additional project/group, and the `after` section will run even in case of failures

mohammad-alisafaee:
> * Move all the data connectors code to its own test (dataConnectorsBasics.cy.ts): we now have enough code to justify having a separate test. It will run slightly faster since tests run in parallel, and understanding what project resource is failing will be more immediate.
> * Move the resources creation and deletion to the `before` and `after` sections. Currently, if some tests fail and need to be re-run, there will likely be stale projects/groups left around since the deletion section won't be reached and a new project/group will be created for each attempt. Using the `after` section you will only create one additional project/group, and the `after` section will run even in case of failures

As I mentioned in Slack, I'll go ahead and merge this PR to https://github.com/SwissDataScienceCenter/renku/pull/4242 and do all the refactoring there before adding more tests, since it's easier to refactor all tests together. Both of your suggestions sound good, and I'll apply them wherever possible.

PR #4257, opened 2025-11-20
build: fix integrations and show better feedback for code repositories

lorenzo-cavazzi:
/deploy extra-values=enableInternalGitlab=false


olevski:
I would call it `feat: ...` . Since `build` is part of conventional commits but it refers to CI/build system not a sprint/build in the way that we use it.

PR #4263, opened 2025-11-25
tests: fix working dir

lorenzo-cavazzi:
This should prevent some annoying linter warnings.

image

/deploy


lorenzo-cavazzi:
Ouch, OK. Thank you for the heads up. I'll change the settings for my IDE locally then

PR #4267, opened 2025-11-27
feat: add session alerts

wesjdj:
/deploy #notest extra-values=alerts.alertmanager.enabled=true

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -782,7 +782,7 @@ ui:
     keepCookies: []
     image:
       repository: renku/renku-ui-server
-      tag: "4.15.0"
+      tag: "4.16.0"
       pullPolicy: IfNotPresent
     imagePullSecrets: []
     nameOverride: ""


docs/docs/10-users/60-sessions/40-alerts/10-high-memory-usage.md
```diff
@@ -0,0 +1,53 @@
...
+### Longer-Term Solutions
+
+1. Optimise your code:
+   - Process data in smaller chunks
+   - Use generators or iterators instead of loading full datasets
+   - Delete variables you no longer need during execution
+   - Use memory-efficient data types and libraries
+
+2. Request more memory:
+   - Stop your current session
```
olevski: You can switch the resource class by just pausing and resuming. Not sure what you mean by stop. It is probably worht mentioning this because it is much easier than fully removing the sessions and then starting a new one.

docs/docs/10-users/60-sessions/40-alerts/20-session-restarted.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. Optimise memory usage:
+   - Process data in smaller chunks instead of loading everything at once
+   - Use memory-efficient data structures and algorithms
+   - Stream data from disk rather than loading it all into memory
+   - Delete large variables when you're done with them: `del variable` in Python, `rm(variable)` in R
+   - Use generators or iterators for large datasets
+
+2. Request more memory:
+   - If your workload needs more memory, stop your session and restart with a larger resource class, if available
```
olevski: Same as above. It is probably worth it to mention both options that you can pause/swap class/resume OR delete and start a new session with more resources.

docs/docs/10-users/60-sessions/40-alerts/30-low-disk-space.md
```diff
@@ -0,0 +1,61 @@
...
+## Steps to Remedy
+
+### Immediate Actions
+
+1. Push any important changes to your Git repository or external storage to avoid data loss.
+
+2. Identify large files:
+   ```bash
+   # Find the largest files in your home directory
+   du -h ~ | sort -h | tail -20
```
olevski: I assume the alert happens when the volume the session uses starts to fill up, right? If so then this is not mounted at ~. The largest file may not be in the mount directory. But we put the mount directory in the `RENKU_MOUNT_DIR` environment variable. So you can use that in your example commands. That will give more accurate results.


```suggestion
   du -h $RENKU_MOUNT_DIR | sort -h | tail -20
```

docs/docs/10-users/60-sessions/40-alerts/30-low-disk-space.md
```diff
@@ -0,0 +1,61 @@
...
+   - See [Resource Pools and Classes](../30-resource-pools-and-classes.md) for more information
+
+3. Optimise your workflow:
+   - Configure applications to use temporary directories that you can clean regularly
+   - Write outputs to compressed formats when possible
+   - Implement automatic cleanup of intermediate files in your scripts
+
+## Prevention
+
+- Monitor disk usage regularly: `df -h ~`
```
olevski:
```suggestion
- Monitor disk usage regularly: `df -h $RENKU_MOUNT_DIR`
```

docs/docs/10-users/60-sessions/40-alerts/30-low-disk-space.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. Manage data more efficiently:
+   - Use a [Data Connector](../../40-data/00-data.md) to access data from external storage without copying it to your session's disk (e.g., S3, Azure Blob Storage, external data repositories)
+   - Use data streaming instead of downloading entire datasets
+   - Clean up output files regularly
+
+2. Request more disk space:
+   - Stop your current session
+   - When starting a new session, if available, select a resource class with more storage. You will need to manually transfer any necessary files from your old session to the new one.
+   - See [Resource Pools and Classes](../30-resource-pools-and-classes.md) for more information
```
olevski, line 48: So most resource classes have the same default storage amount ~10gb. So if you swap the resource class you will get the same amount of disk space most likely. So the thing to do is to modify the session launcher after stopping the session so that you get more disk space and then starting a new session. If the current resource class has storage that is maxed out then you can switch to a new class and get more storage. Anyhow not sure how easy it is to summarize all of this. But you can just mention that the session launcher should be edited to request more disk space.

olevski: Pausing/resuming a session will not result in getting bigger storage if you requested more storage in the meantime. You have to fully delete and restart.

docs/docs/20-admins/30-operation/100-session-alerts.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Prerequisites
+
+    Prometheus and Alertmanager deployed in your cluster
+    Renku Data Services with the alerts feature enabled
+
+## Setting up Alertmanager Webhook Integration
+
+The following steps configure Alertmanager to send session-related alerts to Renku.
+OAuth2 Authentication
```
olevski: I think this is mean to be a title/section?

```suggestion
## OAuth2 Authentication
```

olevski:
> Doesn't this go to 2.11.0?

Yes I missed that. @wesjdj please target the `release-2.11.0` branch.

> And have the component updates already been merged?

That is coming.

helm-chart/renku/values.yaml
olevski: You have to update this to`4.16.0` @wesjdj.

Your changes for the notifications went out in that release. See here: https://github.com/SwissDataScienceCenter/renku-ui/releases/tag/4.16.0

olevski: Also further below in the values file update the ui-server image to the same version.

olevski: All else is good to go imo.

olevski:
```suggestion
      tag: "4.16.0"
```

This has to change too Wes.

olevski (changes requested):

helm-chart/renku/values.yaml
olevski: Dont know why the diff is weird like this. You want to change the tag of the `renku/renku-ui-server` image to the same version as the ui. Both images are published by the ui.

wesjdj: Updated the version, thanks @olevski

olevski:
Just to wanted to mention that Wes setup the alerts and scraping for this and I could check that an alert for disk usage shows up and then automatically clears!

PR #4268, opened 2025-11-28
chore: use retries for requests in Keycloak init

olevski:
Without retries the job immediately restarts and it takes a long time to get a deployment up and running when the ingress TLS is slowly getting propagated.

/deploy

### DIFF
--- a/scripts/init-realm/init-realm.py
+++ b/scripts/init-realm/init-realm.py
@@ -24,6 +24,8 @@ import logging
 import os
 from typing import Dict, List
 
+import requests
+from requests.exceptions import ConnectionError, SSLError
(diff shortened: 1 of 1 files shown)


aledegano (approved):

olevski:
@aledegano I bumped the wait time for tls and name resolution to 2mins. Then we are down to the jobs restarting with an exponential backoff. Now I reached the limit for certificates from letsencrypt. But it should work. I will rerun on Monday and merge then.

olevski:
Ok the 2 minute timeout for the init-realm job is just enough for the realm to be created with a single instance of the job and not have 3-4 restarts.

PR #4271, opened 2025-12-01
acceptance test for sessions

mohammad-alisafaee:
This PR adds more tests for session launchers. It also moves a test that starts a session to a `/slow` subfolder. ATM, all these slow tests are executed while in future we might want to run them on specific cases.

/deploy extra-values=enableInternalGitlab=false renku-ui=build/acceptance-tests #slow

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -74,7 +74,7 @@ jobs:
     steps:
       - id: deploy-comment
         # TODO: Use proper renku-actions release instead of branch
-        uses: SwissDataScienceCenter/renku-actions/check-pr-description@github-actions-for-acceptance-tests
+        uses: SwissDataScienceCenter/renku-actions/check-pr-description@master
         with:
           pr_ref: ${{ github.event.number }}
 


cypress-tests/cypress/e2e/v2/sessions.cy.ts
```diff
@@ -1,245 +1,287 @@
...
+      cy.getDataCy("session-view-resource-class-edit-button").click();
+
+      // Wait for the dialog to open
+      cy.getDataCy("set-resource-class-header").should("be.visible");
+
+      // Open the resource class dropdown
+      cy.get("#addSessionResourceClass").click();
+
+      // Select the "medium" resource class option
+      cy.contains("medium").filter(":visible").click();
```
mohammad-alisafaee: In lines 228-231/232, I wasn't sure how to handle the selection of a resource class in the dropdown menu since it's not a standard one.

cypress-tests/cypress/e2e/slow/sessions.cy.ts
```diff
@@ -0,0 +1,207 @@
+import { TIMEOUTS } from "../../../config";
```
mohammad-alisafaee, line 1: No need to review this file: This is copied from the existing session tests with no modifications and will be updated in future PRs.

cypress-tests/cypress/e2e/v2/sessions.cy.ts
lorenzo-cavazzi, line 1: This test no longer runs sessions, but instead tests the session launcher. I suggest renaming it to something like `sessionLauncher` to make it more immediate

cypress-tests/cypress/e2e/v2/sessions.cy.ts
lorenzo-cavazzi: Ha, it's indeed a little tricky to navigate through this select component -- we are using `react-select`.
Here a few ways to inspect it with Chrome https://stackoverflow.com/questions/49299042/react-select-how-to-keep-dropdown-open-when-styling-in-inspector

The `contains("medium")` is fine. You could limit it so that it's inside the listbox element (the id is `react-select-2-listbox`, or you can use a data-cy instead of the id to limit the scope). If you want to make it even more bullet-proof, you could check the element has the attribute `role="option"`.
There is also the possibility to pick the specific id for the element, which in this case is `react-select-2-option-0-2`, but I think that's less robust because it depends on the indexing position of the element and that might vary if we add/remove more options.

Image

mohammad-alisafaee: I've added a `get('role="option"')` to it.

cypress-tests/cypress/e2e/v2/sessions.cy.ts
mohammad-alisafaee: As you've suggested, I renamed it to `sessionLauncher.cy.ts`. However, I kept the same `sessions.ts` name for types and utils files since they are shared by both sessions and session launchers.

.github/workflows/pull-request-test.yml
```diff
@@ -244,14 +183,24 @@ jobs:
           renku-release: ci-renku-${{ github.event.number }}
           test-user-password: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
 
+      - uses: SwissDataScienceCenter/renku-actions/test-renku-cypress@v1.19.1
```
mohammad-alisafaee, line 186: The problem with not creating a separate job for slow tests (and just adding them as a step) is that they won’t run in parallel with the other tests and also will be executed for each job. We can work around this while there’s only one slow test, but as we add more in the future, things will quickly get messy.

lorenzo-cavazzi: EDIT: sorry I mixed up the reviews :facepalm: Disregard this comment.
~I'm not sure I follow this.
We have a similar mechanism for the `#legacy` switch. Can't `#slow` work in the same way?~

.github/workflows/pull-request-test.yml
```diff
@@ -69,11 +69,12 @@ jobs:
...
       amalthea-sessions: ${{ steps.deploy-comment.outputs.amalthea-sessions }}
       test-enabled: ${{ steps.deploy-comment.outputs.test-enabled == 'true' }}
-      test-legacy-enabled: ${{ steps.deploy-comment.outputs.test-legacy-enabled == 'true' }}
+      test-slow-enabled: ${{ steps.deploy-comment.outputs.test-slow-enabled == 'true' }}
       extra-values: ${{ steps.deploy-comment.outputs.extra-values }}
     steps:
       - id: deploy-comment
-        uses: SwissDataScienceCenter/renku-actions/check-pr-description@v1.19.1
+        # TODO: Use proper renku-actions release instead of branch
+        uses: SwissDataScienceCenter/renku-actions/check-pr-description@github-actions-for-acceptance-tests
```
lorenzo-cavazzi:
```suggestion
        uses: SwissDataScienceCenter/renku-actions/check-pr-description@master
```

PR #4274, opened 2025-12-05
refactoring

mohammad-alisafaee:
* Use proper types.
* Factor repeated code when possible

/deploy extra-values=enableInternalGitlab=false renku-ui=build/acceptance-tests #slow


olevski (approved):
Thanks Mohammad, the changes look really nice!

PR #4277, opened 2025-12-10
test(cypress): slow tests for running sessions

mohammad-alisafaee:
* Tests for session secrets
* Tests for pausing/resuming sessions
* Github action to run slow session tests (requires kubectl)

### NOTE
The PR deployment action was failing but the deployment itself was fine. A `continue-on-error: true` is added to the deployment action to allow running acceptance tests. MAKE SURE TO REMOVE THIS LINE BEFORE MERGING THIS PR.

/deploy extra-values=enableInternalGitlab=false #slow

### DIFF
--- a/.github/workflows/pull-request-test.yml
+++ b/.github/workflows/pull-request-test.yml
@@ -129,7 +129,6 @@ jobs:
       - name: renku build and deploy
         if: needs.check-deploy.outputs.deploy == 'true'
         uses: SwissDataScienceCenter/renku-actions/deploy-renku@v1.20.0
-        continue-on-error: true
         env:
           DOCKER_PASSWORD: ${{ secrets.RENKU_DOCKER_PASSWORD }}
           DOCKER_USERNAME: ${{ secrets.RENKU_DOCKER_USERNAME }}


.github/workflows/pull-request-test.yml
```diff
@@ -73,8 +73,7 @@ jobs:
       extra-values: ${{ steps.deploy-comment.outputs.extra-values }}
     steps:
       - id: deploy-comment
-        # TODO: Use proper renku-actions release instead of branch
-        uses: SwissDataScienceCenter/renku-actions/check-pr-description@master
+        uses: SwissDataScienceCenter/renku-actions/check-pr-description@v1.20.0
```
olevski, line 76: Can you update all the renku actions to the latest version please. It is better for all to use the same version across rather than different ones in different workflows/steps. I just search for `renku-action` and update.

.github/workflows/pull-request-test.yml
```diff
@@ -130,6 +129,7 @@ jobs:
       - name: renku build and deploy
         if: needs.check-deploy.outputs.deploy == 'true'
         uses: SwissDataScienceCenter/renku-actions/deploy-renku@v1.19.1
+        continue-on-error: true
```
olevski: I am a bit worried about this. I have noticed that the deployment takes a long time to be ready recently because of tls propagation for the ingress. So this will hide that issue and cause the tests to fail. Also another case is that even if the deployment fails catastrophically then this would run the tests afterward - which we should not.

olevski: So lets remove this until we understand why the deploy step was failing.

cypress-tests/cypress/support/utils/sessions.ts
```diff
@@ -124,3 +138,50 @@ export function setSessionLauncherEnvironment(
...
+      description: description,
+    },
+  });
+}
+
+/** Extract the namespace from a session URL. */
+export function getNamespace(url: string): string {
+  // URL format: https://{namespace}.dev.renku.ch/p/{userNamespace}/{projectSlug}/sessions/show/{sessionName}
+  return new URL(url).hostname.split(".")[0];
+}
```
olevski: What if this is running on `dev.renku.ch`? Then the namespace would be "".

Also rename the function to explain this is for the K8s namespace. We use "namespace" for a lot of things. I thought you are referring to something else like the user namespace.

In the renku-actions [here](https://github.com/SwissDataScienceCenter/renku-actions/blob/master/deploy-renku/entrypoint.sh#L4) the namespace is extracted from a dedicated env variable or from the release name. So it is best if you use the same logic and do not extract this from the url. Because then you have to deal with the edge cases of renkulab.io and dev.renku.ch. Actually the best case is to just use the dedicated k8s namespace env variable and not have to guess.

.github/workflows/pull-request-test.yml
mohammad-alisafaee: As I mentioned in the PR description, this will be removed.

olevski: Sorry I missed this.

PR #4297, opened 2025-12-22
chore: update docs for adding envidat data connectors

olevski:


docs/docs/10-users/40-data/guides/00-connect-data/50-connect-data-from-data-repositories.md
```diff
@@ -68,6 +75,15 @@ The DOI is also shown on the “Metadata” tab:
 
 ![Screenshot 2025-05-06 at 14.24.20.png](./connect-data-from-data-repositories-60.png)
 
+## How to find the DOI on EnviDat
+
+1. Navigate to [EnviDat](https://envidat.ch).
+2. Search for the specific dataset or simply click "Explore All" on the home page.
+3. Copy the value next to the fingerprint icon near the top of the page, that is the DOI.
+  In the example below this would be `10.16904/envidat.722`.
```
rokroskar, line 83: I found it easier to just right-click and copy the URL - also works. But this is fine.

olevski: Yeah that works too I just thought the explanation for this would be a bit more verbose.

PR #4285, opened 2025-12-12
Jimena/docs updates part 2

jdupre81:
### Content Updates and Design Adjustments

This Pull Request includes several content enhancements and design modifications across the site:

* **Landing Page Design Tweaks:** Various aesthetic adjustments have been implemented on the main page.
    * **⚠️ Dark Mode Issue (Requires UI Team Review):** The Dark Mode styling is not rendering correctly _only on the Landing Page_ (the rest of the site works as expected).
* **Content and Link Cleanup:** Content across several pages has been reviewed and updated. Specifically, numerous links previously redirecting to Notion pages have been migrated to point to internal documentation.
* **Visuals:** The documentation has been enriched with the addition of images and videos.

###  File Structure Request

Please Note: I am unable to resolve the path error for the following file, which needs to be moved for logical organization:

* **Action Required:** Could you please move the file **"How the CSCS integration works in Renkulab"** (`40-css-integration.md`) from its current location to the target path: `60-sessions/guides/` (under the "Compute Sessions/How to..." section). Thanks in advance

PR #4315, opened 2026-01-14
Add user documentation about the search query

eikek:
This is the generated markdown page about the search query that is also available in the swagger docs.

### DIFF
--- a/docs/docs/10-users/75-search/00-search-query.md
+++ b/docs/docs/10-users/75-search/00-search-query.md
@@ -289,19 +289,11 @@ specific fields, which are:
 - score
 
 Sorting by a field is defined by writing the field name, followed by a
-dash and the sort direction. Multiple such definitions can be
-specified, using a comma separated list. Alternatively, multiple
-`sort:…` terms will be combined into a single one in the order they
-appear.
+dash and the sort direction. Multiple such definitions are not yet
+supported.
 
 Example:
 
 ```
-sort:score-desc,created-asc
-```
-
-is equivalent to
-
-```
-sort:score-desc sort:created-asc
+sort:score-desc
 ```


docs/docs/10-users/75-search/00-search-query.md
```diff
@@ -0,0 +1,512 @@
+<!-- -*- fill-column: 80 -*- -->
```
leafty: Please remove this line, it messes up the docs.

```suggestion
```

See as is:
Image


See without:
Image

docs/docs/10-users/75-search/00-search-query.md
```diff
@@ -0,0 +1,512 @@
...
+
+<!-- CODE:START -->
+<!-- from textwrap import dedent -->
+<!-- from renku_data_services.authz.models import Role, Visibility -->
+<!-- from renku_data_services.search.user_query import * -->
+<!-- from renku_data_services.search.user_query_process import CollapseMembers -->
+<!-- for e in Field: -->
+<!--   print(f"- {e.value}") -->
+<!-- CODE:END -->
+<!-- OUTPUT:START -->
```
leafty: What's up with these comments?

leafty (changes requested):

docs/docs/10-users/75-search/00-search-query.md
```diff
@@ -0,0 +1,512 @@
...
+- created
+- score
+
+<!-- OUTPUT:END -->
+
+Sorting by a field is defined by writing the field name, followed by a
+dash and the sort direction. Multiple such definitions can be
+specified, using a comma separated list. Alternatively, multiple
+`sort:…` terms will be combined into a single one in the order they
+appear.
```
leafty: Important note: multiple sort directions will not work with the UI and will most likely break the search results.

docs/docs/10-users/75-search/00-search-query.md
eikek: Oh really? It is a valid markdown comment 😕. I saw the same for other menu points when running locally (`npm run start`), e.g. when you click on "Code", the sub menu point are like this:
 
Selection_012

That's why I figured it is "normal" when running locally.

docs/docs/10-users/75-search/00-search-query.md
eikek: This is left from generating the manual. I can remove it by hand.

docs/docs/10-users/75-search/00-search-query.md
eikek: It is not breaking in my tests when putting in multiple sort fields. The ui shows some "strange" looking query or it will remove redundant terms.

I can write something else here - or remove the section about sorting completely since we have a dropdown in the ui for it.

docs/docs/10-users/75-search/00-search-query.md
leafty: I did fix that on `master` -> https://github.com/SwissDataScienceCenter/renku/pull/4285/changes/2a3f9e9d098febf99bceebeabe25b03bb9e87b45

eikek: Ah thanks. I just saw your other PR #4316 - I was misguided by these comments thinking they are required for some magic we want and then not breaking things. Always a bit sad when comments do this :-). I removed it.

docs/docs/10-users/75-search/00-search-query.md
leafty: I think we can just leave out the multiple sort directions for now. We can try to update the UI's query parser to at least allow it when typed.

PR #4327, opened 2026-01-16
feat: UI upgrade 4.17.0, data svc 0.67.0

olevski:
This mainly adds the feature for explicit session culling.

/deploy


olevski:
I did a quick test to check that the notification about the session being idle will show up. And that clicking the notification resulted in the hibernation time changing.

PR #4328, opened 2026-01-16
build: bump renku-gateway to v1.9.0

leafty:
Main change: the gateway has Renku legacy support removed.

/deploy


leafty:
@sgaist Note: after this is merged, I think you can try to remove more helm chart values related to legacy/gateway.

olevski (approved):

PR #4330, opened 2026-01-16
chore[docs]: add Algolia search to docs

rokroskar:
Docs generated with this PR: https://docs.renkulab.io/en/algolia-search/

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -213,6 +213,7 @@ const config: Config = {
     metadata: [
       { name: 'docsearch:version', content: rtdVersion },
       { name: 'docsearch:docusaurus_tag', content: `docs-default-${rtdVersion}` },
+      { name: 'docusaurus_tag', content: `docs-default-${rtdVersion}` },
     ],
   } satisfies Preset.ThemeConfig,
 };


docs/docusaurus.config.ts
```diff
@@ -170,6 +170,41 @@ const config: Config = {
       theme: prismThemes.github,
       darkTheme: prismThemes.dracula,
     },
+    algolia: {
```
leafty: I think we should have this config be conditional; I'll check how to do this.
If the env vars are provided, then we can add this section, if not, the search bar should not be activated.

docs/docusaurus.config.ts
```diff
@@ -170,6 +170,41 @@ const config: Config = {
...
+      // The application ID provided by Algolia
+      appId: 'process.env.ALGOLIA_APP_ID',
```
leafty:
```suggestion
      appId: process.env.ALGOLIA_APP_ID,
```

docs/docusaurus.config.ts
```diff
@@ -170,6 +170,41 @@ const config: Config = {
...
+
+      // Public API key: it is safe to commit it
+      apiKey: 'process.env.ALGOLIA_API_KEY',
```
leafty:
```suggestion
      apiKey: process.env.ALGOLIA_API_KEY,
```

docs/docusaurus.config.ts
rokroskar: yes, I thought the same - if you can point out how to do it that would be great

docs/static/dns-verify.html
leafty, line 1: I guess we don't need to keep this file now.

rokroskar:
Hmm not quite - I think we need to figure out what to do about all the documentation versions. For example, I see this: 

image

Each of the hits is from a different version of the docs. Ideally it would only show the ones with the same prefix as the page you're coming from...

rokroskar:
It's not quite ready yet - there is an issue with the metadata tags and contextualized search, which is necessary to show search results relevant to the current version. I've gotten the correct tags to be created by the crawler, but they are not properly propagated in the html of the documentation pages.

leafty:
I think we can close this.

PR #4349, opened 2026-01-29
docs: fix link to Amalthea repository README

leafty:
We removed the `new.README.md` in the amalthea repository.


olevski (approved):
Good catch. Thank you!

PR #4352, opened 2026-02-04
doc: csi-rclone deployment on OpenShift

sgaist:
## Describe your changes

This PR adds a chapter to the OpenShift documentation on the deployment of the SDSC fork of csi-rclone.

### DIFF
--- a/docs/docs/20-admins/40-openshift/20-csi-rclone.md
+++ b/docs/docs/20-admins/40-openshift/20-csi-rclone.md
@@ -115,16 +115,4 @@ driver:
   create: false
 storageClass:
   create: false
-
-csiControllerRclone:
-  rclone:
-    image:
-      repository: renku/csi-rclone
-      tag: v0.5.0
-
-csiNodepluginRclone:
-  rclone:
-    image:
-      repository: renku/csi-rclone
-      tag: v0.5.0
 ```


docs/docs/20-admins/40-openshift/20-csi-rclone.md
```diff
@@ -0,0 +1,130 @@
...
+  rclone:
+    image:
+      repository: renku/csi-rclone
+      tag: v0.5.0
+
+csiNodepluginRclone:
+  rclone:
+    image:
+      repository: renku/csi-rclone
+      tag: v0.5.0
```
olevski: question: Why hardcode the image versions here? This should not be necessary because the default values in each published version will have the right versions for the images. And you really should not touch the image versions. Because you may forget to update them and end up deploying helm chart version 0.6.0 for example with the images still being stet to 0.5.0 when they should have been 0.6.0

sgaist: Good point, I needed to do that locally as I was deploying the csi from my local clone of `csi-rclone` for the OpenShift modification tests and there it is pointing to the original images.

sgaist: Fixed

sgaist:
@olevski All good however I am not allowed to merge on the target branch.

PR #4355, opened 2026-02-06
tests: update data connector tests

lorenzo-cavazzi:
Update to data connectors acceptance tests -- see here https://github.com/SwissDataScienceCenter/renku-ui/pull/4054 and https://github.com/SwissDataScienceCenter/renku-data-services/pull/1233

/deploy renku-data-services=add-deposit-job renku-ui=lorenzo/deposit-support


olevski:
The errors are all like:

```
AssertionError: Timed out retrying after 20000ms: Expected to find element: `[data-cy=data-connector-view]`, but never found it.
```

So it seems that we need a bit more tweaks in the tests.

PR #4359, opened 2026-02-11
chore: Disable Gitlab connection for CI deployments

aledegano:
Note: this is intentionally targeted to master as it modifies the CI, but it's irrelevant for proper releases.

/deploy

### DIFF
--- a/minimal-deployment/minimal-deployment-values.yaml
+++ b/minimal-deployment/minimal-deployment-values.yaml
@@ -31,6 +31,7 @@ dataService:
       cpu: 50m
       memory: 750Mi
   replicaCount: 1
+enableInternalGitlab: false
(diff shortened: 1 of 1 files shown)


minimal-deployment/minimal-deployment-values.yaml
```diff
@@ -39,6 +39,7 @@ gitlab:
 global:
   anonymousSessions:
     enabled: true
+  enableInternalGitlab: false
```
olevski:
```suggestion
enableInternalGitlab: false
```

This is not under `gloabl` it is at the top level of the chart.

aledegano: ah, yes, thank you!

PR #4370, opened 2026-02-19
chore(docs): fix typo in repository data title + minor wording

rokroskar:
Updated the title and content for clarity on connecting to datasets via DOI. Revised the list of supported data repositories.

### DIFF
--- a/docs/docs/10-users/40-data/guides/00-connect-data/50-connect-data-from-data-repositories.md
+++ b/docs/docs/10-users/40-data/guides/00-connect-data/50-connect-data-from-data-repositories.md
@@ -9,8 +9,7 @@ description: Zenodo, Dataverse, etc.
 
 We support data from the following repositories:
 
-- Any instance running [Invenio](https://inveniosoftware.org/) 
-- Any instance running [Dataverse](https://dataverse.org/)
+- Almost any instance running [InvenioRDM](https://inveniosoftware.org/) or [Dataverse](https://dataverse.org/)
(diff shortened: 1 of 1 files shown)


aledegano (dismissed):

docs/docs/10-users/40-data/guides/00-connect-data/50-connect-data-from-data-repositories.md
```diff
@@ -3,19 +3,14 @@ title: Data Repositories
...
-We support several such common backends:
-
-- [Invenio](https://inveniosoftware.org/)
-- [Dataverse](https://dataverse.org/)
-
-In addition to the above we support other data repositories that have custom backends
-or other less-common backends:
+We support data from the following repositories:
 
+- Any instance running [Invenio](https://inveniosoftware.org/)
```
leafty: Technically no. See for example the CERN data repositories having the data on the grid.

Also, some admin can totally mix-and-match invenio building bricks in a way where the API is messed up and we can't extract the correct metadata from it.

leafty: Also, it's InvenioRDM https://inveniosoftware.org/products/rdm/

rokroskar: I know, but I'm not sure we need to get into those details here - should we just say "almost any"?

leafty (approved):

PR #4378, opened 2026-02-27
fix: removal of scala acceptance tests

sambuc:
While removing the tests, the use of a configuration block was not
removed.


leafty:
Thanks! I'm not sure why this issue didn't appear when I deployed with this block still in place.

leafty (dismissed):

aledegano (approved):

PR #4379, opened 2026-02-27
Add environment variables to data-tasks supporting resource request tracking feature

eikek:
For allowing the data tasks to list k8s resources, some env var settings have been copied from the data-service deployment.

See SwissDataScienceCenter/renku-data-services#1207

/deploy

### DIFF
--- a/helm-chart/renku/templates/data-service/deployment_data_tasks.yaml
+++ b/helm-chart/renku/templates/data-service/deployment_data_tasks.yaml
@@ -142,6 +142,11 @@ spec:
             {{- include "certificates.env.grpc" . | nindent 12 }}
           volumeMounts:
             {{- include "certificates.volumeMounts.system" . | nindent 12 }}
+            {{- if .Values.dataService.remoteClustersKubeconfigSecretName }}
+            - name: remote-cluster-kubeconfigs
+              mountPath: "/secrets/kube_configs"
+              readOnly: true
+            {{- end }}
           livenessProbe:
             tcpSocket:
               port: 8001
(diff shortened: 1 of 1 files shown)


olevski (changes requested):
Eike this looks good. But lets do this:
1. Merge the stuff in data services
2. Tag a data services release
3. Update the data services in this repo with the new release
4. Deploy once more here to check everything is ok

Then we merge.

Also I think we should have a switch in the values file to turn this off. I think we discussed something like that. In case things get out of hand and the DB fills up too quickly.

I would even keep this flag turned off by default first and then we flip it to ON by default later.

helm-chart/renku/values.yaml
```diff
@@ -1194,6 +1194,7 @@ dataService:
       tag: "0.69.0"
       pullPolicy: IfNotPresent
     resources: {}
+    enableResourceRequestTracking: "false"
```
olevski:
```suggestion
    enableResourceRequestTracking: false
```

helm-chart/renku/templates/data-service/deployment_data_tasks.yaml
olevski, line 144: You have to add the following here:

```
            {{- if .Values.dataService.remoteClustersKubeconfigSecretName }}
            - name: remote-cluster-kubeconfigs
              mountPath: "/secrets/kube_configs"
              readOnly: true
            {{- end }}
```

Sorry Eike the suggested edits were showing up without any indents so it would turn out wrong.

olevski, line 171: Add the following in volumes:

```
        {{- if .Values.dataService.remoteClustersKubeconfigSecretName }}
        - name: remote-cluster-kubeconfigs
          secret:
            secretName: {{ .Values.dataService.remoteClustersKubeconfigSecretName }}
        {{- end }}
```

PR #4383, opened 2026-03-03
Update 50-remote.md

sambuc:
Document Remote clusters configuration

### DIFF
--- a/docs/docs/20-admins/30-operation/50-remote.md
+++ b/docs/docs/20-admins/30-operation/50-remote.md
@@ -3,16 +3,16 @@ title: Remote Clusters
 ---
 
 :::warning[Under construction]
-Coming soon
+This is a first write-up, we welcome any constructive feedback to improve this documentation.
 :::
 
 The following describes how to setup a cluster to be able to run Renku sessions, and connect it to a Renku web portal.
 
 ## Requirements
 
-### Network Requirements
+### Network
 
-- Ingress from public Internet
+- [ ] Ingress from public Internet
   - [ ] load balancer 6443 to 6443 on all cluster nodes (Kubectl API)
   - [ ] load balancer 443 to 443 on all worker nodes for HTTPS access to the user session (JupiterLab or Visual Studio Code for example)
   - [ ] CertManager to provide valid certificates to the user sessions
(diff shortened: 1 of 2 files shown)


docs/docs/20-admins/30-operation/50-remote.md
```diff
@@ -1,7 +1,216 @@
...
+    --generate-name \
+    --create-namespace \
+    --namespace renku-user-sessions \
+    renku/amalthea-sessions
+  ```
+
+- [ ] CSI-rClone Operator (for remote storage)
+
+  ```bash
+  helm install --set-json='csiNodepluginRclone.tolerations=[{"effect": "NoSchedule", "operator": "Exists"}]' csi-rclone renku/csi-rclone
```
aledegano:
```suggestion
  helm install \
  --create-namespace \
  --namespace csi-rclone \
  --set-json='csiNodepluginRclone.tolerations=[{"effect": "NoSchedule", "operator": "Exists"}]' \
  csi-rclone \
  renku/csi-rclone
```

docs/docs/20-admins/30-operation/50-remote.md
```diff
@@ -1,7 +1,216 @@
...
+  kubectl create priorityclass renku-user-sessions-priority --value=1000 --description="default priority for renku sessions"
+  ```
+
+### Storage classes
+
+#### CSI-driver with automatic provisioning support on your cluster
+
+_If you already have automatic volume provisioning setup in your cluster, you may skip this._
+
+For example, to install cinder-csi:
```
aledegano: Remove the example as it is:
1. Very specific
2. Probably not complete, as it is missing authentication necessary for the node plugin to cinder

Instead, I would suggest to simply list a few, example, storageclasses commonly used (cinder-csi, azure etc...)

docs/docs/20-admins/30-operation/50-remote.md
```diff
@@ -1,7 +1,221 @@
...
+### Taints
+
+- [ ] Taints to differentiate node pools as required
+
+## Remote Connection for RenkuLab Services
+
+- [ ] Dedicated **Namespace** for the user sessions (for example `renku-user-sessions`)
+- [ ] Dedicated **ServiceAccount** (for example `renku-session-manager`) with the following rights
+
+  ```yaml
```
aledegano: This role, along with a SA and the RoleBinding should go in the chart, no reason for the admins to have to create it out-of-band.

docs/docs/20-admins/30-operation/50-remote.md
```diff
@@ -1,7 +1,221 @@
...
+### Priority classes
+
+This will be the priority class(es) we can reference when defining quota requirements / limits.
+
+We recommend to create at least one dedicated priority class for the user sessions.
+
+Here as an example on how to create one, adapt as required:
+
+- ```bash
+  kubectl create priorityclass renku-user-sessions-priority --value=1000 --description="default priority for renku sessions"
```
aledegano: This should go in the helm chart, possibly with opt-out, no need to have admins manage this out-of-band.

docs/docs/20-admins/30-operation/50-remote.md
sambuc: As discussed yesterday, I just removed the example, as this is too specific to the cloud provider, and it makes no sense to document multiple ways of achieving the same things. 

The cloud provider documentation should take care of that, and we won't be able to keep up to date with these anyway.

docs/docs/20-admins/30-operation/50-remote.md
sambuc: Applied in https://github.com/SwissDataScienceCenter/amalthea/pull/1088, and adapted the documentation here.

docs/docs/20-admins/30-operation/50-remote.md
sambuc: Applied in https://github.com/SwissDataScienceCenter/amalthea/pull/1088, and updated the documentation

docs/docs/20-admins/30-operation/50-remote.md
olevski: The problem is that this has very little to do with either Amalthea or the full Renku deployment. So it can go in a separate helm chart. But for now the easiest is to manage it out of band.

docs/docs/20-admins/30-operation/50-remote.md
olevski: Lets just start with managing this out of band. We can adapt it later. With the elimination of CSI rclone things will be easier. And if we add this to the amalthea helm chart then we need to document really well what it is for.

PR #4411, opened 2026-03-23
test KC upgrade

### LINKED ISSUE #4373: Init realm script blows up if any client is modified
https://github.com/SwissDataScienceCenter/renku/blob/fa4aa89b1ba23763f73e00ad00371f38336af489/scripts/init-realm/init-realm.py#L168-L170

The code above should NOT run for clients which do not have a service account (which is all of them except for `renku-data-services` or `webhool-alerts`, etc.). Not sure what was the intention here.

aledegano:
When a modified client is recreated, get_client_service_account_user
was called unconditionally on all clients. This caused the script to
blow up for clients that do not use the client_credentials OAuth flow
and therefore have no service account. Guard the call (and the
subsequent role-assignment logic) behind an oauth_flow check, mirroring
the pattern already used in the 'client not found' branch.

Fixes #4373

/deploy

### DIFF
--- a/helm-chart/renku/requirements.yaml
+++ b/helm-chart/renku/requirements.yaml
@@ -4,7 +4,7 @@ dependencies:
     repository: "oci://harbor.renkulab.io/bitnami-mirror"
     condition: postgresql.enabled
   - name: keycloakx
-    version: 5.0.0
+    version: 7.1.4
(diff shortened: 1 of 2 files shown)


rokroskar:
can we close this @aledegano ?

PR #4419, opened 2026-03-31
feat: use self-minted tokens for sessions and migrate resource pool checks to authz

leafty:
Add new internal features to Renku, including:
* refactor access and refresh token handling in the backend, use internal tokens for sessions
* migrate resource pool authorization checks to authz

Note for @SwissDataScienceCenter/yat: the next release will need sessions to be paused during the upgrade.

/deploy

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -6,6 +6,8 @@
 ## that are tuned for a small-scale deployment. For larger deployments, you will need
 ## to adjust these values to match your cluster's capacity and your expected load.
 
+# FAKE CHANGE
+
 ## Global variables
 ## Shared values/secrets
 global:
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -6,8 +6,6 @@
 ## that are tuned for a small-scale deployment. For larger deployments, you will need
 ## to adjust these values to match your cluster's capacity and your expected load.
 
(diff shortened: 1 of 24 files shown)


helm-chart/renku/values.yaml
```diff
@@ -6,6 +6,8 @@
 ## that are tuned for a small-scale deployment. For larger deployments, you will need
 ## to adjust these values to match your cluster's capacity and your expected load.
 
+# FAKE CHANGE
+
```
```suggestion
```

scripts/platform-init/platform-init.py
```diff
@@ -257,6 +260,50 @@ def init_secret_and_data_service_encryption(config: Config):
             ),
         )
 
+def init_data_service_internal_authentication_secret_key(config: Config):
+    """Initialize symmetric signing key for internal authentication in data service."""
```
olevski, line 264: Nice I am glad someone else is adding stuff to here.

olevski (approved):

scripts/platform-init/platform-init.py
olevski: I mean I wish we did not add anything at all. But since we need to add it this is a good place for. Compared to making an extra init job just for making a secret.

leafty: This is already the "extra init job just for making a secret", so I add mine here too 🤣 .

PR #4429, opened 2026-04-14
feat: add AUTHZ config to k8s-watcher deployment chart

SalimKayal:


leafty:
@SalimKayal this would be superseded by #4419.

SalimKayal:
Ok, let's close this one.

On the side I'll also decouple k8s-watcher from authz.

PR #4430, opened 2026-04-15
docs: remove sitemap when deployed

leafty:
It seems that the sitemap breaks the search crawler, as the sitemap generated by docusaurus does not reference the different versions.

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -9,49 +9,50 @@ const url = new URL(
 );
 const versionSlug = process.env.READTHEDOCS_VERSION;
 
-const algoliaConfig = process.env.ALGOLIA_APP_ID && process.env.ALGOLIA_API_KEY
-  ? ({
-      // The application ID provided by Algolia
-      appId: process.env.ALGOLIA_APP_ID,
+const algoliaConfig =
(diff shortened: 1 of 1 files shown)


ciyer (approved):
👍

andre-code:
Does removing the sitemap affect SEO? Since `ALGOLIA_APP_ID` and `ALGOLIA_API_KEY` are only set in production, this strips `sitemap.xml` from exactly the deployment Google indexes.
If so, it might be worth exploring whether the Algolia crawler config could be adjusted instead, for example by adding explicit startUrls for each RTD version so the sitemap can stay for SEO purposes.

leafty:
> Does removing the sitemap affect SEO? Since `ALGOLIA_APP_ID` and `ALGOLIA_API_KEY` are only set in production, this strips `sitemap.xml` from exactly the deployment Google indexes. If so, it might be worth exploring whether the Algolia crawler config could be adjusted instead, for example by adding explicit startUrls for each RTD version so the sitemap can stay for SEO purposes.

Docusaurus doesn't know about the fact that there are alternative doc versions hosted on readthedocs. The point is to remove the `sitemap.xml` generated by Docusaurus and use the one generated by readthedocs. The sitemap from rtd contains links to each version which is publicly visible.

PR #4431, opened 2026-04-15
test: update data connector tests to support new linking

lorenzo-cavazzi:
Update data connectors tests to support the new search modal when linking/adding data connectors

/deploy

### DIFF
--- a/cypress-tests/cypress/e2e/v2/dataConnectors.cy.ts
+++ b/cypress-tests/cypress/e2e/v2/dataConnectors.cy.ts
@@ -111,17 +111,16 @@ describe("Data Connectors", () => {
     visitCurrentProject();
     cy.getDataCy("project-namespace-link").click();
 
+    // Open the data connector view and delete from the menu
     cy.getDataCy("data-connector-box")
       .find(`[data-cy=data-connector-name]`)
       .contains(dataConnectorName)
-      .click();
-
-    // Open the dropdown menu and click delete
-    cy.getDataCy("data-connector-menu-dropdown")
       .should("be.visible")
-      .scrollIntoView()
       .click();
-    cy.getDataCy("data-connector-delete").should("be.visible").click();
+    cy.getDataCy("data-connector-view").within(() => {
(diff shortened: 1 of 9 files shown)


olevski:
@lorenzo-cavazzi release 2.16.0 was merged please move to #4434

lorenzo-cavazzi:
> @lorenzo-cavazzi release 2.16.0 was merged please move to #4434

Ok :+1: 
I'll keep `master` as the target until the 2.17.0 PR is updated, then I'll swap and we can merge this (the UI has been tagged to this PR is fully ready)

cypress-tests/cypress/e2e/v2/dataConnectors.cy.ts
```diff
@@ -37,13 +40,13 @@ describe("Data Connectors", () => {
         slug: projectSlug,
         visibility: "private",
       }).then((response) => {
-        projectId = response.body.id;
+        projectId = response.body.id ?? "";
```
leafty: why? If the response is not typed, maybe add types.

lorenzo-cavazzi: It seems that `Project` is defined with `id?: string;`. Perhaps it's used for patches?

I would rather not change the function used in all other tests and limit this PR's scope to the UI feature. I pushed this ~https://github.com/SwissDataScienceCenter/renku/pull/4431/commits/f6c2ecba07ac0346c4f15c603478663cdd84ae3e~ https://github.com/SwissDataScienceCenter/renku/pull/4431/commits/61437eb34ad711b907b677ad62d66d508d581b24 and we can change that in a separate PR if we wish

leafty: I fixed the types earlier, so the response.body always has a id field.

leafty (approved):

leafty:
@lorenzo-cavazzi Please merge with an appropriate commit message, e.g. `build: bump renku-ui to v4.22.0`

PR #4444, opened 2026-04-27
fix: Allow config of WebOrigins & RedirectUris

sambuc:
/deploy extra-values=notebooks.oidc.extraRedirectUris=["https://domainA.example.org/\*"\,"https://domainB.example.org/\*"],notebooks.oidc.extraWebOrigins=["https://domainC.example.org/\*"]

### DIFF
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -879,6 +879,8 @@ notebooks:
     tokenUrl:
     authUrl:
     allowUnverifiedEmail: false
+    extraRedirectUris:
+    extraWebOrigins:
   sessionIngress:
     host:
     ## If you want to use the default cluster tls cert, set the flag below to true.
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -879,8 +879,8 @@ notebooks:
     tokenUrl:
     authUrl:
     allowUnverifiedEmail: false
-    extraRedirectUris:
-    extraWebOrigins:
+    extraRedirectUris: '[]'
+    extraWebOrigins: '[]'
   sessionIngress:
     host:
     ## If you want to use the default cluster tls cert, set the flag below to true.
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -879,7 +879,15 @@ notebooks:
     tokenUrl:
     authUrl:
     allowUnverifiedEmail: false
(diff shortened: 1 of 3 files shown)


leafty (commented):
Why is this not added as new parameters to the cluster object? This way we would keep security features tight and not need to weaken security.

sambuc:
This is set on the _keycloak client_, and used by keycloak to valid the authentication request.

The keycloak init-realm scripts (which reset the keycloak client in case of deviation with the expected parameters) are called every time there is a helm update/upgrade.

I went for the quickest and smallest fix as this is causing problems for users in prod _right now_ every time we deploy a release. I also did not want to rewrite the whole process for the keyclaok configuration management on something which is more of a bug fix.

Adapting the current init script should be part of pitch about keycloak and its configuration management in Renku, IMHO. To make it retrieve the config from the PostgreSQL DB, introduces a dependency on the DB to be up and running before you can initialise keycloak. That new / adapted script would have to be self standing to not introduce chicken-and-egg problems.

I also believe we should be a bit more mindful of the runtime inter-service dependencies and make sure they stay an acyclic graph. We could even improve deployment speed by ensuring we have proper `depends_on`-like link between the services to avoid restarts with exponential backoffs.

leafty:
> I went for the quickest and smallest fix as this is causing problems for users in prod _right now_ every time we deploy a release.

What is that issue? Do we have a GitHub issue? Or any other documented logs?

sambuc:
We had the issue on slack, literally yesterday. This is on a project channel, you may not have had access to it. @olevski is also aware of it.

helm-chart/renku/values.yaml
```diff
@@ -879,6 +879,16 @@ notebooks:
     tokenUrl:
     authUrl:
     allowUnverifiedEmail: false
+    ## This value is a yaml string of a json list of URIs strings, i.e.
+    ## '["https://domain-a.example.org/*", "https://domain-b.example.net/*"]'
+    ## Each URI should follow the format expected by Keycloak in the Redirect
+    ## URIs field of a keycloak client.
```
```suggestion
    ## Each URI should follow the format expected by Keycloak in the Redirect
    ## URIs field of a keycloak client.
    ## /!\ Important: these extra URIs must be trusted as they will have access to Renku user access tokens.
```

leafty (commented):
Config change looks OK, though using these new values must be done with caution: all the URIs listed (especially if they contain wildcards) must be trusted.

leafty (approved):

olevski (approved):

PR #4445, opened 2026-04-28
chore: use the git-clone image from amalthea

leafty:
Use the new git-clone image from the amalthea repository (`renku/cloner`).

See: https://github.com/SwissDataScienceCenter/amalthea/pull/1083

Tested with public repos, private repos, empty repos, pause/resume cycles.

/deploy


helm-chart/renku/values.yaml
```diff
@@ -996,9 +996,10 @@ notebooks:
     port: 65480
     healthPort: 65481
   gitClone:
+    # TODO: change the image source here
```
```suggestion
```

helm-chart/renku/values.yaml
```diff
@@ -997,8 +997,8 @@ notebooks:
     healthPort: 65481
   gitClone:
     image:
-      name: renku/git-clone
-      tag: "1.31.1"
+      name: renku/cloner
+      tag: 0.26.2
```
```suggestion
      tag: 0.28.0
```

olevski (approved):

PR #4457, opened 2026-05-15
release 2.17.1

olevski:
/deploy #slow


olevski:
This makes amalthea "drop all capabilities" in user sessions by default. Which helps with a lot of the zero days vulnerabilities we have been seeing.

sambuc (approved):

PR #4459, opened 2026-05-18
feat: add r-basic global image

leafty:
Needs work 😄 

cc @rokroskar


rokroskar:
ah thanks, was about to go do this - looks like maybe the problem is rstudio on arm? I remember there being some specific version dependencies at one point to get this to work

leafty:
> ah thanks, was about to go do this - looks like maybe the problem is rstudio on arm? I remember there being some specific version dependencies at one point to get this to work

Yes, it seems that building for arm64 is not working at the moment.

The other point is that I'm not sure how we want to update the folder structure. Maybe we can have:
```
global-images/py-basic
global-images/py-datascience
global-images/r-basic
global-images/r-datascience
```

Or as an alternative:
```
global-images/python/basic
global-images/python/datascience
global-images/r/basic
global-images/r/datascience
```

In the current WIP here, the structure is a bit "hacky" to shove the "basic R" image into the other Python ones.

rokroskar:
good point - if you go with the second option maybe it's cleaner then in the action to separate out which frontends get used?

PR #4460, opened 2026-05-19
release 2.17.2

olevski:
The secret storage service pod did not have access to the kubeconfigs that would allow it to create secrets in remote clusters.

/deploy extra-values=dataService.remoteClustersKubeconfigSecretName=remote-clusters-kubeconfigs


olevski:
I need to setup a remote cluster connection to test that it works.

olevski:
I added a remote cluster and tested launching a session on the remote cluster in the ci deployment with a saved secret. Now it works as expected.

sambuc (approved):

PR #4465, opened 2026-05-28
chore: bump ui to 4.14.0 and data-services to 0.75.0

lorenzo-cavazzi:
/deploy extra-values=dataService.dataTasks.enableResourceRequestTracking=true

### DIFF
--- a/helm-chart/renku/templates/data-service/rbac_data_tasks.yaml
+++ b/helm-chart/renku/templates/data-service/rbac_data_tasks.yaml
@@ -38,6 +38,10 @@ rules:
       - get
       - list
       - watch
+      {{- if .Values.dataService.dataTasks.enableResourceRequestTracking }}
+      # Patch is needed for the data tasks to pause sessions when usage limits are enforced
+      - patch
+      {{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/helm-chart/renku/values.yaml
+++ b/helm-chart/renku/values.yaml
@@ -549,7 +549,7 @@ ui:
     replicaCount: 1
     image:
       repository: renku/renku-ui
-      tag: "4.23.0"
+      tag: "4.24.0"
       pullPolicy: IfNotPresent
       ## Optionally specify an array of imagePullSecrets.
       ## Secrets must be manually created in the namespace.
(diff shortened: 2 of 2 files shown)


leafty (approved):

olevski:
NOTE: This comment is WIP, I will keep modifying it as I find things.

| Behaviour                                | With enforcement (planned) | With enforcement (actual)                                                                                                           | Without enforcement (planned) | Without enforcement (actual) |
|------------------------------------------|----------------------------|-------------------------------------------------------------------------------------------------------------------------------------|-------------------------------|------------------------------|
| Notifications                            | show notifications         | :x:  cannot see notifications about approaching quota exceedance                                                                                                           | show notifications            | notifications show up        |
| Starting new sessions                    | do not allow               | ok not allowed, error message explains why                                                                                          | allowed                       | allowed                      |
| Auto-pausing of running sessions         | sessions should be paused  | sessions are auto-paused, notification in the iframe says why (see screenshot below)                                                                                                        | sessions should not be paused | ok sessions are not paused   |
| Resume of paused session when over quota | sessions should not resume | :x:  cannot resume sessions but error message does not show up, there is just a toas that says session cannot start, but gives no reason | sessions should resume        | ok sessions resume           |
|                                          |                            |                                                                                                                                     |                               |                              |

An error from the data tasks pod about not being able to hibernate perhaps? That is why the session is not paused.

olevski:
The notification that shows up when your session is auto-paused for exceeding the quota when quota enforcement is enabled:

image

olevski:
There were also problems with running expensive sql queries (which have been optimized a bit) that may be slow in production because we have a lot more data. See https://github.com/SwissDataScienceCenter/renku-data-services/pull/1306

These queries are invoked only in the following cases:
- GET /resource_pools (to list all resource pools)
- data tasks for sending alerts
- starting a session
- resuming a session
- GET /resource_pools/<id>/usage

Removing the usage limits on a resource pool should eliminate some calls.

olevski:
Ok I am just recording all my observations here. I uploaded 5 millioin rows into the resource requests log into this CI deployment. This breaks things. I will submit a fix for the query. Basically we cannot use the view because filtering on the view requires the query that creates the view to run every time - over all rows. The filter does not "penetrate" the view. So I will fully remove the view and just use a single (larger) query with a cte or subquery. This way the filtering by user/resource pool or other criteria is done before the query from the view runs.

olevski:
Lets merge this and I will work on a fix to roll into the `2.18.0` release. The other features in here have been held up long enough.

olevski (approved):

PR #4477, opened 2026-06-05
fix: add self-signed certs to ui deployment

olevski:
Since we are using a full stack framework now the UI deployment is making API calls for the UI. So it needs any self-signed certs just like the ui server.

/deploy


lorenzo-cavazzi (approved):
Lgtm 👍

PR #4486, opened 2026-06-23
chore: Use pull-through cache for CI container images

aledegano:
Create and use specific values file for SDSC Renkulab CI-deployments.

/deploy

### DIFF
--- a/minimal-deployment/minimal-deployment-values.yaml
+++ b/minimal-deployment/minimal-deployment-values.yaml
@@ -31,12 +31,9 @@ dataService:
       cpu: 50m
       memory: 750Mi
   replicaCount: 1
-enableInternalGitlab: false
 enableV1Services: false
 gateway:
   replicaCount: 1
-gitlab:
-  enabled: false
 global:
   anonymousSessions:
     enabled: true
@@ -49,7 +46,6 @@ global:
     sentinel:
       enabled: false
 ingress:
-  className: webapprouting.kubernetes.azure.com
   enabled: true
   hosts:
     - <deployment-FQDN>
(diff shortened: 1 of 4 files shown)


leafty (dismissed):

minimal-deployment/sdsc-azure-ci-deployment-values.yaml
```diff
@@ -0,0 +1,227 @@
...
+enableV1Services: false
+gateway:
+  image:
+    repository: harbor.dev.renku.ch/dockerhub_cache/renku/renku-gateway
+  replicaCount: 1
+gitlab:
+  enabled: false
+global:
+  anonymousSessions:
+    enabled: true
```
leafty: Do we need to redefine these?

aledegano: The default is `false` and we do not set this flag anywhere else, although it's needed for some of the CI tests.
So as it stand yes we need it, perhaps we could make `true` the default but I don't know which ramifications that might have outside of our deployments...

minimal-deployment/minimal-deployment-values.yaml
```diff
@@ -63,65 +62,6 @@ keycloakx:
...
-          # Without this the session pods cannot reach services (Keycloak, data services) on the public URL.
-          - namespaceSelector:
-              matchLabels:
-                kubernetes.io/metadata.name: app-routing-system
-            podSelector:
-              matchLabels:
-                app: nginx
-        ports:
-          - port: 443
-            protocol: TCP
```
olevski, line 105: So with this change we are not using the minimal deployment values? I thought all CI deployments use this value. 

Regardless of whether we use or do not use the values we should not remove this. Because if someone uses these values they will operate in an unsafe way.

leafty (commented):
IMPORTANT NOTICE: this will break CI deployments on all component repositories.

The component repositories will need to use the new values file.

minimal-deployment/sdsc-azure-ci-deployment-values.yaml
leafty: and for `gitlab.enabled=false`? I thought it was already set to `false` by default.

olevski (changes requested):
Why do we need a separate minimal values file now?

leafty:
> IMPORTANT NOTICE: this will break CI deployments on all component repositories.
> 
> See:
> 
> ```yaml
>     steps:
>       - name: Checkout renku repository
>         uses: actions/checkout@v4
>         with:
>           repository: SwissDataScienceCenter/renku
>           sparse-checkout: |
>             minimal-deployment
> ```
> 
> The `sparse-checkout` argument will need to be updated in all component repositories.

🤔 Wait... This should still work, though the CI/CD should be updated.

minimal-deployment/minimal-deployment-values.yaml
olevski: So yeah removing the last part of the netpol for accessing the ingress controller will break ci deployments in other repos. But that is ok we will open prs to switch to the new file.

olevski (dismissed):

minimal-deployment/sdsc-azure-ci-deployment-values.yaml
aledegano: I've removed that key and `internalGitlab` as well.

olevski (approved):

PR #4494, opened 2026-07-01
chore: use dependabot groups for docs/

leafty:
cc @rokroskar : I hope this will help get working updates for react and docusaurus in docs/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,3 +28,11 @@ updates:
     directory: "docs/"
     schedule:
       interval: "weekly"
+    groups:
+      docusaurus:
+        patterns:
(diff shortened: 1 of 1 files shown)


rokroskar:
ah thank you, I had forgotten about this feature!

rokroskar (approved):

rokroskar:
I'm tempted to just close the other docs/ PRs and let dependabot recreate them next time there are updates

leafty:
> ah thank you, I had forgotten about this feature!

I prefer avoiding `*` groups because you just get way too many things at the same time. But then we should tell dependabot about the deps which need to be handled together.

PR #4515, opened 2026-07-08
integrations page image updated

jdupre81:
integrations page image updated


bethcg (approved):

leafty (changes requested):
The previous image showed the drop down menu used to access the integrations page. Now, we have a double screenshot but we don't show where the integrations link is (top-right menu).

PR #4516, opened 2026-07-08
image for side panel data connectors updated

jdupre81:
image for side panel data connectors updated


bethcg (approved):

leafty (changes requested):
The previous image had multiple projects linked, can we get the same result?

PR #4575, opened 2026-08-21
chore: add option to skip CI cleanup on closing PRs

olevski:
/deploy #no-test


wesjdj (approved):

leafty:
I would strongly suggest to use a non-CI namespace, because it is otherwise very easy to not see the label on the PR (is there a label on the namespace?).

PR #4582, opened 2026-09-02
chore: update buildpacks version to 0.10.0

SalimKayal:


olevski:
@SalimKayal has this been tested at all - with or without self-signed cas? Or registries that just use http.

olevski:
@SalimKayal I moved this to 2.22.0. I dont think it is that urgent. But if it is we can make a patch release.

SalimKayal:
> @SalimKayal has this been tested at all - with or without self-signed cas? Or registries that just use http.

I'm not sure I understand the question? What does using pixi has to do with self-signed CAs?

olevski:
I am sorry Salim I got mixed up. This can go in.

olevski (approved):

PR #4586, opened 2026-09-04
feat: add a ssh proxy that can connect to running sessions

eikek:
WIP

---

/deploy #notest extra-values=sshProxy.openSshHostKey=LS0tLS1CRUdJTiBPUEVOU1NIIFBSSVZBVEUgS0VZLS0tLS0KYjNCbGJuTnphQzFyWlhrdGRqRUFBQUFBQkc1dmJtVUFBQUFFYm05dVpRQUFBQUFBQUFBQkFBQUJsd0FBQUFkemMyZ3RjbgpOaEFBQUFBd0VBQVFBQUFZRUFzbTl6ZTM5LzY2K0s1NUxsMTFpY1EzUDc0Um4rdy91dHVyKzNna0RwWGhacDFHbHE2VGoyCjlyd21jWWZIWld0aG5YUU5TMVpMN292Z1hDZXRXS3FCREhidVZGbXZheWVVUW1heHRwZ2dFQTYvN2ovUFdEdHdMZDB6N04KZHlpZjFUU3hzdHQ5OGxVQmpUemUvb3JyenBQeTVtbXBLak13UVVwLy9oWkNpV2dqcVUvRmlsUG41WEJRa2FZcVY1Q1JFeQo2UVhhd0pxTzhOTGZpT3ZHR3lrc0V2cm9EcXp6aHJrR01ZT0pSektJNC9VMXcySU1VcWppZW1xcWRSSDMySGY1NjBxUFZtCldMcU9xSTNKY29xKzJ4ZDlPOVFSSmQ3V09qTjZ0V09PV2hBRU9HYUp1T1dTN3FIdlcwdXd0SDhyUVFjT1FCN1VlM1g2TXIKYXI3b3I4RDdkVmlHUlU3bVpXNjJkc2JxYmNEWE1oc1RtWEMvV2VpdU5hR1M5UXZzUlcwNEllOEhCUEhheEdQYlBRZlZoSgptNmp4bXlsS1pQMWZ4Y3V0VjFWUWl0SmVVcW5QeEk4UjVIRVl4cmhHam9tYVBXRmxyU1Z0UGtoUnU2cUpCdk5oR3VyRzQ4CnZaeUZ2ZEJEVU9LM1BqcGRrWDNZMDhSckR0OVhFcDRxTlVoY292TTNBQUFGZ01sUTRnN0pVT0lPQUFBQUIzTnphQzF5YzIKRUFBQUdCQUxKdmMzdC9mK3V2aXVlUzVkZFluRU56KytFWi9zUDdyYnEvdDRKQTZWNFdhZFJwYXVrNDl2YThKbkdIeDJWcgpZWjEwRFV0V1MrNkw0RnduclZpcWdReDI3bFJacjJzbmxFSm1zYmFZSUJBT3YrNC96MWc3Y0MzZE0relhjb245VTBzYkxiCmZmSlZBWTA4M3Y2SzY4NlQ4dVpwcVNvek1FRktmLzRXUW9sb0k2bFB4WXBUNStWd1VKR21LbGVRa1JNdWtGMnNDYWp2RFMKMzRqcnhoc3BMQkw2NkE2czg0YTVCakdEaVVjeWlPUDFOY05pREZLbzRucHFxblVSOTloMytldEtqMVpsaTZqcWlOeVhLSwp2dHNYZlR2VUVTWGUxam96ZXJWampsb1FCRGhtaWJqbGt1Nmg3MXRMc0xSL0swRUhEa0FlMUh0MStqSzJxKzZLL0ErM1ZZCmhrVk81bVZ1dG5iRzZtM0ExekliRTVsd3Yxbm9yaldoa3ZVTDdFVnRPQ0h2QndUeDJzUmoyejBIMVlTWnVvOFpzcFNtVDkKWDhYTHJWZFZVSXJTWGxLcHo4U1BFZVJ4R01hNFJvNkptajFoWmEwbGJUNUlVYnVxaVFiellScnF4dVBMMmNoYjNRUTFEaQp0ejQ2WFpGOTJOUEVhdzdmVnhLZUtqVklYS0x6TndBQUFBTUJBQUVBQUFHQUFNenJNYlNpdUVYelU1S2ZJVll1YlVjSjdECnFYbi8xekRCRnNHMU43T0liVk5yUzZYQTE3WGlQOThDaHh4b3l5eFdnTjN6YWw1STRRRlptLzkxUGNmRmZVUGVDaEh0UHcKb0ZPaG5IWVdQcnd5TkRscERyWkZqSThmZW9LRTZ3V042QnBaSENnT0hNdmZ0MTdLSm02R2hWZ0o4Y3VTbTluNjFiNGFIcgo3bjNmR3hWV2g0WlUxZnlvLzk1RTlsdVpxd0pjZDI4Y0kvdmVqVVErTDY0b2hPd0U4TDEzVURLcW1OMDQ4SmdKNURhdkpnCnFWMjVQRzJ4aG1jUUdMSlo4RWlwZm01N2R1K3JDampaTWVKeGE3MDl4Z0RJdzcvdUIrNHFUNExwRElnd1Z5aytVc2d4SW0KaVhtSENRbC9vOCt5emVJdk9lMUpDVmVVVFRZVnRNMEhWQ1F4OWlrdGx1LzNpTTh6RWV4WnE1S3VqQmIyc2pBWTlGR29nSwpOSTczOEV0ODFKWFovWDlEMlVnUUpHKzVNaEI5Sk5mVHUvQUhJQUwxZ04ySzF6Ty9odzlpZHFlaHZsZnNqRmpFbWtEYktZCmFBOTExeUV2aWYrcDk2Wlk0Q3loQ2lGOWxIakR5dXJ5U3U5SDA0TEJMQXI1RDZwTmNFdmVHTUdYWUR3QVhUM090UkFBQUEKd0ZZMlF0QWoxbW45UEFqN3EzektXRVQvS0JzMXczYUg3TnYxRHlPQnZZNENJNEhpTnN0N3V4cmplNGJMdXI4UGh6dERSSwpYRDJSWFhHRDlEWjlIaDZhZXZYS2wrR1NJV1dzckphU2ZxVnlZSzRBazVaZUkvU1JEZ0VVUWE0ajdCbXdPc2hMRjBqWDlWClhMN3lHS3ZzQzdvSkdKWjFCWVZ3VU9pTWNOMUJjNHMvSlF4MGo4ZHUzWCt0b2d6TldJQ0RFSTh1Tmk0OEgydVhPcW01ZlIKZS9hWitoMVAvV3pGOUJVQUpUVThMY1lqeTZwdnZiVlN3bERNUUVFWUd5NllONldnQUFBTUVBNWhZZjFDNGJwNFg2dUZrWgpJSjUrVkxkR1N6ZGkwdnpHVmFJQnFhRVh1RXh4NU1ZTmFPc1NhWC9wejNkaW1qa0xqU3RFeGswL3F1T2M3aG4wcldaZlE4ClhjSzNTS1NDYUZNWGNGWVE1aTJHUWx5L2dRbzZYSWxtVjRqNEcrc0tJUUIzUU5TTXd6VnloQmJSalltbTZsQUtCS3lVek8KVEFwcVhoa2tPS25rR3Voam5LajJRa1FqRCtON0hhQVpscTJqL3FDMjdxZUErSzBuNEQ1RXdQSWN2RUxtSEZpY3Rqa2plQwoxZVM3UmhvT2FydFdPNGtzakswOWxUMDI5T3NWTWRBQUFBd1FER2lCNHFKOGpxMmI1RDU1YzQ0ZWpwSi95WE5icktIMEJ1CkNOZW9UL2VQdXBPZlA4N0ljdHdJNlpLNFdaZDlMdjRVc2ZaMTJ4WlBleEt1amlvMHl5NGxzSWpPOWx0VHFWbU1wd0JJV1AKZElYZEFMcXREdmZZZkJFYXByZTlBdWkzdWh1U3NRWXkwTHlTbnI2aFY2L01ET0VHbTluL3RveUpLN20zdi8vbDRibTZzYQpYV3pUdG9qSEMyQzFuNFdKV1FocE4vdXljdlo5T3k2ZUs0Mm9JU3lEQUZPUlpESlRCbjZibXFZd2FYV1JHWnI5bFU1dnBWCjRXblhoUER5UHUyMk1BQUFBS1pXbHJaVUJ3YzJGeVlRRT0KLS0tLS1FTkQgT1BFTlNTSCBQUklWQVRFIEtFWS0tLS0tCg==,sshProxy.targetIdentity=LS0tLS1CRUdJTiBPUEVOU1NIIFBSSVZBVEUgS0VZLS0tLS0KYjNCbGJuTnphQzFyWlhrdGRqRUFBQUFBQkc1dmJtVUFBQUFFYm05dVpRQUFBQUFBQUFBQkFBQUJsd0FBQUFkemMyZ3RjbgpOaEFBQUFBd0VBQVFBQUFZRUFzbTl6ZTM5LzY2K0s1NUxsMTFpY1EzUDc0Um4rdy91dHVyKzNna0RwWGhacDFHbHE2VGoyCjlyd21jWWZIWld0aG5YUU5TMVpMN292Z1hDZXRXS3FCREhidVZGbXZheWVVUW1heHRwZ2dFQTYvN2ovUFdEdHdMZDB6N04KZHlpZjFUU3hzdHQ5OGxVQmpUemUvb3JyenBQeTVtbXBLak13UVVwLy9oWkNpV2dqcVUvRmlsUG41WEJRa2FZcVY1Q1JFeQo2UVhhd0pxTzhOTGZpT3ZHR3lrc0V2cm9EcXp6aHJrR01ZT0pSektJNC9VMXcySU1VcWppZW1xcWRSSDMySGY1NjBxUFZtCldMcU9xSTNKY29xKzJ4ZDlPOVFSSmQ3V09qTjZ0V09PV2hBRU9HYUp1T1dTN3FIdlcwdXd0SDhyUVFjT1FCN1VlM1g2TXIKYXI3b3I4RDdkVmlHUlU3bVpXNjJkc2JxYmNEWE1oc1RtWEMvV2VpdU5hR1M5UXZzUlcwNEllOEhCUEhheEdQYlBRZlZoSgptNmp4bXlsS1pQMWZ4Y3V0VjFWUWl0SmVVcW5QeEk4UjVIRVl4cmhHam9tYVBXRmxyU1Z0UGtoUnU2cUpCdk5oR3VyRzQ4CnZaeUZ2ZEJEVU9LM1BqcGRrWDNZMDhSckR0OVhFcDRxTlVoY292TTNBQUFGZ01sUTRnN0pVT0lPQUFBQUIzTnphQzF5YzIKRUFBQUdCQUxKdmMzdC9mK3V2aXVlUzVkZFluRU56KytFWi9zUDdyYnEvdDRKQTZWNFdhZFJwYXVrNDl2YThKbkdIeDJWcgpZWjEwRFV0V1MrNkw0RnduclZpcWdReDI3bFJacjJzbmxFSm1zYmFZSUJBT3YrNC96MWc3Y0MzZE0relhjb245VTBzYkxiCmZmSlZBWTA4M3Y2SzY4NlQ4dVpwcVNvek1FRktmLzRXUW9sb0k2bFB4WXBUNStWd1VKR21LbGVRa1JNdWtGMnNDYWp2RFMKMzRqcnhoc3BMQkw2NkE2czg0YTVCakdEaVVjeWlPUDFOY05pREZLbzRucHFxblVSOTloMytldEtqMVpsaTZqcWlOeVhLSwp2dHNYZlR2VUVTWGUxam96ZXJWampsb1FCRGhtaWJqbGt1Nmg3MXRMc0xSL0swRUhEa0FlMUh0MStqSzJxKzZLL0ErM1ZZCmhrVk81bVZ1dG5iRzZtM0ExekliRTVsd3Yxbm9yaldoa3ZVTDdFVnRPQ0h2QndUeDJzUmoyejBIMVlTWnVvOFpzcFNtVDkKWDhYTHJWZFZVSXJTWGxLcHo4U1BFZVJ4R01hNFJvNkptajFoWmEwbGJUNUlVYnVxaVFiellScnF4dVBMMmNoYjNRUTFEaQp0ejQ2WFpGOTJOUEVhdzdmVnhLZUtqVklYS0x6TndBQUFBTUJBQUVBQUFHQUFNenJNYlNpdUVYelU1S2ZJVll1YlVjSjdECnFYbi8xekRCRnNHMU43T0liVk5yUzZYQTE3WGlQOThDaHh4b3l5eFdnTjN6YWw1STRRRlptLzkxUGNmRmZVUGVDaEh0UHcKb0ZPaG5IWVdQcnd5TkRscERyWkZqSThmZW9LRTZ3V042QnBaSENnT0hNdmZ0MTdLSm02R2hWZ0o4Y3VTbTluNjFiNGFIcgo3bjNmR3hWV2g0WlUxZnlvLzk1RTlsdVpxd0pjZDI4Y0kvdmVqVVErTDY0b2hPd0U4TDEzVURLcW1OMDQ4SmdKNURhdkpnCnFWMjVQRzJ4aG1jUUdMSlo4RWlwZm01N2R1K3JDampaTWVKeGE3MDl4Z0RJdzcvdUIrNHFUNExwRElnd1Z5aytVc2d4SW0KaVhtSENRbC9vOCt5emVJdk9lMUpDVmVVVFRZVnRNMEhWQ1F4OWlrdGx1LzNpTTh6RWV4WnE1S3VqQmIyc2pBWTlGR29nSwpOSTczOEV0ODFKWFovWDlEMlVnUUpHKzVNaEI5Sk5mVHUvQUhJQUwxZ04ySzF6Ty9odzlpZHFlaHZsZnNqRmpFbWtEYktZCmFBOTExeUV2aWYrcDk2Wlk0Q3loQ2lGOWxIakR5dXJ5U3U5SDA0TEJMQXI1RDZwTmNFdmVHTUdYWUR3QVhUM090UkFBQUEKd0ZZMlF0QWoxbW45UEFqN3EzektXRVQvS0JzMXczYUg3TnYxRHlPQnZZNENJNEhpTnN0N3V4cmplNGJMdXI4UGh6dERSSwpYRDJSWFhHRDlEWjlIaDZhZXZYS2wrR1NJV1dzckphU2ZxVnlZSzRBazVaZUkvU1JEZ0VVUWE0ajdCbXdPc2hMRjBqWDlWClhMN3lHS3ZzQzdvSkdKWjFCWVZ3VU9pTWNOMUJjNHMvSlF4MGo4ZHUzWCt0b2d6TldJQ0RFSTh1Tmk0OEgydVhPcW01ZlIKZS9hWitoMVAvV3pGOUJVQUpUVThMY1lqeTZwdnZiVlN3bERNUUVFWUd5NllONldnQUFBTUVBNWhZZjFDNGJwNFg2dUZrWgpJSjUrVkxkR1N6ZGkwdnpHVmFJQnFhRVh1RXh4NU1ZTmFPc1NhWC9wejNkaW1qa0xqU3RFeGswL3F1T2M3aG4wcldaZlE4ClhjSzNTS1NDYUZNWGNGWVE1aTJHUWx5L2dRbzZYSWxtVjRqNEcrc0tJUUIzUU5TTXd6VnloQmJSalltbTZsQUtCS3lVek8KVEFwcVhoa2tPS25rR3Voam5LajJRa1FqRCtON0hhQVpscTJqL3FDMjdxZUErSzBuNEQ1RXdQSWN2RUxtSEZpY3Rqa2plQwoxZVM3UmhvT2FydFdPNGtzakswOWxUMDI5T3NWTWRBQUFBd1FER2lCNHFKOGpxMmI1RDU1YzQ0ZWpwSi95WE5icktIMEJ1CkNOZW9UL2VQdXBPZlA4N0ljdHdJNlpLNFdaZDlMdjRVc2ZaMTJ4WlBleEt1amlvMHl5NGxzSWpPOWx0VHFWbU1wd0JJV1AKZElYZEFMcXREdmZZZkJFYXByZTlBdWkzdWh1U3NRWXkwTHlTbnI2aFY2L01ET0VHbTluL3RveUpLN20zdi8vbDRibTZzYQpYV3pUdG9qSEMyQzFuNFdKV1FocE4vdXljdlo5T3k2ZUs0Mm9JU3lEQUZPUlpESlRCbjZibXFZd2FYV1JHWnI5bFU1dnBWCjRXblhoUER5UHUyMk1BQUFBS1pXbHJaVUJ3YzJGeVlRRT0KLS0tLS1FTkQgT1BFTlNTSCBQUklWQVRFIEtFWS0tLS0tCg==,sshProxy.targetUser=renku,sshProxy.targetPort=2222,dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user amalthea-sessions=salimkayal/feat/2222-service-in-ssh-sessions renku-data-services=salimkayal/feat/ssh-keys_proxy-api


helm-chart/renku/templates/ssh-proxy/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+          image: "{{ .Values.sshProxy.image.repository }}:{{ .Values.sshProxy.image.tag }}"
+          imagePullPolicy: {{ .Values.sshProxy.image.pullPolicy }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          ports:
+            - name: ssh
+              containerPort: 2222
+              protocol: TCP
+          env:
+            - name: RUST_LOG
```
SalimKayal, line 41: The container never sets `RENKU_SSH_PROXY_LISTEN`, so the proxy falls back to its default `127.0.0.1:2222` (config.rs:131).

helm-chart/renku/templates/ssh-proxy/secret.yaml
```diff
@@ -0,0 +1,19 @@
+{{- $openSshHostKey := .Values.sshProxy.openSshHostKey | b64enc | quote }}
```
SalimKayal: `helm lint` on this branch fails

```bash
❯ helm lint renku/
==> Linting renku/
[ERROR] templates/: renku/templates/ssh-proxy/secret.yaml:1:57
  executing "renku/templates/ssh-proxy/secret.yaml" at <b64enc>:
    invalid value; expected string
```

`sshProxy.openSshHostKey` is nil in values.yaml, and Helm's `b64enc` rejects nil instead of coercing.

helm-chart/renku/templates/ssh-proxy/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+              value: "info,renku_ssh_proxy=debug"
+            - name: RENKU_SSH_PROXY_TARGET_PORT
+              value: {{ .Values.sshProxy.targetPort | quote }}
```
SalimKayal, line 45: With default values, `targetPort | quote` on nil renders `value: ""`. The env var is then present but empty, and clap's `default_value_t = 22` only applies when the env var is absent so it parses `""` as u16 and the process exits immediately.

helm-chart/renku/values.yaml
```diff
@@ -1376,6 +1376,50 @@ dataService:
...
+    targetCPUUtilizationPercentage: 75
+  updateStrategy: {}
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
+  # Trusted proxies configuration
+  # See: https://sanic.dev/en/guide/advanced/proxy-headers.md
+  trustedProxies:
+    proxiesCount: 1
+    realIpHeader: x-real-ip
```
SalimKayal:
```suggestion
```

Don't these only configure Sanic ? shouldn't we remove them ?

helm-chart/renku/values.yaml
```diff
@@ -1376,6 +1376,50 @@ dataService:
...
+  targetPort:
+  sentry:
+    enabled: false
+    dsn:
+    environment:
+    sampleRate: 0.1
+  replicaCount: 1
+  podAnnotations: {}
+  resources: {}
+  autoscaling:
```
SalimKayal: Setting `autoscaling.enabled: true` does nothing: there is no HPA template for ssh-proxy (compare `templates/authz/hpa.yaml`, which does ship one). Enabling it suppresses the `replicas` field while creating no autoscaler.

helm-chart/renku/values.yaml
```diff
@@ -1376,6 +1376,50 @@ dataService:
...
+sshProxy:
+  rbac:
+    create: true
+  capacityReservation:
+    # Creates a PriorityClass with value -1 for placeholder pods so they are evicted
+    # when real user session pods need resources.
+    createPriorityClass: true
+    # Name of an existing PriorityClass to use for placeholder pods when createPriorityClass is false.
+    # If left empty and createPriorityClass is false, placeholder pods will have no priority class assigned.
+    existingPriorityClass: ""
```
SalimKayal: Shouldn't we remove the `RBAC` and `CapacityReservation` until they are actually referenced by a template ?

helm-chart/renku/templates/ssh-proxy/service.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "renku.fullname" . }}-ssh-proxy
+  labels:
+    app: renku-ssh-proxy
+    chart: {{ template "renku.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  annotations:
```
SalimKayal: Shouldn't we remove this until the `:8000/metrics` endpoint exists ?

helm-chart/renku/templates/ssh-proxy/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+              value: {{ .Values.sshProxy.sentry.environment | quote }}
+            - name: SENTRY_SAMPLE_RATE
+              value: {{ .Values.sshProxy.sentry.sampleRate | quote }}
+            - name: RENKU_URL
+              value: {{ (include "renku.baseUrl" .) | quote }}
+          volumeMounts:
+            - mountPath: "/secrets/openSshHostKey"
+              name: open-ssh-host-key
+              readOnly: true
+            - mountPath: "/secrets/internalSecretKey"
```
SalimKayal, line 71: This seems to be read nowhere. What's the goal there ?

helm-chart/renku/templates/ssh-proxy/service.yaml
```diff
@@ -0,0 +1,23 @@
...
+    prometheus.io/port: '8000'
+spec:
+  type: {{ .Values.sshProxy.service.type }}
+  ports:
+    - port: {{ .Values.sshProxy.service.port }}
+      targetPort: 2222
+      protocol: TCP
+      name: ssh
+  selector:
+    app: {{ template "renku.name" . }}-ssh-proxy
```
SalimKayal: `helm-chart/renku/templates/ssh-proxy/deployment.yaml` [R6](https://github.com/SwissDataScienceCenter/renku/pull/4586/changes#diff-c3038198dd24d2b85c66a30b0fb2c01f9ca327100a8c589c530ad51a08ae79afR6), [R18](https://github.com/SwissDataScienceCenter/renku/pull/4586/changes#diff-c3038198dd24d2b85c66a30b0fb2c01f9ca327100a8c589c530ad51a08ae79afR18), [R22](https://github.com/SwissDataScienceCenter/renku/pull/4586/changes#diff-c3038198dd24d2b85c66a30b0fb2c01f9ca327100a8c589c530ad51a08ae79afR22)

and 

`helm-chart/renku/templates/network-policies.yaml` [R659](https://github.com/SwissDataScienceCenter/renku/pull/4586/changes#diff-29043e5bc4d20a6dc93740aec9f9c115815db49be7f90185cf24dcfba4e08049R659), [R685](https://github.com/SwissDataScienceCenter/renku/pull/4586/changes#diff-29043e5bc4d20a6dc93740aec9f9c115815db49be7f90185cf24dcfba4e08049R685)

hardcode `renku-ssh-proxy`. This will fail when `renku.name` is not `renku`

helm-chart/renku/templates/network-policies.yaml
SalimKayal, line 1: Q: should we also open ingress **to** the proxy ?

SalimKayal (approved):
Ok a few things on the helm side.

SalimKayal (changes requested):
oups, approved instead of requesting changes, sorry.

helm-chart/renku/templates/ssh-proxy/deployment.yaml
eikek: It will be needed once we connect to data-services, afaiu. Currently it is not used.

helm-chart/renku/templates/ssh-proxy/secret.yaml
eikek: The key is mandatory. So if it is failing with `nil` that is ok, but perhaps we can make it look better in that case?

eikek: Perhaps we can use this: `{{- $openSshHostKey := (required "sshProxy.openSshHostKey is missing" .Values.sshProxy.openSshHostKey) | b64enc | quote }}`

helm-chart/renku/templates/ssh-proxy/service.yaml
eikek: Yes, I also think it's better. First I thought to leave it as it is probably needed later. But changed my mind. We add it once we get there.

helm-chart/renku/templates/network-policies.yaml
eikek: Yes, I tried that with a very open rule 😅 needs to be revisited as well.

renku-ssh-proxy/src/main.rs
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    let settings = Settings::create()?;
+
+    env_logger::builder()
+        .filter_level(settings.log_level.log_level_filter())
+        .init();
+
+    println!("{} running. Listening on {}", info, settings.listen);
+    let _ = renku_ssh_proxy::serve_proxy(&settings).await;
```
Panaetius, line 25: if this errors, maybe the error should be printed to stderr, not just consumed.

renku-ssh-proxy/src/proxy_server.rs
```diff
@@ -0,0 +1,432 @@
...
+    type Error = color_eyre::eyre::Error;
+
+    async fn check_server_key(
+        &mut self,
+        server_public_key: &PublicKey,
+    ) -> Result<bool, Self::Error> {
+        if let Some(pk) = &self.expected_host_key {
+            Ok(server_public_key == pk)
+        } else {
+            Ok(true) //todo: don't allow that irl
```
Panaetius, line 44: is this tracked somewhere so it does not get forgotten? could this just return false instead to be safe?

renku-ssh-proxy/src/proxy_server.rs
```diff
@@ -0,0 +1,432 @@
...
+
+#[derive(Clone)]
+pub struct ProxyHandler {
+    pub target: Option<Target>,
+    upstream: Option<Arc<client::Handle<TargetHandler>>>,
+    channels: HashMap<ChannelId, mpsc::Sender<SshMsg>>,
+    fixed_key: PublicKey,
+}
+
+static FIXED_KEY_STR: &str = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDizMVvJFeooFSwv6qovyXn8SyoFkzKXZFwZFyzAgEmwK8h+OOLrI/Wcea9XuMKIh1QOUdi7i9bmTkUUoMAbK4iFGvRP+ScrmZ9y/8asOj25Vvj484ZhRKv6edVCb4PZyebB2+bvURx7fgaxwzTphxz6ilaJ0bR7PEEWeXKAFLoPPY24ClVGc+MnsBXcyaOO7/ekpDRdrsYBxDltMM9fBOvDISdB8UITGM6tdycm6Lb0qwotdtEK8ly58s6BlWyLtQyIAtKpp7Z6ubCGi74wRBiXPe7GaIaGr8GteiG7NQwPnraNtZdyTQ3/KFqULCa9jrIDuCHhPkaS60AiEGJ6qhZWottRgTDSLk1JpWDfXBjC0ISg2cc1PyCsW2lrooIV3Fvfo/048IMH8x6T5oB37AaR6icKzLaW6RaXLm+/bHNv3tFUl+DGfKAVahThEw9al86JTz7npXhV+0Dlx0vxMKSZO+kMCcaqSXmkh5pp8WH5NoxQL3e9ZsSvexGCo2f0Fk=";
```
Panaetius, line 81: might be cleaner if this came from a file using https://doc.rust-lang.org/std/macro.include_str.html

renku-ssh-proxy/src/proxy_server.rs
```diff
@@ -0,0 +1,432 @@
...
+            config,
+            (target.host.as_str(), target.port),
+            TargetHandler {
+                expected_host_key: target.expected_host_key,
+            },
+        )
+        .await?;
+        // let halg = client::Handle::best_supported_rsa_hash(&handle)
+        //     .await?
+        //     .unwrap_or(Some(ssh_key::HashAlg::Sha256));
```
Panaetius, line 124: Should this be removed?

renku-ssh-proxy/src/proxy_server.rs
```diff
@@ -0,0 +1,432 @@
...
+        // log::debug!("Using hash-alg with target host: {:?}", halg);
+        let auth = handle.authenticate_none(&target.user).await?;
+        if !auth.success() {
+            let auth = handle.authenticate_password(&target.user, "").await?;
```
Panaetius, line 128: I don't know that we want to allow password authentication, it's not very secure.

renku-ssh-proxy/src/proxy_server.rs
```diff
@@ -0,0 +1,432 @@
...
+    type Error = color_eyre::eyre::Error;
+
+    async fn auth_publickey(
+        &mut self,
+        user: &str,
+        public_key: &PublicKey,
+    ) -> std::prelude::v1::Result<server::Auth, Self::Error> {
+        if self.target.is_none() || user.trim().is_empty() {
+            log::warn!("No target host set!");
+            Ok(server::Auth::reject())
```
Panaetius, line 165:
```suggestion
            return Ok(server::Auth::reject());
```
then the else below is not needed, reducing nesting.

renku-ssh-proxy/src/config.rs
```diff
@@ -0,0 +1,198 @@
...
+struct FileConfig {
+    listen: Option<SocketAddr>,
+    host_key: Option<PathBuf>,
+    #[serde(deserialize_with = "deserialize_verbosity")]
+    log_level: Option<Verbosity>,
+    target_port: Option<u16>,
+    target_user: Option<String>,
+    #[serde(with = "humantime_serde")]
+    inactivity_timeout: Option<Duration>,
+}
```
Panaetius, line 89: suggestion: you could manually implement Default or use serde default annotations to have defaults set here directly. Otherwise it's a bit odd to be deriving Default here, everything is just None, and then actual defaults are only set in `resolve()`. Might be nicer to have defaults here where it's implemented.

renku-ssh-proxy/src/config.rs
```diff
@@ -0,0 +1,198 @@
...
+                Some(path) if path.exists() => FileConfig::load(&path)?,
+                _ => FileConfig::default(),
+            },
+        };
+        Settings::resolve(cli, file)
+    }
+}
+
+fn default_config_path() -> Option<PathBuf> {
+    ProjectDirs::from("io", "renku", "ssh-proxy").map(|dirs| dirs.config_dir().join("config.toml"))
```
Panaetius, line 197: this is not something that is expected to run on a user's machine, right? In that case I don't know if using ProjectDirs is such a good idea, as you don't even know if there is a user home for the user the service is running under or that `$HOME` is set etc.. Maybe defaulting to a fixed path would be safer?

helm-chart/renku/templates/ssh-proxy/deployment.yaml
```diff
@@ -0,0 +1,120 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "renku.fullname" . }}-ssh-proxy
+  labels:
+    app: renku-ssh-proxy
+    chart: {{ template "renku.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
```
sgaist, line 9: I am trying to harmonize the labels used throughout the helm chart as part of my build.
There is currently a mix of what you are using here and the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) from k8s.

Thus I would suggest to use the following
```suggestion
        app.kubernetes.io/name: renku-ssh-proxy
        app.kubernetes.io/instance: {{ .Release.Name }}
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/component: data-services
        app.kubernetes.io/part-of: renku
        chart: {{ template "renku.chart" . }}
        heritage: {{ .Release.Service }}
```
I have kept chart and heritage as there are no equivalent.

helm-chart/renku/templates/ssh-proxy/deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+    {{- toYaml .Values.sshProxy.updateStrategy | nindent 4 }}
+  selector:
+    matchLabels:
+      app: renku-ssh-proxy
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: renku-ssh-proxy
+        release: {{ .Release.Name }}
```
sgaist, line 22: Follow up from above
```suggestion
        app.kubernetes.io/name: renku-ssh-proxy
        app.kubernetes.io/instance: {{ .Release.Name }}
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/component: ssh
        app.kubernetes.io/part-of: renku
        chart: {{ template "renku.chart" . }}
        heritage: {{ .Release.Service }}
```

Here I suggest to also add custom labels to make it clear what other services/pods this proxy will access.

From the looks of it, it will likely be `renku-data-service` and `keycloak`

For these I am suggesting:
```
        renku.io/data-service-access: "true"
        {{- if .Values.keycloakx.enabled}}
        renku.io/keycloak-access: "true"
        {{- end }}
```

This will make things easier for devs and admins:
- easier to write network policies
- easier to know what this pod depends on

The if clause is present because if the instance of Renku is connected to an external keycloak, this would be the job of that cluster admin to adjust the network policies.

helm-chart/renku/templates/network-policies.yaml
```diff
@@ -635,3 +635,76 @@ spec:
...
+    - Ingress
+  ingress:
+    - from:
+        - podSelector:
+            matchLabels:
+              app: renku-ssh-proxy
+              release: {{ .Release.Name }}
+      ports:
+        - protocol: TCP
+          port: http
```
sgaist, line 689: With the labeling I am suggesting and the work currently done with the network policy hardening, this is typically a policy that would not need to be added as there will be one allowing the connections between `renku-data-service` and pods with the label `renku.io/data-service-access: true`.

sgaist (changes requested):
Just some suggestion around labeling this new backend that should allow for improved handling in the near future.

helm-chart/renku/templates/ssh-proxy/deployment.yaml
eikek: This is addressed in PR #4616

helm-chart/renku/values.yaml
```diff
@@ -1376,6 +1376,29 @@ dataService:
...
+    dsn:
+    environment:
+    sampleRate: 0.1
+  replicaCount: 1
+  podAnnotations: {}
+  resources: {}
+  updateStrategy: {}
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
```
olevski: Please add a flag to make this enabled/disabled. And make it disabled by default. Some people may not want to allow this at all. Or they may not be able to setup the ingress/load balancing properly on a more bare-metal deployment.

helm-chart/renku/values.yaml
```diff
@@ -1376,6 +1376,29 @@ dataService:
...
+sshProxy:
+  image:
+    repository: harbor.dev.renku.ch/dev-images/renku-ssh-proxy
+    tag: "0.7.0"
+    pullPolicy: IfNotPresent
+  service:
+    port: 2222
+  openSshHostKey:
+  targetUser:
+  targetPort:
```
olevski: Also for openSshHostKey can you add a comment to explain what it is. Maybe how to generate it. Same for `targetPort` and for `targetUser` it would be good to have short ocmment to explain what they mean.

helm-chart/renku/values.yaml
```diff
@@ -1376,6 +1376,29 @@ dataService:
...
+  sentry:
+    enabled: false
+    dsn:
+    environment:
+    sampleRate: 0.1
+  replicaCount: 1
```
olevski: We need to add options for horizontal pod autoscaling too. Look at the data service for example. We can add this in a followup PR. But we need this to go to prod.

helm-chart/renku/templates/ssh-proxy/secret.yaml
```diff
@@ -0,0 +1,19 @@
+{{- $openSshHostKey := (required "sshProxy.openSshHostKey is missing" .Values.sshProxy.openSshHostKey) | b64enc | quote }}
+{{- $renkuFullname := include "renku.fullname" . -}}
+{{- $secretName := cat $renkuFullname "-ssh-proxy" | nospace }}
```
olevski: suggestion: all of these are used just once - no reason to have them as template variables then - just template them directly. Was there a specific reason you decided to use variables?

renku-ssh-proxy/src/config.rs
```diff
@@ -0,0 +1,198 @@
...
+            keys: vec![load_private_key_from_file(&host_key_file)?],
+            preferred: Preferred::default(),
+            ..Default::default()
+        };
+        let ssh_server_config = Arc::new(ssh_server_config);
+
+        let target_user = cli
+            .target_user
+            .or(file.target_user)
+            .unwrap_or("renku".to_string());
```
olevski: Hardcoding this is going to be a problem (not right now but in the future yes). The most reliable way is for the data service to tell you what user you should use in the response for authentication. Users can run the sessions with different users. The buildpacks all use the same one but we do not enforce or check this very strongly. It just cannot be a user with elevated priviliges.

olevski: Can you open an issue please so that we do not forget about this.

olevski (changes requested):

PR #4587, opened 2026-09-04
refactor(chart): bitnami redis -> valkey helm dependency

cmdoret:
## Context
Renku ships redis through bitnami's deprecated chart.

## Goal

Migrate RenkuHelm chart from Bitnami's Redis to the official Valkey chart, a redis-compatible open-source fork.

## Summary

* Subchart updated to valkey
* Adapted values with a new `valkey` section.
* Removed sentinel architecture (not supported in valkey chart), and added persistence in a dedicated volume instead.
* Updated admin and migration docs

## Additional changes

* Fixed valkey resource naming to be release-scoped (was not the case in redis)

/deploy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/docs/20-admins/10-architecture/10-services.md
+++ b/docs/docs/20-admins/10-architecture/10-services.md
@@ -28,7 +28,7 @@ in Kubernetes.
 ### Third party
 
 - **Postgres**: Stores all information about projects, users and sessions.
-- **Redis**: Stores user login sessions.
+- **Valkey**: Stores user login sessions.
 - **Authzed (SpiceDB)**: Stores authorization information and is queries for authorization decisions.
 - **Solr**: Stores search documents and indices.
 - **Keycloak**: Stores user information and is used for Authentication.
@@ -49,7 +49,7 @@ flowchart LR
     end
     DS[Data services]
     PG[(Postgres)]
-    Redis[(Redis)]
+    Valkey[(Valkey)]
     Solr(Solr)
     Ingress --/api/*--> Proxy
     Proxy --/api/data/*--> DS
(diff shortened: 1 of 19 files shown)


cmdoret:
Tested with and without persistent volume on ci deployment:
* Both work
* Persistent volume correctly retains user sessions upon deleting the pod

leafty (commented):
It looks like the `valkey` chart is not great for production use until `sentinel` is available (may be in https://github.com/valkey-io/valkey-helm/pull/234).

With `replica` mode, we only get read replicas with a single point of failure for writes which means that the `master` pod is locked down and cannot be moved except during maintenance windows. The `redis+sentinel` chart allows redis pods to be moved (PDB allowing one pod at a time to be disrupted) so that background maintenance can be performed (e.g. node rotation).

cmdoret:
> It looks like the `valkey` chart is not great for production use until `sentinel` is available (may be in [valkey-io/valkey-helm#234](https://github.com/valkey-io/valkey-helm/pull/234)).
> 
> With `replica` mode, we only get read replicas with a single point of failure for writes which means that the `master` pod is locked down and cannot be moved except during maintenance windows. The `redis+sentinel` chart allows redis pods to be moved (PDB allowing one pod at a time to be disrupted) so that background maintenance can be performed (e.g. node rotation).

The long term solution would be to use the [valkey-operator](https://github.com/valkey-io/valkey-operator), which has support for clusters, but it is currently in alpha.

Indeed, sentinel support is the next planned feature for [valkey-helm's 1.0 milestone.](https://github.com/valkey-io/valkey-helm/milestone/4)

Shall we then stall this PR until they merge sentinel support?

leafty:
> Shall we then stall this PR until they merge sentinel support?

If temporary deployments use valkey and it works it's fine to have the new chart alongside the redis one. Production deployments would have to stay on redis until there is a viable solution to use valkey.

PR #4604, opened 2026-09-10
fix: use trixie for Python scripts

leafty:
Using `bullseye` fails to build -> https://github.com/SwissDataScienceCenter/renku/actions/runs/34371500998/job/102780079033

/deploy


sambuc (approved):
LGTM

PR #4620, opened 2026-09-18
feat: fetch token from keycloak service account for ssh proxy

SalimKayal:
## Summary

The SSH proxy authenticates to data-services with a short-lived Keycloak `client_credentials` token (realm role `ssh-proxy`).

## Motivation and context

#4616 made `POST /internal/sessions/{id}/authorize` require the `ssh-proxy` realm role; the proxy sent no credentials. This adds the token client so the proxy can present it, with no JWT library and no master signing key in the internet-facing pod.

## Behavior

- Token fetched from `{keycloak_url}/realms/{realm}/protocol/openid-connect/token`, cached with a 30s refresh margin.
- On a data-services `401` the token is refreshed and the call retried once; still failing → connection rejected (fail closed).
- Keycloak unreachable or missing Keycloak config → startup error / rejected connection; the client secret is never logged.

## Changes

- `src/keycloak.rs` (new): `TokenProvider` + `KeycloakSettings` (manual `Debug` redacts the secret).
- `src/config.rs`: required `keycloak_{url,realm,client_id,client_secret}` (flags/`RENKU_SSH_PROXY_KEYCLOAK_*`/file), added to `Settings`.
- `src/data_services.rs`, `src/proxy_server.rs`: attach `Authorization: Bearer`, refresh + retry once on 401, build the provider.
- `nix/default.nix`: devshell placeholder Keycloak env vars.
- `Cargo.toml`/`Cargo.lock`: `form_urlencoded` as a direct dep, `wiremock` dev-dep.

/deploy #notest extra-values=sshProxy.targetUser=renku,sshProxy.targetPort=2222,dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user amalthea-sessions=salimkayal/feat/2222-service-in-ssh-sessions renku-data-services=salimkayal/feat/ssh-proxy-service-account-authn


renku-ssh-proxy/src/keycloak.rs
```diff
@@ -0,0 +1,198 @@
...
+                .map_err(|_| eyre!("Cannot extend base url: {}", settings.url))?;
+            path.extend([
+                "realms",
+                settings.realm.as_str(),
+                "protocol",
+                "openid-connect",
+                "token",
+            ]);
+        }
+        let client = reqwest::Client::builder().timeout(*timeout).build()?;
```
eikek, line 75: Perhaps we could reuse an existing client here, by passing it as an argument? Inside `Client::new` we could also instantiate the `TokenProvider` and let it be owned there. Right now, we don't need it anywhere else afaics.

renku-ssh-proxy/src/proxy_server.rs
```diff
@@ -13,7 +14,15 @@ use tokio::sync::mpsc;
 
 /// Creates and runs a proxy server
 pub async fn serve_proxy(settings: &Settings) -> Result<()> {
-    let client = Client::new(&settings.data_services_url, &settings.data_services_timeout)?;
+    let tokens = Arc::new(TokenProvider::new(
+        &settings.keycloak,
+        &settings.data_services_timeout,
+    )?);
+    let client = Client::new(
```
eikek, line 21: From my other comment: here I would make the token provider inside the client constructor (passing it in could still be left as an alternate way of a creating a client). I think we don't need it anywhere else and this way we could have just one client hanging around. wdyt?

eikek (changes requested):
Very nice! Just one small idea

PR #4613, opened 2026-09-16
Authorize ssh access using data services

eikek:
- call to data_services to authorize access to a ssh session

----

[...]

PR #4627, opened 2026-09-21
feat: ssh proxy session authn

SalimKayal:
## Summary

The proxy pins the session host key and authenticates to `session:2222` with a private key instead of `authenticate_none`/empty password; the proxy pod mounts the two hop-2 Secrets, and the chart passes their names to data-services. Stacked on the keygen PR (`salimkayal/feat/ssh-proxy-session-keygen`).

## Motivation and context

Hop 2 was passwordless: the proxy did `authenticate_none` + `authenticate_password("")` and `check_server_key` accepted any host key, so any pod that could reach `session:2222` got a shell and the proxy could not tell a real session host from an impostor. Hop 1 (user → proxy) is already authenticated and authorized by data-services, so hop 2 only needs to answer "is the caller the proxy?" and "is the upstream a legitimate session host?". This PR consumes the Secrets created by the keygen PR.

## Behavior

- The proxy pins the session host key: `check_server_key` compares key data only (the wire key carries no comment) and rejects when no pin is configured.
- The proxy authenticates with a private key (`authenticate_publickey`) and fails closed: missing key config is a startup error, auth/key-load errors reject the connection, and the auth private key has a redacting `Debug`.
- The proxy pod mounts the session host public key and the proxy auth private key; data-services is told the two Secret names so it can mount the matching files into sessions (see the data-services session-mounts PR).

## Changes

- `renku-ssh-proxy/src/config.rs`: required `session_host_key` / `session_auth_key` paths (`RENKU_SSH_PROXY_SESSION_{HOST,AUTH}_KEY`), a public-key loader, a redacting `SessionAuthKey`, and the pin wired into `Target`.
- `renku-ssh-proxy/src/proxy_server.rs`: `check_server_key` pins by key data and fails closed without a pin; `connect_target` uses `authenticate_publickey` (`PrivateKeyWithHashAlg`); `ProxyHandler::with_auth_key`.
- `helm-chart/renku/templates/ssh-proxy/dployment.yaml`: mount the session host public key and the proxy auth private key, and set the two env vars.
- `helm-chart/renku/templates/data-service/session-env-secret.yaml`: pass `NB_SESSIONS__SSH__PROXY_{HOST,AUTH}_KEY_SECRET` to data-services.
- `renku-ssh-proxy/nix/default.nix`: dev-only: devshell values and dev-VM host key / `authorized_keys` so hop 2 can be exercised locally.

## Notes

- Depends on the #4626 and, for end-to-end sessions, the data-services session-mounts PR (mounts the files the buildpack and dropbear consume).
- Verification: `cd renku-ssh-proxy && cargo test` (11 tests) and `cargo clippy`; the proxy deployment renders with the two env vars, mounts and volumes; the whole chart renders.

/deploy #notest extra-values=sshProxy.targetUser=renku,sshProxy.targetPort=2222 amalthea-sessions=salimkayal/feat/2222-service-in-ssh-sessions renku-data-services=salimkayal/feat/ssh-proxy-session-mounts


renku-ssh-proxy/src/config.rs
```diff
@@ -42,6 +42,24 @@ where
     }
 }
 
+fn load_public_key_from_file<P>(file: P) -> Result<russh::keys::PublicKey>
```
eikek, line 34: When using k8s secrets, I had to base64 decode it first. Maybe that is not required here as it is public data, just thought mention it (see `load_private_key_from_file` above).

eikek (approved):
👍🏼

renku-ssh-proxy/src/config.rs
SalimKayal: that shouldn't be an issue since k8s secrets are base64 decoded when mounted. in https://github.com/SwissDataScienceCenter/renku/pull/4613 you provided the value pre base64 encoded in the values so k8s re encoded it. maybe that's why it was needed ?

afaik it shouldn't be needed anymore. let me removing it to see whether it works.

SalimKayal: yep, seems to work without it.

eikek: Does this mean we need to setup a secret manually for pr deployments? (for the hostkey presented to the outside)

SalimKayal: no, it's generated by helm

https://github.com/SwissDataScienceCenter/renku/blob/35845b741c45202c10a1b571e2ce33b25c4d87b4/helm-chart/renku/templates/ssh-proxy/secret.yaml#L3

it's introduced in #4616

